PR #15, opened 2020-03-27
role data interface and user to role data interface with tests

evmiguel:
This PR is for the:

  * ~~`/role` POST API:~~

  * `/user/role` POST API:
```
curl --location --request POST 'localhost:5000/api/user/role' \
--header 'Content-Type: application/json' \
--data-raw '{
	"user_id": 1,
	"role_id": 1
}'
```

run `cd server; yarn jest` to run the tests


server/data/models/__mocks__/roleModel.js
```diff
@@ -0,0 +1,7 @@
+const create = require('../../../tests/util/create');
+
+const Role = {
```
spectranaut: I'm guessing we don't need this model, for the same reason, that it won't be exposed? Or do we somehow need it to get the foreign key for the users to roles table?

server/data/models/roleModel.js
```diff
@@ -0,0 +1,18 @@
+const { DataTypes } = require('sequelize');
+const sequelize = global.sequelize;
+
+const Role = sequelize.define(
+    'Role',
```
spectranaut: Same question about whether role needs to exist!

server/data/services/__mocks__/roleService.js
```diff
@@ -0,0 +1,7 @@
+const create = require('../../../tests/util/create');
+
+const RoleService = {
+    addRole: create
```
spectranaut: Same Q about roles.

server/data/services/roleService.js
```diff
@@ -0,0 +1,15 @@
+const Role = require('../models/roleModel');
+
+class RoleService {
+    static async addRole(role) {
```
spectranaut: Same Q about roles.

PR #21, opened 2020-03-31
express session integration

evmiguel:
This PR is the preliminary request session infrastructure that will be used by the login flow.

### DIFF
--- a/server/middleware/session.js
+++ b/server/middleware/session.js
@@ -1,0 +1,16 @@
+const session = require('express-session');
+const { Pool } = require('pg');
+const pgSession = require('connect-pg-simple')(session);
+const { getDbConfig } = require('../util/db-config');
+const poolConfig = getDbConfig();
+
+module.exports = session({
+    secret: process.env.SESSION_SECRET || 'aria at report',
+    resave: false,
+    saveUninitialized: true,
+    store: new pgSession({
+        pool: new Pool(poolConfig),
+        tableName: 'session'
+    }),
+    cookie: { maxAge: 30 * 24 * 60 * 60 * 1000 } // 30 days
+});
--- a/server/tests/integration/server.int.test.js
+++ b/server/tests/integration/server.int.test.js
@@ -1,13 +1,42 @@
 const request = require('supertest');
-const listener = require('../../app');
+const { app, listener } = require('../../app');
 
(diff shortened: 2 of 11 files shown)


server/middleware/session.js
```diff
@@ -0,0 +1,20 @@
...
+            user: process.env.PGUSER,
+            password: process.env.PGPASSWORD,
+            database: process.env.PGDATABASE,
+            host: process.env.PGHOST,
+            port: process.env.PGPORT
+        }),
+        tableName: 'session'
+    }),
+    cookie: { maxAge: 30 * 24 * 60 * 60 * 1000 } // 30 days
+});
```
rwaldron, line 20: I migrated another Heroku app to use the following: 

(db-config.js)
```js
const envConfig = require('dotenv').config().parsed || {};
const devConfig = require('../database.json').dev;

exports.getDbConfig = function() {
    // This is the environment variable Heroku exposes
    const connectionString = process.env.DATABASE_URL || envConfig.DATABASE_URL;
    if (connectionString) {
        return {
            connectionString,
            ssl: true,
            sslmode: 'require'
        };
    }

    return devConfig;
};
```

Then you can have local credentials in a `database.json` and use the Heroku provided `DATABASE_URL`. 

You would then change this code to something like: 

```js
const pgSession = require('connect-pg-simple')(session);
const poolConfig = getDbConfig();

module.exports = session({
    secret: process.env.SESSION_SECRET || 'aria at report',
    resave: false,
    saveUninitialized: true,
    store: new pgSession({
        pool: new Pool(poolConfig()),
        tableName: 'session'
    }),
    cookie: { maxAge: 30 * 24 * 60 * 60 * 1000 } // 30 days
});
```

The reason you want to use `DATABASE_URL` is because that is tied directly to Heroku's built-in DB maintenance capabilities.

server/tests/integration/server.int.test.js
```diff
@@ -1,13 +1,42 @@
...
+    });
+
+    it('uses session middleware', async () => {
+        const response = await agent.get('/');
+        expect(response.status).toBe(404);
+        expect(session).toHaveBeenCalledTimes(1);
+    });
+});
+
+jest.mock('../../routes/users');
```
rwaldron, line 39: This is great

PR #22, opened 2020-04-01
Oauth login api

evmiguel:
This PR is the implementation of a login flow with GitHub. @spectranaut I'd like to walk you through it

Here is a [DEMO](https://drive.google.com/file/d/1_-TLvmHzakZJcbgkU_YvdBXOhnoJ_Psi/view?usp=sharing)


server/data/services/__mocks__/githubService.js
```diff
@@ -0,0 +1,8 @@
+const create = require('../../../tests/util/create');
```
rwaldron, line 1: I went and looked at this and I don't think that code warrants its own util module. It can also be rewritten as: 

```js
const create = o => Promise.resolve(o);
```

server/data/services/__mocks__/githubService.js
```diff
@@ -0,0 +1,8 @@
...
+
+const GithubService = {
+    login: () => 'localhost:5000/login',
+    authorize: create
+};
+
+module.exports = GithubService;
```
rwaldron, line 8: Given my notes about `create` above, I'd suggest : 

```js
const authorize = o => Promise.resolve(o);
const login = () => 'localhost:5000/login';
module.exports = {
  authorize, 
  login
};
```

server/data/services/githubService.js
```diff
@@ -0,0 +1,25 @@
...
+            },
+            {
+                headers: { Accept: 'application/json' }
+            }
+        );
+        return response.data.access_token;
+    }
+}
+
+module.exports = GithubService;
```
rwaldron, line 25: As a general convention, we should name the files the same as the class/object/module that is exported (unless the component is a directory, then name it index.js or the NameOfComponentClass.js). I noticed it in other files in this PR, I think you should fix these all in a follow up PR.

server/data/services/githubService.js
```diff
@@ -0,0 +1,25 @@
+const axios = require('axios');
+
+class GithubService {
```
rwaldron, line 3: Same points I wrote in `AuthController`, this is a "class" here, but doesn't need to be

server/routes/auth.js
```diff
@@ -0,0 +1,9 @@
+const { Router } = require('express');
+const AuthController = require('../data/controllers/authController');
```
rwaldron, line 2: This could be `const { login, authorize } = ...`

client/static/index.html
```diff
@@ -7,9 +7,11 @@
     <meta http-equiv="Content-Language" content="en">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="stylesheet" type="text/css" href="/root-styles.css">
+    <title>ARIA-AT REPORT</title>
 </head>
 <body>
     <div id="root"></div>
+    <p>ARIA-AT REPORT</p>
```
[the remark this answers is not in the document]
evmiguel: We talked about moving this into the `<App />` Component for now

client/components/Login/index.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Button } from 'reakit';
+
+const Login = () => {
+    const loginURL = `${process.env.API_SERVER}/api/auth/login?referer=${location.origin}&service=github`;
```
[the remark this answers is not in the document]
evmiguel: There is a question about `location.origin` and iframe. This might be irrelevant.

server/data/controllers/authController.js
```diff
@@ -0,0 +1,24 @@
...
+        const { service, code } = req.query;
+        const authService =
+            services[`${service}Service`] || services.githubService;
+        req.session.accessToken = await authService.authorize(code);
+        res.redirect(303, req.session.referer);
+        delete req.session.referer;
+        res.end();
+    }
+}
+
```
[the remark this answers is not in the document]
evmiguel: Will implement change to have bag of methods

server/data/services/githubService.js
evmiguel: Will do this!

server/data/controllers/AuthController.js
```diff
@@ -0,0 +1,28 @@
+const services = require('../services');
+
+const capitalizeServiceString = service =>
+    `${service.charAt(0).toUpperCase()}${service.slice(1)}`;
+
+module.exports = {
+    login: (req, res) => {
```
rwaldron, line 7: Since there's no explicit use of `this` value in this function, you should just write these `login(req, res) {`

server/data/controllers/AuthController.js
```diff
@@ -0,0 +1,28 @@
...
+            services[`${capitalizeServiceString(service)}Service`] ||
+            services.GithubService;
+        req.session.referer = referer;
+        // get URL instead
+        const loginURL = authService.url;
+        res.redirect(303, loginURL);
+        res.end();
+    },
+
+    authorize: async (req, res) => {
```
rwaldron, line 19: For the same reason as above: `async authorize(req, res) {`

server/data/controllers/UsersController.js
```diff
@@ -1,16 +1,16 @@
-const UsersService = require('../services/usersService');
+const UsersService = require('../services/UsersService');
 
-class UsersController {
-    static async addUser(req, res) {
+module.exports = {
+    addUser: async (req, res) => {
```
rwaldron, line 4: Same as the previous comments

PR #29, opened 2020-04-10
Manage cycles page and navigation

spectranaut:


client/components/CycleRow/index.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { connect } from 'react-redux';
+import { Button } from 'react-bootstrap';
+import { deleteCycle } from '../../actions/cycles';
+
+class CycleRow extends Component {
+
+    deleteCycle() {
+        const { id } = this.props.cycle;
+
+        // TODO: How do we call the delete cycle action..? :/
```
spectranaut, line 12: @evmiguel, tbh I'm not sure whats the best way to dispatch the `deleteCycle` action... how do I get the `dispatch` function at this level?:/

client/components/App/index.jsx
```diff
@@ -1,29 +1,40 @@
...
         const { location, dispatch } = this.props;
         const { login } = queryString.parse(location.search);
+
         if (login) {
             dispatch(handleLogin());
         }
+
+        // TODO: I don't think this is the appropriate place to get this into the application
+        // maybe I should do this on the testCycles app. Or, maybe I should only do this AFTER log in...?
+        dispatch(getTestCycles());
```
spectranaut, line 22: Also I'm not sure exactly when to get the test cycle data either.

client/actions/cycles.js
```diff
@@ -0,0 +1,28 @@
+import axios from 'axios';
+import { CYCLES, DELETE_CYCLE } from './types';
+
+export const deleteCycleDispatch = data => ({
```
evmiguel, line 4: Let's just call data `payload`. Per Rick's suggestions on my last PR

PR #32, opened 2020-04-10
login and logout persistence

evmiguel:
This PR persists login and logout state.

Here is a [DEMO](https://drive.google.com/file/d/1TQjmYazRYVLb7IXZRwyegtFLY5uoQQiE/view?usp=sharing)

### DIFF
--- a/server/controllers/AuthController.js
+++ b/server/controllers/AuthController.js
@@ -28,7 +28,7 @@ module.exports = {
             res.status(401);
             res.end();
         }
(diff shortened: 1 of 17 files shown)


server/controllers/AuthController.js
```diff
@@ -28,7 +28,7 @@ module.exports = {
             res.status(401);
             res.end();
         }
-        res.redirect(303, `${req.session.referer}?login=true`);
```
evmiguel, line 31: No longer need this query param to persist login

PR #33, opened 2020-04-10
User settings page

evmiguel:
This is the first iteration on the user settings page. Here is a [DEMO](https://drive.google.com/file/d/1OfpS-6tXPvDAprDjvI0j5vBFh4o60HGc/view?usp=sharing)

Things that need to be done in a follow up:
- [ ] Better styling
- [ ] Save the ATs for a given user (after sign up flow implemented)

### DIFF
--- a/client/components/UserSettings/index.jsx
+++ b/client/components/UserSettings/index.jsx
@@ -1,14 +1,85 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import Row from 'react-bootstrap/Row';
+import Col from 'react-bootstrap/Col';
+import Form from 'react-bootstrap/Form';
+import Button from 'react-bootstrap/Button';
+import { handleGetValidAts } from '../../actions/ats';
 
 class UserSettings extends Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {};
+
+        this.onCheckboxClicked = this.onCheckboxClicked.bind(this);
+    }
+
+    componentDidMount() {
+        const { dispatch } = this.props;
+        dispatch(handleGetValidAts());
+    }
+
+    componentDidUpdate(prevProps) {
+        if (prevProps.ats !== this.props.ats) {
+            this.setState(
+                Object.assign(
+                    {},
+                    ...this.props.ats
+                        .map(at => [at, { checked: false }])
+                        .map(([k, v]) => ({ [k]: v }))
+                )
+            );
+        }
+    }
+
+    onCheckboxClicked(event) {
+        if (event.target) {
+            const { id } = event.target;
+            // Need to do this check because some SyntheticEvents have a null target
+            if (id) {
+                this.setState(prevState => {
+                    return {
+                        ...prevState,
+                        [id]: {
+                            checked: !prevState[id].checked
+                        }
+                    };
+                });
+            }
+        }
+    }
     render() {
-        const { isLoggedIn, username, name } = this.props;
+        const { ats, isLoggedIn, username, email } = this.props;
         const contents = isLoggedIn ? (
-            <p data-test="user-settings-authorized">
-                {username} {name}
-            </p>
+            <Row data-test="user-settings-authorized">
+                <Col>
+                    <h2>USER DETAILS</h2>
+                    <p>{username}</p>
+                    <p>{email}</p>
+                </Col>
+                <Col>
+                    <h2>Assistive Technology</h2>
+                    <p>Add the assistive technologies that you can use</p>
+                    <Form>
+                        <Form.Group controlId="formBasicCheckbox">
+                            {ats &&
+                                ats.map(at => (
+                                    <Form.Check
+                                        id={at}
+                                        key={at}
+                                        label={at}
+                                        onClick={this.onCheckboxClicked}
+                                    />
+                                ))}
+                        </Form.Group>
+                        <Button variant="primary" type="submit">
+                            Save
+                        </Button>
+                    </Form>
+                </Col>
+            </Row>
         ) : (
             <p data-test="user-settings-unauthorized">Unauthorized</p>
         );
@@ -20,12 +91,15 @@ class UserSettings extends Component {
 UserSettings.propTypes = {
     isLoggedIn: PropTypes.bool,
     username: PropTypes.string,
-    name: PropTypes.string
+    email: PropTypes.string,
+    ats: PropTypes.array,
+    dispatch: PropTypes.func
 };
 
 const mapStateToProps = state => {
-    const { isLoggedIn, username, name } = state.login;
-    return { isLoggedIn, username, name };
+    const { isLoggedIn, username, email } = state.login;
+    const { names: atNames } = state.ats;
+    return { isLoggedIn, username, email, ats: atNames };
 };
 
 export default connect(mapStateToProps)(UserSettings);
--- a/client/reducers/ats.js
+++ b/client/reducers/ats.js
@@ -1,0 +1,18 @@
+import { ATS } from '../actions/types';
+
+const initialState = {
+    names: []
+};
+
+export default (state = initialState, action) => {
+    switch (action.type) {
+        case ATS: {
+            const { names } = action.payload;
+            return {
+                names
+            };
+        }
+        default:
+            return state;
+    }
+};
--- a/client/tests/integration.test.js
+++ b/client/tests/integration.test.js
@@ -1,5 +1,6 @@
 import { storeFactory } from './util';
 import { handleCheckLoggedIn, handleLogout } from '../actions/login';
+import { handleGetValidAts } from '../actions/ats';
 import moxios from 'moxios';
 
 describe('login actions dispatchers', () => {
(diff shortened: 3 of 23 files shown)


client/components/Checkbox/index.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Form from 'react-bootstrap/Form';
+
+const Checkbox = props => {
+    const { id, label, onClick } = props;
+    return (
+        <Form.Check type="checkbox" id={id} label={label} onClick={onClick} />
```
rwaldron, line 8: Can this be simplified to: `<Form.Check type="checkbox" {...props} />`?

client/components/UserSettings/index.jsx
```diff
@@ -1,14 +1,86 @@
...
+    }
+
+    componentDidUpdate(prevProps) {
+        if (prevProps.ats !== this.props.ats) {
+            this.setState(
+                Object.assign(
+                    {},
+                    ...this.props.ats
+                        .map(at => [at, { checked: false }])
+                        .map(([k, v]) => ({ [k]: v }))
```
rwaldron, line 32: Very sharp!!

client/components/UserSettings/index.jsx
```diff
@@ -1,14 +1,86 @@
...
+                )
+            );
+        }
+    }
+
+    checkBoxClicked(event) {
```
rwaldron, line 38: Suggestion: It might be easier to associate this in the future if it's named closer to the functionality it provides. I just ran into this in TSL/TM. A convention that helps for me is "on" + "thing" + "event", ie. `onCheckboxClick`. This is obviously not a blocker, but like I said, I encountered some confusing event handler names in TSL/TM yesterday and ended up renaming them to match this pattern.

client/components/UserSettings/index.jsx
```diff
@@ -1,14 +1,86 @@
...
+        if (event.target) {
+            const { id } = event.target;
+            // Need to do this check because some SyntheticEvents have a null target
```
rwaldron, line 41: Smart

client/components/UserSettings/index.jsx
```diff
@@ -1,14 +1,86 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import Row from 'react-bootstrap/Row';
+import Col from 'react-bootstrap/Col';
+import Form from 'react-bootstrap/Form';
+import Button from 'react-bootstrap/Button';
+import Checkbox from '@components/Checkbox';
+import { handleGetValidAts } from '../../actions/ats';
```
rwaldron, line 9: Suggestion: make a module that imports and re-exports only the components that your application needs, then import from _that_ module: 

(`client/components/UI/index.js`)
```js
import Button from 'react-bootstrap/Button';
import Checkbox from '@components/Checkbox';
import Col from 'react-bootstrap/Col';
import Form from 'react-bootstrap/Form';
import Row from 'react-bootstrap/Row';

export {
  Button,
  Checkbox,
  Col,
  Form,
  Row,
};
```

Then in your other components: 

```js
import {
  Button,
  Checkbox,
  Col,
  Form,
  Row
} from '@components/UI';
```


(Then you can even group your custom components in that single module, as I've done here with `Checkbox`)

client/components/UserSettings/index.jsx
```diff
@@ -20,12 +92,15 @@ class UserSettings extends Component {
...
+    email: PropTypes.string,
+    ats: PropTypes.array,
+    dispatch: PropTypes.func
 };
 
 const mapStateToProps = state => {
-    const { isLoggedIn, username, name } = state.login;
-    return { isLoggedIn, username, name };
+    const { isLoggedIn, username, email } = state.login;
+    const { ats } = state.ats;
```
rwaldron, line 102: This is unpacking a property named `ats` from an object value stored at `state.ats`, that is to say that it assumes `state` looks like this: 

```js
{
  ats: {
    ats: ...
  }
}
```

When I log `state.ats`, I see an array. So I think this should be: 

```js
const { ats } = state;
```

**Amendment:** 

As I continued to review, I saw that `state.ats.ats` and `state.cycles.cycles` are currently correct. I think this needs to be fixed so that `ats` are at `state.ats` and `cycles` are at `state.cycles`

client/reducers/ats.js
```diff
@@ -0,0 +1,18 @@
+import { ATS } from '../actions/types';
+
+const initialState = {
+    ats: []
+};
```
rwaldron, line 5: For my recommendation above, this should be `const initialState = []`;

client/reducers/ats.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default (state = initialState, action) => {
+    switch (action.type) {
+        case ATS: {
+            const { ats } = action.payload;
+            return {
+                ats
+            };
```
rwaldron, line 13: For my recommendation above, this should be `return [ ...ats ];`

client/tests/integration.test.js
```diff
@@ -17,10 +18,14 @@ describe('login actions dispatchers', () => {
...
                 username: 'foobar',
-                name: 'Foo Bar'
+                name: 'Foo Bar',
+                email: 'foo@bar.com'
             },
             cycles: {
                 cycles: []
+            },
+            ats: {
+                ats: []
```
rwaldron, line 28: For my recommendation above, these should be changed to: 

```js
cycles: [], 
ats: [],
```

client/tests/integration.test.js
```diff
@@ -48,6 +54,9 @@ describe('login actions dispatchers', () => {
...
```
rwaldron, line 59: And again here

client/tests/integration.test.js
```diff
@@ -63,3 +72,41 @@ describe('login actions dispatchers', () => {
...
+        const expectedState = {
+            login: {
+                isLoggedIn: false
+            },
+            cycles: {
+                cycles: []
+            },
+            ats: {
+                ats: ['JAWS', 'NVDA', 'VoiceOver']
+            }
```
rwaldron, line 95: And here

rwaldron:
I forgot this in my review: when the browser window narrows and the responsive UI reacts, the "Menu" gets hidden by a hamburger icon, but then requires clicking the hamburger icon _and_ the "Menu" dropdown to see the menu items.

client/components/Checkbox/index.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Form from 'react-bootstrap/Form';
+
+const Checkbox = props => {
```
spectranaut, line 5: I probably wouldn't have broken this out into it's on component, are you expecting that we will have a checkbox component that differs somehow from the underlying bootstrap checkbox component? I think it's possible we should just use the bootstrap checkbox directly if we are not modifying, for simplicity's sake.

client/components/UserSettings/index.jsx
spectranaut: nice trick!

client/tests/integration.test.js
spectranaut: Seeing these comments makes me think we should handle the actions/reducers a bit differently. @rwaldron I added more information to the "cycles" reducer here in a way that we can't simplify the cycles state to just `[]`: https://github.com/bocoup/aria-at-report/pull/34/files#diff-3b3898c55e860045e6333773efb63da8

I think having a whole `action` and `reducer` for each item we want to keep in global state is a little bit extreme.. which is the direction Rick's comments push us in. On the other hand, I think it's going to get confusing what item should be in which `reducer` file (which is then imported in `reducers/index.js` and will lead to things like `state.cycle.cycle`. Right now -- I'm thinking perhaps we should put all the actions in one file and all the reducers in one file. Or, we should really architect how they should be split up -- because right now I'm not positive how to split them up.

rwaldron: > On the other hand, I think it's going to get confusing what item should be in which reducer file (which is then imported in reducers/index.js and will lead to things like state.cycle.cycle

That's where you're model is now: `state.cycles.cycles`. If revise and follow my recommendation, your `cycles` reducer will only deal with reducing for `cycles`, then it can always just return that `cycles` array. You don't even need to split things into different files, you just need to reorganize them structurally and not rely on default exports. I will comment on your changeset with how I imagine that can be done.

rwaldron: @spectranaut https://github.com/bocoup/aria-at-report/pull/34#discussion_r408991800

client/components/UserSettings/index.jsx
evmiguel: @rwaldron, I think I would like the state to be `state.ats.names` because I intend to use the `state.ats` object to have more information

client/reducers/ats.js
evmiguel: Same design decision as above: https://github.com/bocoup/aria-at-report/pull/33/files/8b0dd24941eaaf2d296a4eb46c180d335d17321f#r409082767

PR #34, opened 2020-04-14
Initiate and manage test cycles

spectranaut:
This feature is not quite done, but it would probably be good to get it reviewed and in.

Things that need to be done in follow up tasks (or here):
- [ ] Better styling
- [ ] Alerts to handle when things automatically change on the initiate test cycle page (like when you try to assign a user who cannot test a specific AT to a run with that AT)
- [ ] On the initialize test cycle page, the message before a example configuration which tells you how many runs don't have testers assigned to them
- [ ] On the initialize test cycle page, the ability to create a test run with only some of the examples
- [ ] A page to edit the test cycle after it's been initialized
- [ ] Form validation/when you try to click "initialize test cycle". For example, if there is no "AT version" specified, you should be prevent from being able to initialize the run, and the dom focus should be put on that field.


spectranaut:
I'm confused why the build fails because the latest commit fixes that test!

client/components/InitiateCycle/index.jsx
```diff
@@ -0,0 +1,344 @@
...
+        };
+
+        /*
+         * at_version,
+         * browser_version,
+         * at_id,
+         * browser_id,
+         * apg_example_id,
+         * users: [user_id, user_id]
+         */
```
spectranaut, line 30: I need to delete this comment!

client/components/InitiateCycle/index.jsx
```diff
@@ -0,0 +1,344 @@
...
+
+        this.handleVersionChange = this.handleVersionChange.bind(this);
+        this.handleNameChange = this.handleNameChange.bind(this);
+        this.handleTechnologyRowChange = this.handleTechnologyRowChange.bind(
+            this
+        );
+        this.addTechnologyRow = this.addTechnologyRow.bind(this);
+        this.assignTesters = this.assignTesters.bind(this);
+        this.removeAllTestersFromRun = this.removeAllTestersFromRun.bind(this);
+        this.initiateCycle = this.initiateCycle.bind(this);
```
spectranaut, line 40: I should normalize all these function names.

client/reducers/cycles.js
```diff
@@ -1,26 +1,46 @@
...
-            };
+            });
+        }
+        case SAVE_CYCLE: {
+            const cycle = action.payload;
+            let newCycleList = [...state.cycles];
+            newCycleList.unshift(cycle);
+            return Object.assign({}, state, {
+                cycles: newCycleList
+            });
```
rwaldron, line 43: I recommend either putting `testSuiteVersions` in its own reducer file, or changing the structure of this file to not rely on default exports: 

```js
import {
    CYCLES,
    DELETE_CYCLE,
    SAVE_CYCLE,
    TEST_SUITE_VERSIONS
} from '../actions/types';

const initialCycleState = [];

export const cycles = (state = initialCycleState, action) => {
    switch (action.type) {
        case CYCLES: {
            const cycles = action.payload;
            return [
                ...cycles
            ];
        }
        case DELETE_CYCLE: {
            const { cycleId } = action.payload;
            let remainingCycles = state.cycles.filter(
                cycle => cycle.id !== cycleId
            );
            return [
                ...remainingCycles
            ];
        }
        case SAVE_CYCLE: {
            const cycle = action.payload;
            let newCycleList = [...state.cycles];
            newCycleList.unshift(cycle);
            return [
                ...newCycleList
            ];
        }
        default:
            return state;
    }
};

const initialTestSuiteVersionsState = [];

export const testSuiteVersions = (state = initialTestSuiteVersionsState, action) => {
    switch (action.type) {
        case TEST_SUITE_VERSIONS: {
            const testSuiteVersions = action.payload;
            return [
                ...testSuiteVersions
            ];
        }
        default:
            return state;
    }
};
```

Then in your `reducers/index.js`, you'd import the reducers by export name: 

```js
import { combineReducers } from 'redux';
import login from './login';
import {cycles, testSuiteVersions} from './cycles';
import users from './users';

export default combineReducers({
    login,
    cycles,
    testSuiteVersions,
    users
});
```

client/components/InitiateCycle/index.jsx
```diff
@@ -0,0 +1,344 @@
...
+            version => version.id === this.state.selectedVersion
+        )[0];
+
+        let cycle = {
+            name: this.state.name,
+            test_version_id: this.state.selectedVersion,
+            created_user_id: 1
+        };
+
+        // TODO: Do not allow saving of cycle without a name. Put focus on name field
```
evmiguel, line 80: Wanted to bump this!

client/reducers/users.js
```diff
@@ -0,0 +1,19 @@
...
+const initialState = {
+    users: []
+};
+
+export default (state = initialState, action) => {
+    switch (action.type) {
+        case USER_CONFIGS: {
+            const userConfigs = action.payload;
+            return Object.assign({}, state, {
+                users: userConfigs
```
evmiguel, line 13: Non-blocking. I'm wondering if we should change this to to userConfigs

client/actions/users.js
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios';
+import { USER_CONFIGS } from './types';
+
+export const getAllUsersDispatch = payload => ({
+    type: USER_CONFIGS,
+    payload
+});
+
+export function getAllUsers() {
```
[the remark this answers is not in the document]
spectranaut: I was thinking this should be the api you use to get users.. get user config is perhaps misnamed on the backend. If you want all the users in the system for some reason (as an admin, to override their settings, or to know who can test what) you would just use this endpoint. It will come with their configs by default, but you need that data almost every time  you need the list of users anyway.

client/actions/users.js
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios';
+import { USER_CONFIGS } from './types';
+
+export const getAllUsersDispatch = payload => ({
```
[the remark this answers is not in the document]
spectranaut: yeah idk I was thinking it was the "dispatch". I guess, do you "dispatch" a "dispatch"? What do you call the object that you dispatch? It doesn't "getAllUsers", it's "gotAllUsers", like "we have the users now do something with them"

client/reducers/users.js
spectranaut: ok I think I should def change the back end of "getUserConfigs" to "getUsers"

PR #36, opened 2020-04-16
Results and test queue backend

spectranaut:
Hey Erika, you or I will need to add a route to "skip" a test in the future, I didn't get to it today and figured we could add when that feature is being built out.

Otherwise, before merging to master or checking, you are going to have to manually unwind on migration. I wanted to add more data to a view! You should be able to do the following:
```
psql
> delete from flyway_schema_history where version='4';
> drop view run_data;
> exit

yarn db-migrate:dev
```

To test the new routes, you will need to populate the database with cycles and things... which makes me think maybe I should clean up the cycle initiation PR first thing tmr so you can use the front end to populate cycles.

Once there is data, you can test:
```
curl --location --request GET 'localhost:5000/api/cycle/runs' --header 'Content-Type: application/json' --data-raw '{"cycle_id": 24, "user_id": 1}'

curl --location --request POST 'localhost:5000/api/cycle/result' --header 'Content-Type: application/json' --data-raw '{"test_id": 90, "run_id": 14, "user_id": 1}'
```

### DIFF
--- a/db/migrations/V4__Add_view_for_runs.sql
+++ b/db/migrations/V4__Add_view_for_runs.sql
@@ -4,11 +4,15 @@ SELECT
   run.test_cycle_id AS test_cycle_id,
   browser_version.version AS browser_version,
   browser.name AS browser_name,
+  browser.id AS browser_id,
   at.key AS at_key,
+  at.id AS at_id,
   at_name.name AS at_name,
+  at_name.id AS at_name_id,
   at_version.version AS at_version,
   apg_example.directory AS apg_example_directory,
-  apg_example.name AS apg_example_name
+  apg_example.name AS apg_example_name,
+  apg_example.id AS apg_example_id
 FROM
   run,
   browser_version,
--- a/server/services/CycleService.js
+++ b/server/services/CycleService.js
@@ -310,7 +310,6 @@ async function getAllTestVersions() {
     }
 }
 
-
 /**
  * Saves a test result and marks the test as "complete"
  *
@@ -344,7 +343,9 @@ async function saveTestResults(testResult) {
              INSERT INTO
                test_result(test_id, run_id, user_id, status_id, result)
              VALUES
-               (${testResult.test_id}, ${testResult.run_id}, ${testResult.user_id}, ${statusId}, ${testResult.result ? testResult.result : "NULL"})
+               (${testResult.test_id}, ${testResult.run_id}, ${
+                testResult.user_id
+            }, ${statusId}, ${testResult.result ? testResult.result : 'NULL'})
              RETURNING ID
         `)
         )[0];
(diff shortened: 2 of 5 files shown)


db/migrations/V4__Add_view_for_runs.sql
```diff
@@ -4,11 +4,15 @@ SELECT
...
   at.key AS at_key,
+  at.id AS at_id,
   at_name.name AS at_name,
+  at_name.id AS at_name_id,
   at_version.version AS at_version,
   apg_example.directory AS apg_example_directory,
-  apg_example.name AS apg_example_name
+  apg_example.name AS apg_example_name,
+  apg_example.id AS apg_example_id
 FROM
```
rwaldron, line 16: @spectranaut my understanding is that all sql changes need to be done in new migrations.  @evmiguel will this migration be skipped on an existing deployment, since it already exists?

server/services/CycleService.js
```diff
@@ -294,9 +310,171 @@ async function getAllTestVersions() {
...
+        let statusId = (
+            await sequelize.query(`
+             SELECT
+               id
+             FROM
+               test_status
+             WHERE
+               test_status.name = 'complete'
+        `)
+        )[0][0].id;
```
rwaldron, line 339: `[0][0].id` seems brittle.

server/services/CycleService.js
```diff
@@ -177,6 +178,21 @@ async function getAllCycles(id) {
...
+                let users = (
+                    await sequelize.query(`
+                  select
+                    user_id
+                  from
+                    tester_to_run
+                  where
+                    tester_to_run.run_id = ${run.id}
+                `)
+                )[0];
```
rwaldron, line 191: What if there's no records returned?

server/services/CycleService.js
```diff
@@ -294,9 +310,171 @@ async function getAllTestVersions() {
...
+            await sequelize.query(`
+             INSERT INTO
+               test_result(test_id, run_id, user_id, status_id, result)
+             VALUES
+               (${testResult.test_id}, ${testResult.run_id}, ${
+                testResult.user_id
+            }, ${statusId}, ${testResult.result ? testResult.result : 'NULL'})
+             RETURNING ID
+        `)
+        )[0];
```
rwaldron, line 351: Same question here, about no records being returned

server/services/CycleService.js
```diff
@@ -294,9 +310,171 @@ async function getAllTestVersions() {
...
+                test_status.name as status
+              from
+                test_result,
+                test_status
+              where
+                test_result.status_id = test_status.id
+                and test_result.run_id = ${run.id}
+                and test_result.user_id = ${userId}
+            `)
+            )[0];
```
rwaldron, line 450: Here and above?

server/services/CycleService.js
spectranaut: As we discussed, I have a backlog task in asana to switch all this code to sequalize

PR #38, opened 2020-04-23
Tester queue

spectranaut:
Right now, this just shows all the relevant runs for a user. To test:
1. Initiate a test cycle with runs:
- Assign your self to some runs.
- Unassigned yourself from some runs that you COULD be assigned to (like, if you have VoiceOver configured, make sure that you are not assigned to all of the voice over runs).
- Make sure there are some runs you cannot perform (for example, if you don't have JAWS set, then make sure there are runs with JAWS)

Then go to the "Tester home" and select the run you just configured. Make sure that:
- the run you where assigned to list you as a tester and you can click the design pattern name
- the run that you were not assigned to don't list you and you cannot click the design pattern name
-  the runs that you cannot perform are not listed.

Things that still need to be done:
- [ ] in a follow up task, switch the state.users to an object key'd by user id
- [ ] after I finish cycle summary page, add the ability to assign or unassigned yourself from this page

### DIFF
--- a/client/reducers/login.js
+++ b/client/reducers/login.js
@@ -7,10 +7,15 @@ const initialState = {
 export default (state = initialState, action) => {
     switch (action.type) {
         case CHECK_LOGGED_IN: {
+            const { username, name, email, id, roles } = action.payload;
             return {
                 ...state,
                 isLoggedIn: true,
-                ...action.payload
+                username,
+                name,
+                email,
+                id,
+                roles
             };
         }
         case LOG_OUT: {
--- a/client/reducers/login.js
+++ b/client/reducers/login.js
@@ -7,15 +7,10 @@ const initialState = {
 export default (state = initialState, action) => {
     switch (action.type) {
         case CHECK_LOGGED_IN: {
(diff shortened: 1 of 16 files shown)


client/reducers/login.js
```diff
@@ -7,10 +7,15 @@ const initialState = {
 export default (state = initialState, action) => {
     switch (action.type) {
         case CHECK_LOGGED_IN: {
+            const { username, name, email, id, roles } = action.payload;
```
evmiguel, line 10: The `name` field doesn't exist in the payload, and it is introducing a bug

client/reducers/login.js
```diff
@@ -7,10 +7,15 @@ const initialState = {
...
             return {
                 ...state,
                 isLoggedIn: true,
-                ...action.payload
```
evmiguel, line 13: Revert to `...action.payload` to remove the bug. Otherwise, cannot save the user's ATs

client/reducers/login.js
evmiguel, line 10: The `name` field doesn't exist in the payload, and it is introducing a bug

PR #39, opened 2020-04-24
follow up on save user ats

evmiguel:
This PR addresses the comments I made in #37


client/reducers/users.js
```diff
@@ -14,25 +13,38 @@ export default (state = initialState, action) => {
...
+                    } else {
+                        updatedAts.push(at);
+                    }
+                }
+
+                let currentUserIdx = users.findIndex(
+                    user => user.id === currentUser.id
+                );
+                users[currentUserIdx].configured_ats = updatedAts;
+            }
```
spectranaut, line 44: This will be much better when we key user by id! Maybe I'll slip that refactor in a follow up to my current huge feature PR XD

PR #44, opened 2020-04-28
test run component

evmiguel:
This renders the headings for the Test Run
Screen Shot 2020-04-28 at 5 54 50 PM
Screen Shot 2020-04-28 at 5 56 13 PM

### DIFF
--- a/client/components/ManageCycles/index.jsx
+++ b/client/components/ManageCycles/index.jsx
@@ -23,7 +23,7 @@ class ManageCycles extends Component {
                     <title>Test Management | ARIA-AT</title>
                 </Helmet>
                 <h2>Initiate a Test Cycle</h2>
-                <Button as={Link} to="/new">
+                <Button as={Link} to="/cycles/new">
                     Initiate
                 </Button>
                 <h2>Test Cycle Status</h2>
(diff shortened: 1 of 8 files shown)


server/services/UsersService.js
```diff
@@ -213,7 +213,7 @@ async function getAllTesters() {
 }
```
evmiguel, line 213: There was a bug in signup. It was not propagating the user id to the `req.session.user`

client/components/ManageCycles/index.jsx
```diff
@@ -23,7 +23,7 @@ class ManageCycles extends Component {
                     <title>Test Management | ARIA-AT</title>
                 </Helmet>
                 <h2>Initiate a Test Cycle</h2>
-                <Button as={Link} to="/new">
+                <Button as={Link} to="/cycles/new">
```
spectranaut, line 26: whoops thanks for catching this!

client/components/TestQueue/index.jsx
```diff
@@ -36,7 +36,7 @@ class TestQueue extends Component {
         let designPatternLinkOrName;
         if (currentUserAssigned) {
             designPatternLinkOrName = (
-                <Link to={`/cycle/${cycleId}/run-tests/${run.id}`}>
+                <Link to={`/cycles/${cycleId}/run-tests/${run.id}`}>
```
spectranaut, line 39: we could change "run-tests" to "run" but only if you want to!

server/services/UsersService.js
spectranaut: ahh thanks for catching

PR #48, opened 2020-04-29
add more information to readme and code of conduct

boazsender:


boazsender:
@zcorpan yes, I'll file another issue. Thanks for the review folks!

PR #50, opened 2020-04-29
Everytime a user logs in, set their roles according to the github teams

spectranaut:
To test you will be adding and removing yourself from the following teams:
- [admin team](https://github.com/orgs/bocoup/teams/aria-at-report-admin/members)
- [tester team](https://github.com/orgs/bocoup/teams/aria-at-report-testers/members)

1. Add yourself to the tester team, but remove yourself from the admin team
    - log out
    - log back in
    - you should see the "test queue" but not the "test cycle management"
    - there should be no entry for you in the `user_to_role` table where `role_id `is the admin role.
    - you cannot go to pages "/cycles/new" or "/cycles" -- you get redirected to 404 page

2. Add yourself to the admin team and remove yourself from the tester team.
    - log out
    - log back in
    - you should see the "test queue" and the "test cycle management" (an admin is sort of like a tester by default)
    - there should be no entry for you in the `user_to_role` table where `role_id `is the tester role.

3. Remove yourself from both admin and tester team
   - log out
   - log back in
   - you should see no options, just a log out option

### DIFF
--- a/client/components/App/index.jsx
+++ b/client/components/App/index.jsx
@@ -28,7 +28,7 @@ class App extends Component {
     }
 
     render() {
-        const { route, isLoggedIn } = this.props;
+        const { route, isLoggedIn, isAdmin, isTester } = this.props;
         return (
             <Fragment>
                 <Container fluid>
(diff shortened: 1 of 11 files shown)


client/components/ManageCycles/index.jsx
```diff
@@ -16,7 +17,12 @@ class ManageCycles extends Component {
     }
 
     render() {
-        const { cyclesById } = this.props;
+        const { cyclesById, isAdmin } = this.props;
+
+        if (!isAdmin) {
+            return <Redirect to={{ pathname: '/'}} />;
```
spectranaut, line 23: We should make a 404 page and redirect here, and block all pages based on access control. I was thinking there ought to be better place to do this but the way we define routes... I couldn't think of a clean way.

spectranaut: I added 404 redirects

isaacdurazo:
@spectranaut I've reviewed this PR and everything works as expected. The only thing I couldn't verify because I didn't know where I should look is:

> there should be no entry for you in the `user_to_role` table where `role_id` is the admin role.

client/components/App/index.jsx
```diff
@@ -82,13 +86,17 @@ class App extends Component {
...
     location: PropTypes.object,
     route: PropTypes.object
 };
 
 const mapStateToProps = state => {
-    const { isLoggedIn } = state.login;
-    return { isLoggedIn };
+    const { isLoggedIn, roles } = state.login;
+    let isAdmin = roles ? roles.includes('admin') : false;
+    let isTester = isAdmin || (roles && roles.includes('tester'));
```
evmiguel, line 98: I like this!

PR #53, opened 2020-05-01
Add ability to assign users in test queue

spectranaut:
Hey Erika, this PR does three things:
1. Makes users key'd by ID. This caused me to changes some of the user setting logic, because it was all based on username before. Also, it led me to find and fix that bug where sometimes all the ATs disappear on the user setting page! :)
2. Adds ability to assign and unassign yourself from runs in the user queue

### DIFF
--- a/client/components/TestQueueRun/index.jsx
+++ b/client/components/TestQueueRun/index.jsx
@@ -1,10 +1,9 @@
-import React, { Component, Fragment } from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Button } from 'react-bootstrap';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
 import { deleteUsersFromRun, saveUsersToRuns } from '../../actions/cycles';
-
 
 class TestQueueRow extends Component {
     constructor(props) {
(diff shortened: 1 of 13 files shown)


rwaldron:
@spectranaut is the reason that you're switching from "array of users" to "object of users by userId key" solely because it may be a faster lookup?

client/reducers/users.js
```diff
@@ -1,50 +1,47 @@
...
 export default (state = initialState, action) => {
     switch (action.type) {
         case GET_USERS: {
             const users = action.payload;
-            return Object.assign({}, state, {
-                users
+            const usersById = {};
+            users.map(u => {
+                usersById[u.id] = u;
             });
```
rwaldron, line 14:
```js
const usersById = users.reduce((accum, user) => {
  accum[u.id] = u;
  return accum;
}, {});
```

You're not mapping an array's values to a new array, you're reducing an array to object.

spectranaut:
> @spectranaut is the reason that you're switching from "array of users" to "object of users by userId key" solely because it may be a faster lookup?

That's right -- and it makes the code cleaner when you are getting user data.

client/components/TestQueueRun/index.jsx
```diff
@@ -0,0 +1,83 @@
+import React, { Component } from 'react';
```
evmiguel, line 1: This component is super clean and well written!

PR #57, opened 2020-05-05
Save results during cycle run

spectranaut:


client/components/TestRun/index.jsx
```diff
@@ -33,6 +39,42 @@ class TestRun extends Component {
...
+        let resultsEl = this.testIframe.current.contentDocument.querySelector(
+            '#__ariaatharness__results__'
+        );
+
+        if (!resultsEl) {
+            // TODO: show popup here that says "You cannot save results until you complete and review the test"
+            return;
+        }
+
+        const result = JSON.parse(resultsEl.innerHTML);
```
rwaldron, line 63: Is it ever possible that the contents can be malformed? Should this be wrapped in a try/catch? Also, can you do `resultsEl.innerText`?

client/components/TestRun/index.jsx
```diff
@@ -90,19 +133,46 @@ class TestRun extends Component {
...
+                } else {
+                    content = (
+                        <React.Fragment>
+                            <div>
+                                RESULTS EXIST!!!!!! TODO: Make a view of the
+                                existng tests
+                            </div>
+                            <Button
+                                variant="primary"
+                                onClick={this.handleNextTestClick}
```
rwaldron, line 169: Missing aria-label?

client/components/TestRun/index.jsx
```diff
@@ -90,19 +133,46 @@ class TestRun extends Component {
...
+                                onClick={this.handleSkipTestClick}
+                            >
+                                Skip to next test
+                            </Button>
+                            <Button
+                                variant="primary"
+                                onClick={this.handleSaveResultClick}
+                            >
+                                Save results and go to next test
+                            </Button>
```
rwaldron, line 157: Do all of these buttons need an aria-label?

client/reducers/cycles.js
```diff
@@ -50,6 +51,31 @@ export default (state = initialState, action) => {
                 }
             });
         }
+        case SAVE_RESULT: {
+            const result = action.payload;
+
+            const tests =
+                state.runsForCycle[result.cycle_id][result.run_id].tests;
```
rwaldron, line 58: Is it guaranteed that there will always be `result.cycle_id` and `result.run_id`? Similarly, is it safe to get of `state.runsForCycle[...][...].tests;`?

client/reducers/cycles.js
```diff
@@ -50,6 +51,31 @@ export default (state = initialState, action) => {
...
+            const testIndex = tests.findIndex(t => {
+                return t.id === result.test_id;
+            });
```
rwaldron, line 61: Nit: 

```
const testIndex = tests.findIndex(t => t.id === result.test_id);
```

server/services/CycleService.js
```diff
@@ -349,31 +349,78 @@ async function getAllTestVersions() {
...
-        `)
-        )[0][0].id;
+               test_result.user_id = ${user_id}
+               AND test_result.run_id = ${run_id}
+               AND test_result.test_id = ${test_id}
+        `);
 
-        let testResultId = (
+        if (resultRows[0].length) {
+            testResultId = resultRows[0][0].id;
```
rwaldron, line 367: This is brittle

server/services/CycleService.js
```diff
@@ -349,31 +349,78 @@ async function getAllTestVersions() {
...
+            statusId = (
+                await sequelize.query(`
+                 SELECT
+                   id
+                 FROM
+                   test_status
+                 WHERE
+                   test_status.name = 'complete'
+            `)
+            )[0][0].id;
```
rwaldron, line 381: Same issue as above

client/components/TestRun/index.jsx
spectranaut: it's possible, but that would be a bug within ARIA-AT repo that would prevent us from being able to consume/run the tests. but I can guard anyway, we should know if that bug exists.

client/components/TestRun/index.jsx
spectranaut: these buttons don't need labels because their inner text is descriptive enough! Only when there is additional context about the button based on position do we add aria-label specifically

client/components/TestRun/index.jsx
spectranaut: same as other button

client/reducers/cycles.js
spectranaut: yes all of that data is definitely there if you are in the middle of a test run.

PR #59, opened 2020-05-06
Update README to provide better info node version

rmeritz:
- Also fix the link to the DB
- I was going to change the DB documentation about flyway to, but upon
re-reading I see it was really reader error and there was nothing wrong
with the docs.

### DIFF
--- a/docs/Local Development.md
+++ b/docs/Local Development.md
@@ -2,7 +2,8 @@
 
 ## Dependencies
 - Node
-  - Version 10.13.0 or greater
+  - Version >= 10.13.0 and < 13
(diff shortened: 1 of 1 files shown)


docs/Local Development.md
```diff
@@ -22,7 +23,7 @@ yarn install
 
 ### Set up local database
 
-Follow the instructions provided in [Database.md](https://github.com/bocoup/aria-at-report/blob/master/docs/Database.md).
+Follow the instructions provided in [Database.md](https://github.com/bocoup/aria-at-report/blob/main/docs/Database.md).
```
rwaldron, line 26: 👍

PR #60, opened 2020-05-06
Add popups for the "save and exit", "save and next test" and "skip to next test" buttons

spectranaut:
To test:
Start a test run and don't complete the first test:
- [x] click the "save and exit" button, see that a modal appears and it won't let you save and exit
- [x] click the "save and next test" button, see that a modal appears and it won't let you save and exit
- [x] click "skip and go to next test", see that a modal appears, make sure all the buttons do what you would expect

Complete the first test:
- [x] click the "save and exit" button, and you will be redirected to the tester queue for that cycle
- [x] click the "save and next test" button, and you will see the next test

### DIFF
--- a/client/components/App/index.jsx
+++ b/client/components/App/index.jsx
@@ -78,9 +78,9 @@ class App extends Component {
                         </Navbar.Collapse>
                     </Navbar>
                 </Container>
-                <main>
+                <Container as="main" fluid>
                     {renderRoutes(route.routes)}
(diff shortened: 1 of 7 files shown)


rwaldron:
I can't get to any of these parts. There are no ATs available for me to save:

image

rwaldron:
Once I get a test running, I don't see any "save and exit" buttons

PR #63, opened 2020-05-07
TestRun: layout

rwaldron:
image

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -1,10 +1,17 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import './TestRun.css';
 import { connect } from 'react-redux';
 import { Helmet } from 'react-helmet';
-import Button from 'react-bootstrap/Button';
-import Modal from 'react-bootstrap/Modal';
+import {
+    Button,
+    ButtonGroup,
+    ButtonToolbar,
+    Col,
+    Container,
+    Modal,
+    Row
+} from 'react-bootstrap';
 import {
     getTestCycles,
     getRunsForUserAndCycle,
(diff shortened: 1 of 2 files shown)


client/components/TestRun/index.jsx
```diff
@@ -231,47 +238,69 @@ class TestRun extends Component {
...
+                            src={`/aria-at/${git_hash}/${
+                                tests[this.state.currentTestIndex - 1].file
+                            }?at=${at_key}`}
+                            id="test-iframe"
+                            ref={this.testIframe}
+                        ></iframe>
+                    );
+
+                    menuUnderIframe = (
+                        <ButtonToolbar className="testrun__button-toolbar--margin">
```
spectranaut, line 272: I wonder if we should give these some kind of landmark region or at least an accessible name to the group.

Make just "region" -- the point being, I want it to be easy for a screen reader user to know those buttons are there and skip to them if partially thru a test..? Regions are all about skipping to area of the web page. I don't know what to label the two different button groups -- or if we make them into "regions", what to call the regions.

client/components/TestRun/index.jsx
```diff
@@ -295,41 +324,54 @@ class TestRun extends Component {
...
-                <Helmet>
-                    <title>{`Testing ${apg_example_name} for ${at_name} ${at_version} with ${browser_name} ${browser_version} | ARIA-AT`}</title>
-                </Helmet>
-                {heading}
-                {content}
+            <Fragment>
+                <Container fluid>
+                    <Row>
+                        <Col>
+                            <Helmet>
```
spectranaut, line 350: I think the Helmet doesn't have to be in a row and col, it's just for changing the title element of the webpage.

client/components/TestRun/index.jsx
zcorpan: I'd need to see the generated HTML (or DOM) of the whole page to evaluate this. But a few buttons doesn't necessarily require a landmark.

spectranaut: ok ignore this concern @rwaldron

PR #64, opened 2020-05-08
View results

spectranaut:
This PR does a lot of things because I wanted to get in some bigger changes early on Friday so that it could be code reviewed and tests. I'm sorry it's so much at once!

1. It adds a read only view of the tests
2. It simplifies the buttons, per Matts request. There is no longer a difference between "save and go to next" and "skip and go to next test". There is now just a "next test" button.
3. When editing the code to change the buttons around, the code became nightmarish, and I have a feeling this is not the "final state" of the buttons/actions on the test run page. So I did a minor refactor to break part of the page out into it's own component.

To test this PR:
- Start a test run
- Try all the buttons except for the "issue" button and the "redo test"button
- Fill in all the data for a test and click "review test" in the iframe, then try click "next or "previous", then go back to that test and see the read only view.

### DIFF
--- a/client/components/TestResult/index.jsx
+++ b/client/components/TestResult/index.jsx
@@ -1,0 +1,89 @@
+import React, { Component, Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { Table } from 'react-bootstrap';
+
+class TestResult extends Component {
+    renderPassingAssertions(assertions) {
+        if (assertions.length === 0) {
+            return (
+                <ul><li>No passing assertions.</li></ul>
+            );
+        }
+        return (
+            <ul>
+              {assertions.map((a, index) => <li key={index}>{`${a.pass}: ${a.assertion}`}</li>)}
+            </ul>
+        );
+    }
+
+    renderFailingAssertions(assertions) {
+        if (assertions.length === 0) {
+            return (
+                <ul><li>No failing assertions.</li></ul>
+            );
+        }
+        return (
+            <ul>
+              {assertions.map((a, index) => <li key={index}>{`${a.fail}: ${a.assertion}`}</li>)}
+            </ul>
+        );
+    }
+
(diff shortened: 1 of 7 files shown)


client/components/TestResult/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+            );
+        }
+        return (
+            <ul>
+                {unexpecteds.map((u, index) => (
+                    <li key={index}>{u}</li>
+                ))}
+            </ul>
+        );
+    }
```
rwaldron, line 55: All three of these could be reduced to a single function: 

```js
    renderOutcomeList(outcomes) {
        if (!outcomes.length) {
            outcomes.push('None.');
        }
        return (
            <ul>
                {outcomes.map((outcome, index) => (
                    <li key={useId()}>{outcome}</li>
                ))}
            </ul>
        );
    }
```
(Note that `useId()` assumes that you use the module I recommended above)

Then your code below becomes: 

```js
let passing = command.assertions.reduce((accum, a) => 
    accum.concat(a.pass !== undefined ? [`${a.pass}: ${a.assertion}`] : []), []
);
let failing = command.assertions.reduce((accum, a) => 
    accum.concat(a.fail !== undefined ? [`${a.fail}: ${a.assertion}`] : []), []
);

return (
    <tr key={useId()}>
        <td>{command.command}</td>
        <td>{command.support}</td>
        <td>
            <p>Output: {command.output}</p>
            <p>Passing Assertions:</p>
            {this.renderOutcomeList(
                passing
            )}
            <p>Failing Assertions:</p>{' '}
            {this.renderOutcomeList(
                failing
            )}
            <p>Unexpected Behaviors:</p>{' '}
            {this.renderOutcomeList(
                command.unexpected_behaviors
            )}
        </td>
    </tr>
);
```

PR #65, opened 2020-05-08
Add redo and github link

spectranaut:
Now to test, start a test run:
- Fill in a test and save the results (you can save by clicking "next" after the results are filled in and "review results" has been checked)
- Click next and previous, refresh, see that the results are always there
- Click "redo", and the read only view should turn back into an iframe
- Click next and previous, refresh, make sure that the results are always displayed as an iframe now
- NOW, refill in that test and save, maybe sure the save persists

### DIFF
--- a/client/components/DisplayTest/index.jsx
+++ b/client/components/DisplayTest/index.jsx
@@ -53,7 +53,7 @@ class DisplayTest extends Component {
     handleRedoClick() {
         const { deleteResultFromTest, test } = this.props;
 
-        if (!test.result || test.result.status !== 'complete') {
+        if (!this.testHasResult) {
             document
                 .getElementById('test-iframe')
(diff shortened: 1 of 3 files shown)


client/components/DisplayTest/index.jsx
```diff
@@ -46,6 +50,19 @@ class DisplayTest extends Component {
         this.performButtonAction();
     }
 
+    handleRedoClick() {
+        const { deleteResultFromTest, test } = this.props;
+
+        if (!test.result || test.result.status !== 'complete') {
+            document
+                .getElementById('test-iframe')
+                .contentWindow.location.reload();
```
rwaldron, line 59: If, for any reason, this iframe is not in the dom, then this is going to blow up.

PR #68, opened 2020-05-08
Add user status

spectranaut:
Complete some tests and confirm on the "Test Run Queue for Cycle..." page you see an update of the status of the user:

status

### DIFF
--- a/client/components/TestQueue/index.jsx
+++ b/client/components/TestQueue/index.jsx
@@ -33,15 +33,15 @@ class TestQueue extends Component {
     renderAtBrowserList(runIds) {
         const { userId, usersById, cycle, testsForRuns } = this.props;
         const {
-            at_name,
-            at_version,
-            browser_name,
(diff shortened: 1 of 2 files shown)


client/components/TestQueue/index.jsx
```diff
@@ -52,21 +52,25 @@ class TestQueue extends Component {
...
-                                testers={cycle.runsById[runId].testers}
-                                cycleId={cycle.id}
-                                usersById={usersById}
-                                userId={userId}
-                                atName={at_name}
-                                browserName={browser_name}
-                            />
-                        ))}
+                        {runIds.map(runId => {
+                            let testsForRun = testsForRuns[runId].tests;
```
rwaldron, line 56: You can remove this and just use `testsForRuns[runId].tests` below

spectranaut: very reasonable

PR #71, opened 2020-05-09
Tolerate login attempts from unrecognized users

jugglinmike:
Because the `getUser` function may return `false`, the value of `dbUser`
should not be assumed to be an object.


rwaldron:
@spectranaut Was the original behavior intentional?

spectranaut:
Probably the getUser function shouldn't return false. If you log in and fail, this code should execute exactly the same as if you haven't logged in. I appreciate the suggested fix, Mike, but I think I'm going to make a backlog task to fix the greater issue!

PR #74, opened 2020-05-12
Update node version in deploy

spectranaut:


spectranaut:
omg, the error is unrelated, and a lint error.

PR #76, opened 2020-05-13
Don't load environment variables with xargs for dev server

rmeritz:
When the variables are loaded with xargs any values with spaces in them
get split so`export $(cat config/dev.env | xargs)` produces `GITHUB_TEAM_QUERY=ARIA-AT Report GITHUB_TEAM_TESTER=ARIA-AT Report Testers GITHUB_TEAM_ADMIN=ARIA-AT Report Admin`

Which means that those env vars end up looking like:
```
process.env.GITHUB_TEAM_QUERY=ARIA-AT 
process.env.GITHUB_TEAM_TESTER=ARIA-AT
process.env.GITHUB_TEAM_ADMIN=ARIA-AT
```

This was causing the login to fail because authentication wasn't working correctly.

In the server, the environment is already loaded with `dotenv`. So we don't need to use `export` at all.

The `export` was just for the `client` now use `webpack-dotenv` for the client which is like `dotenv` but `webpack` specific. It works with the dev server and doesn't leak unused variables to the client.

https://www.npmjs.com/package/dotenv-webpack

This PR only fixes the immediately broken login in the development environment problem. But the (potential) problem with using `xargs` is in several places throughout the app:
- Importing tests in dev via the `yarn run db-import-tests:dev`
- Running anything in vagrant with the `/deploy/scripts/export-and-exec.sh` script currently used to build the frontend packages and import the tests

I'm working on a fix to switch all of these to `dotenv` so this problem doesn't creep up in other places its non-trivial to test that fix so I thought I would open this PR separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "private": true,
   "scripts": {
     "a11y": "yarn workspace client lighthouse",
-    "dev": "export $(cat config/dev.env | xargs); npm-run-all --parallel dev:*",
+    "dev": "npm-run-all --parallel dev:*",
     "dev:client": "yarn workspace client run dev",
     "dev:server": "yarn workspace server run dev",
     "jest": "yarn workspaces run jest",
--- a/client/package.json
+++ b/client/package.json
@@ -24,6 +24,7 @@
   "homepage": "https://github.com/bocoup/aria-at-report#readme",
   "dependencies": {
     "bootstrap": "^4.4.1",
+    "dotenv-webpack": "^1.8.0",
     "node-sass": "^4.13.1",
     "react-bootstrap": "^1.0.0",
     "react-helmet": "^6.0.0",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "private": true,
   "scripts": {
     "a11y": "yarn workspace client lighthouse",
-    "dev": "export $(cat config/dev.env | xargs); npm-run-all --parallel dev:*",
```
spectranaut, line 9: this data is actually needed for the server, not the client, for what it's worth. this webpack file is for the client? I'm not sure how yarn does this under the hood, but it's probably the same env for both the client and serve process.

rmeritz: It is already loaded by the server in `yarn workspace server dev` by `dotenv` this was just overwriting it with the wrong variables.

rwaldron:
Excellent work!

PR #77, opened 2020-05-13
Load environment in import-tests without xargs in local enviornment

rmeritz:
Use dotenv to load enviornment instead of xargs. Xargs is more brittle.

Also, maybe more importantly the `yarn run db-import-tests:dev` was broken since it was switched to script file. This also fixes that issue so it now can be run locally and not just for deployment.

### DIFF
--- a/deploy/roles/application/tasks/main.yml
+++ b/deploy/roles/application/tasks/main.yml
@@ -44,7 +44,7 @@
     chdir: '{{source_dir}}'
 
 - name: Import tests into database
-  command: DOTENV_CONFIG_PATH={{environment_config.dest}} node -r dotenv/config ./scripts/import-tests/index.js -c {{aria_at_test_revision}}
+  command: ./deploy/scripts/export-and-exec.sh {{environment_config.dest}} node ./server/scripts/import-tests/index.js -c {{aria_at_test_revision}}
   args:
     chdir: '{{source_dir}}'
 
--- a/deploy/roles/application/tasks/main.yml
+++ b/deploy/roles/application/tasks/main.yml
@@ -44,7 +44,7 @@
     chdir: '{{source_dir}}'
 
 - name: Import tests into database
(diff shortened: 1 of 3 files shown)


rmeritz:
Both @rwaldron & @spectranaut said they cannot get it working... So it will try again in a moment.

rmeritz:
I need to verify I can deploy... And that the data data is truly imported.

rwaldron:
Verified locally: 

```
rwaldron in ~/clonez/aria-at-report on remove-xargs-rmeritz
$ yarn db-import-tests:dev  -c ff1ed8a347dfc5ca7be47fa8620cfac681af0439
yarn run v1.22.4
$ yarn workspace server db-import-tests:dev -c ff1ed8a347dfc5ca7be47fa8620cfac681af0439
$ DOTENV_CONFIG_PATH=../config/dev.env node -r dotenv/config ./scripts/import-tests/index.js -c ff1ed8a347dfc5ca7be47fa8620cfac681af0439
Cloned aria-at.git to /Users/rwaldron/clonez/aria-at-report/server/scripts/import-tests/tmp/
Test version information exists for: https://github.com/w3c/aria-at.git Wed May 06 2020 15:55:54 GMT-0400 (Eastern Daylight Time)
Done, no errors
✨  Done in 3.13s.
```

rmeritz:
@rwaldron - I was able to verify it locally as well. But I wasn't able to figure out how to get it working in `vagrant` and `ansible` so I removed the change to the ansible scripts but this change just fixes the script that was broken (See the second commit).

I still want to be able to totally remove the use of `/deploy/scripts/export-and-exec.sh` script which calls `xargs` because it might cause other issues. But I'm missing understanding about `ansible` or `vagrant` to get past where I'm stuck in my debugging. 

But as far as I know, there is nothing breaking with the use of it currently.

PR #81, opened 2020-05-13
Accessibility fixes requested by Matt

spectranaut:
This does three things:
1. The nav link for the current page now has `aria-current="page"` set, and when you click a click, the focus is put on the nav item so the SR user hears the `aria-current` change. This is to better handle the "single-paged-ness" of the app.
2. All the tables have accessible names
3. A confusing phrase updated to be more wordy

### DIFF
--- a/client/components/App/index.jsx
+++ b/client/components/App/index.jsx
@@ -14,21 +14,47 @@ class App extends Component {
         super(props);
 
         this.signOut = this.signOut.bind(this);
+        this.currentNavItem = React.createRef();
     }
+
     async componentDidMount() {
         const { dispatch } = this.props;
 
-        await dispatch(handleCheckSignedIn());
-        await dispatch(handleGetValidAts());
-        await dispatch(getAllUsers());
+        dispatch(handleCheckSignedIn());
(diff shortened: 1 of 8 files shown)


client/components/ConfigureRunsForExample/index.jsx
```diff
@@ -94,11 +95,14 @@ class ConfigureRunsForExample extends Component {
             return b.at_id - a.at_id;
         });
 
+        let exampleTableTitle = example.name || example.directory;
+        let tableId = `table_name_${nextId()}`;
```
rwaldron: `nextId()` accepts a prefix argument, so you can just use it as `key={nextId('table_name_')}`

client/components/App/index.jsx
```diff
@@ -35,11 +35,11 @@ class App extends Component {
     }
 
     async signOut() {
-        const { dispatch, location } = this.props;
+        const { dispatch } = this.props;
         await dispatch(handleSignout());
         // Avoid the flash of "logged in user" after
         // pressing "log out"
-        location.href = '/';
```
spectranaut, line 31: Whoops sorry!

PR #86, opened 2020-05-14
Refactor: All database Tables have a matching Sequelize model

rmeritz:
This PR changes no actual functionality in the app it changes the setup of Sequelize to be more conventional and it adds models for all db tables which will allow for future features to use Sequelize more easily as a full-featured ORM and to generate new models via the cli.

## Refactoring Done:
- Installed the [sequelize cli](https://github.com/sequelize/cli)
- Ran the `init` to generate the `model/index.js` file which now loads all models and provides a `db` object provides all database access
- Used the tool [sequalize-auto](https://github.com/sequelize/sequelize-auto) to generate models from the existing tables
- Switched all database access over to the new `db` and removed all old models
- Updated the mocks/test so that that the new `db` object gets mocked
- Added a `.sequelizerc` and tested that new models can now be generated successfully via the cli

## Additional Notes:
- I didn't touch any of the naming and I don't think it is worthwhile to do so at this point, but I would recommend using Sequalize conventions moving forward
- I didn't touch the migrations yet. They are all still in flyway. But this change is completely independent of that work which I will look into more next. (I did look into this a little and it doesn't seem super straightforward but should be possible).
- I didn't touch any of the existing raw SQL, but the PR sets things up so moving forward the models can be used more easily and the old raw SQL can be refactored whenever convenient.
- I didn't define all the associations that are part of the models but when they are needed to be used in future queries they can easily be added via an `associations` attribute. See `Users.js` for an example.
- It was unclear to me the conventions of the project for `At` vs `AT` and for file name capitalization as it seemed to be mixed.
- I tried to test the functioning of the app manually as well as I could and I believe that everything is working. However, I'm not very familiar with the app and this is a sweeping change that could break anything so it is probably worth a very through QA. Please let me know about any hidden parts of the app I might have missed.


rwaldron:
This is exceptional refactoring work—thanks for taking this on!

rwaldron:
I just noticed this is change fixes the oddball `Error: TypeError: sequelize.query is not a function` error that I kept getting when running the tests locally.

server/models/FlywaySchemaHistory.js
```diff
@@ -0,0 +1,53 @@
+module.exports = function(sequelize, DataTypes) {
```
evmiguel, line 1: @rmeritz Is this model something to discard when we move away from Flyway?

server/models/index.js
```diff
@@ -0,0 +1,46 @@
...
+fs.readdirSync(__dirname)
+    .filter(file => {
+        return (
+            file.indexOf('.') !== 0 &&
+            file !== basename &&
+            file.slice(-3) === '.js'
+        );
+    })
+    .forEach(file => {
+        const model = sequelize['import'](path.join(__dirname, file));
```
evmiguel, line 33: This is slick!

server/models/index.js
```diff
@@ -0,0 +1,46 @@
+const fs = require('fs');
```
evmiguel, line 1: I love this file!

server/services/ATService.js
```diff
@@ -1,8 +1,8 @@
-const { ATModel } = require('../models/ATModel');
+const db = require('../models/index');
 
 async function getATs() {
     try {
-        const ats = await ATModel.findAll();
+        const ats = await db.AtName.findAll();
```
evmiguel, line 5: This is super understandable! I love the new readability with your changes!

server/services/UsersService.js
```diff
@@ -1,18 +1,10 @@
-const {
-    Users,
-    UserToRole,
-    Role,
-    TesterToRun,
-    UserToAt
-} = require('../models/UsersModel');
+const db = require('../models/index');
```
evmiguel, line 1: So fresh! So clean!

server/tests/integration/server.int.test.js
```diff
@@ -11,8 +11,7 @@ const newUserToRole = require('../mock-data/newUserToRole.json');
 const listOfATs = require('../mock-data/listOfATs.json');
 
 // mock the models to circumvent the actual database
-jest.mock('../../models/UsersModel');
-jest.mock('../../models/ATModel');
+jest.mock('../../models/index.js');
```
evmiguel, line 14: Nice!

server/jest.config.js
```diff
@@ -1,8 +1 @@
-module.exports = {
-    globals: {
-        sequelize: {
-            define: obj => obj,
-            query: () => [[{ at_name_id: 1, at_name: 'Foo', active: true }]]
-        }
-    }
-};
+module.exports = {};
```
spectranaut, line 1: do we need this file anymore, @evmiguel ?

evmiguel: Not anymore!

PR #88, opened 2020-05-18
[Fix] Test Run Queue: Page is blank if user has no ATs configured.

rwaldron:


client/components/TestQueue/index.jsx
```diff
@@ -87,15 +88,33 @@ class TestQueue extends Component {
...
         if (!testsFetched || !cycle || !Object.keys(usersById).length) {
-            return <div>Loading</div>;
+            return <div data-test="test-queue-loading">Loading</div>;
         }
 
-        let currentUser = usersById[userId];
-        let configuredAtNames = currentUser.configured_ats
+        const currentUser = usersById[userId];
+
+        if (!currentUser.configured_ats.length) {
```
spectranaut, line 96: So this fix doesn't work right now -- each configured_at sent from the back end an "active" flag, and to tell if an user has an AT configured you need to check whether there is anything int the list and whether the "active" flag is set. The reason for this is that the database many-to-many table has an "active" column (the argument was to prevent constantly deleting/adding new rows) and the back end just sends this data up instead of doing filtering on this column, which it should.

I can do that change to the backend today, because it makes sense with the work I'm doing, then we can rebase this commit and land it.

(To test it not working -- configure all ATs, then unselect all and save, and you will get a blank page)

spectranaut:
@rwaldron if you want to rebase this on main, I'll test and merge

rwaldron:
@spectranaut rebased and tested, works as expected

PR #93, opened 2020-05-19
Test queue -- admin view + assign users

spectranaut:
To test this PR, you probably will want to add another user to the database. You can add a user like this:
```
insert into users (username) values ('foobar');
insert into user_to_at (user_id, at_name_id, active) values (${userID}, 1, true);
```

Before testing, you will also want to make sure not all AT's are selected in your user settings and to initiate a cycle with runs for all possible ATs.

Now, first, make sure you are NOT a in the [admin group](https://github.com/orgs/bocoup/teams/aria-at-report-admin/members), and log out and back in again:
- Navigate to the cycle by navigating to the "test queue" page and clicking "contribute tests" for that cycle.
- You should only see the runs you can contribute to.
- You should be able to assign and unassign yourself to all runs
- Focus should move to the list of users after you assign/unassign yourself

Now, make it so that you ARE in the admin group:
- Navigate to the cycle by navigating to the "test queue" page and clicking "contribute tests" for that cycle.
- You should see all runs!!
- You should be able to assign and unassign users to runs, based on their user preferences
- If you can assign or unassign no more users to run, the drop down button should be disabled
- Focus should move to the list of users after you assign/unassign users

### DIFF
--- a/client/components/TestQueueRun/index.jsx
+++ b/client/components/TestQueueRun/index.jsx
@@ -1,7 +1,7 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import nextId from 'react-id-generator';
-import { Button } from 'react-bootstrap';
+import { Button, Dropdown } from 'react-bootstrap';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
 import { deleteUsersFromRun, saveUsersToRuns } from '../../actions/cycles';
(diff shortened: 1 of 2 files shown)


client/components/TestQueueRun/index.jsx
```diff
@@ -35,91 +59,192 @@ class TestQueueRow extends Component {
...
-        } else {
-            designPatternLinkOrName = apgExampleName;
+        let testrun = `${apgExampleName} for ${atName} on ${browserName}`;
+
+        let canUnassignTesters = [];
+        let canAssignTesters = [];
+        for (let uid of Object.keys(usersById)) {
+            const tester = usersById[uid];
+
+            if (testers.includes(tester.id)) {
```
evmiguel, line 85: I found this to be pretty readable!

client/components/TestQueueRun/index.jsx
```diff
@@ -35,91 +59,192 @@ class TestQueueRow extends Component {
...
+                <Link to={`/cycles/${cycleId}/run/${runId}`}>
+                    {apgExampleName}
+                </Link>
+            );
+        } else {
+            designPatternLinkOrName = apgExampleName;
+        }
+
+        let assignOptions;
+        if (admin) {
```
evmiguel, line 207: Nonblocking: I like what you did here! Maybe in the long term, we can separate these in the separate components for admin and tester

spectranaut: yeah I was also thinking about that, I have more work to do on this component so I might separate them out in the next couple days

PR #94, opened 2020-05-19
Executing Test: Tester View: Test Conflict Resolution, Pt 1

rwaldron:
- Run `yarn db-migrate:dev`
- Run `yarn`
- Run `yarn dev`
- Note that when running from dev configuration, your issues will be created here: https://github.com/bocoup/aria-at/issues.
- Open the app at http://localhost:3000
- Sign out of the app
- In another tab, go here: https://github.com/settings/applications, and revoke your authentication for "ARIA-AT Report API (Development)"
- Return to the app and click "Sign in with Github"
- Authorize
- Once back at the app, make sure you have some ATs set up. 
- Go to http://localhost:3000/cycles, click "Initiate"
- Create a "Test Cycle Configuration"
- Check the checkboxes and assign yourself to each of the 
Test Plans"
- Click "Initiate Cycle"
- Go to http://localhost:3000/test-queue
- Click "Contribute Tests" 
- Click the name of any "Test Plan". If none are clickable, click "Assign Me" then click the name.
- You're now in the test and can actually start testing this feature 🙄 

1. Observe that there are no Status notifications on the page. 
1. Click "Raise an issue", observe that you are now at the "Create an issue" prompt. 
1. Observe there is a form and two buttons: "Cancel" and "Submit new issue"
1. Click "Cancel". Observe that the prompt closes. 
1. Click "Raise an issue", observe that you are now at the "Create an issue" prompt. 
1. Make any changes you want and click "Submit new issue"
1. Observe that you are now at the "Issue created" view and are presented with a message that says "Issue #XX created" where "XX" is the issue number that was created. Click "Done", observe that the prompt closes. 
1. Click "Raise an issue", observe that you are now at the "Review existing issues
" prompt, and shown a list that has your issue in it. 
1. Click "My issue exists in this list", observe that the prompt closes. 
1. Click "Raise an issue"
1. Click "My issue is not in this list", observe that you are at the "Create an issue" prompt. 
1. Click "Back to issues list", observe that you are at the "Review existing issues
" prompt.
1. Click the "X" in the upper right of the prompt. 
1. Refresh the test page
1. Observe that there is now a "status bar" that says "This test has open issues"
1. Click the "Review issues" button, observe that you are now at the "Review existing issues" prompt.
1. Scroll to the "See on github.com" link and click it.
1. Close the issue on Github
1. Return to the ARIA-AT test page and refresh. 
1. Observe that the status bar is gone. 
1. Click "Raise an issue", observe that you are now at the "Create an issue" prompt.


spectranaut:
I reported an issue on the first test, and I see it in the bocoup/aria-at fork. Then I went to the next test, and the next test, and they all have "This test has an open issue banner", and if you click on it, you see this:

no_issues

client/components/DisplayTest/index.jsx
```diff
@@ -160,36 +178,74 @@ class DisplayTest extends Component {
...
-                    </Button>
-                    <Button
-                        variant="secondary"
-                        onClick={this.handleModalConfirmClick}
-                    >
-                        Continue
-                    </Button>
-                </Modal.Footer>
-            </Modal>
+            <>
```
spectranaut, line 181: omg, what is this? replacement for fragment?

client/actions/cycles.js
```diff
@@ -116,3 +140,73 @@ export function deleteUsersFromRun(users, runId, cycleId) {
...
+//       be completed in a follow up changeset.
+//
+
+/**
+ * getConflictsByCycleId        Returns an array of conflicts for a
+ *                              test in a given cycle.
+ * @param  {Number} cycle_id    The id of the cycle to retrieve conflicts
+ * @return {Array}              An array of conflict objects.
+ */
+export function getConflictsByCycleId(cycle_id) {
```
spectranaut, line 184: We actually don't need this. All the results for a cycle are already in the front end -- so we can do the conflict calculation on the front end.

As it is right now, having all test results for all test for a cycle when you contribute to a cycle -- might seems a little heavy handed.. but for now it is not too much data. we might need to later serve things in smaller pieces.

client/actions/cycles.js
```diff
@@ -116,3 +140,73 @@ export function deleteUsersFromRun(users, runId, cycleId) {
...
+    };
+}
+
+/**
+ * getResultsByCycleId          Returns an array of results for a
+ *                              test in a given cycle.
+ * @param  {Number} cycle_id    The id of the cycle to retrieve results
+ * @return {Array}              An array of result objects.
+ */
+export function getResultsByCycleId(cycle_id) {
```
spectranaut, line 202: Same as above, we have all this data.

client/components/StatusBar/index.jsx
```diff
@@ -0,0 +1,126 @@
...
+        if (this.props.results.length) {
+            action = (
+                <Button variant="info" onClick={this.props.handleCloseRunClick}>
+                    Close
+                </Button>
+            );
+            message = 'This test is complete.';
+            variant = 'info';
+            visible = true;
+        }
```
spectranaut, line 69: I'm a bit confused by this sequence of steps. So if there is a conflict, but the test is complete, the conflict is not shown in the status bar?

Also, if there is a conflict, I think the option should not be to "redo test" but "raise an issue".

But then who will re-do the test with the correct information once the issue is resolve? Maybe we need to leave the issue in the status bar until the tester marks the issue as "resolved for this test."

client/components/DisplayTest/index.jsx
rwaldron: Yes

client/actions/cycles.js
rwaldron: We might be thinking of something different here. The conflicts/diffs that @s3ththompson and I planned out are based on a serialization of the actual form itself within the iframe. That would also be used to repopulate those forms the next time you open the test to run it. The current data that I saw in the result objects does not have the details necessary to do that.

s3ththompson: i think the issue is that there are 2 concepts here with similar terminology. here's my conception of them, hopefully it matches yours?
- [1] the [results conflict diff](https://github.com/w3c/aria-at/issues/132#issuecomment-617225005), which Matt sketched out, which will go at the bottom of the GitHub issue, which Val points out can be generated on the frontend.
- [2] the (basically) vdom state that will need to be serialized from the iframe form, saved to the db, and then retrieved to re-hydrate the form during editing.

@rwaldron it sounds like this stub is for [2]? if so, is it appropriately named? don't we need to save the serialized data for all results, whether it's conflicted or not?

rwaldron: @s3ththompson yes, you're correct. @spectranaut and I had a call to discuss this and I described our idea and she's got it now.

server/migrations/20200521003201-create_table_test_issues.js
```diff
@@ -0,0 +1,36 @@
...
+                allowNull: false,
+                references: {
+                    model: 'run',
+                    key: 'id'
+                }
+            },
+            issue_number: {
+                type: Sequelize.INTEGER,
+                allowNull: false
+            }
```
s3ththompson, line 30: I think you may be missing the `open` boolean column?

server/models/TestIssue.js
```diff
@@ -0,0 +1,41 @@
...
+                    key: 'id'
+                }
+            },
+            issue_number: {
+                type: INTEGER,
+                allowNull: false
+            },
+            open: {
+                type: BOOLEAN,
+                allowNull: true
```
s3ththompson, line 33: I believe boolean false is *not* null, so `allowNull` can be false for consistency / safety.

server/migrations/20200521003201-create_table_test_issues.js
spectranaut: I just got an error about this @s3ththompson , I'll push a commit...

PR #97, opened 2020-05-19
Remove Flyway and move migrations to Sequelize

rmeritz:
- Remove the flyway migration and replace with sequelize migrations and sequelize seeds
- I tried several methodologies to create the initial migration. I tried using a command-line tool but it was no longer maintained. I tried writing it by hand with the generated models as my starting point, but there were very subtle differences in the index creation that caused errors. I ended up going with using a pgdump of the original state of the database and running the commands from that because it creates a db in the exact same state as before. This means the first migrations is non-traditional but it still allows us to use normal sequelize migrations moving forward which was the main point of this work.
- Setup new migrations to run on both Vagrant and locally
- Change all `json` datatypes to a `jsonb`
- Update the local Vagrant configuration to use a separate github oauth app so the sign up flow works correctly when testing with Vagrant
- Update documentation of how to work with the db locally
- Install the `dotfile-cli` as a way to easily load the environment variables
- Add configuration for the sequelize-cli so it can be used with `yarn run sequelize -- <ARGS>`
- Update documentation for information on how to deploy and how to test deploys with Vagrant
- I've tested the app locally and in Vagrant and deleted the db on staging and deployed there


docs/database.md
```diff
@@ -42,12 +36,36 @@ yarn db-import-tests:dev
...
 ```
-export PGDATABASE=aria_at_report
-export PGUSER=atr
-export PGPASSWORD=atr
-export PGHOST=localhost
-export PGPORT=5432
-export ROOTUSER=$USER
-```
\ No newline at end of file
+yarn run dotenv -e config/dev.env psql
```
evmiguel, line 41: This is awesome!

server/migrations/20200518185559-initial-migration.js
```diff
@@ -0,0 +1,21 @@
+const Promise = require('bluebird');
+const fs = require('fs');
+
+module.exports = {
```
evmiguel, line 4: I like this, and I think it's ok to create our initial state from the flyway dump.

server/seeders/20200518222159-role.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
```
evmiguel, line 3: Nice!

server/seeders/20200518223540-test_status.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
```
evmiguel, line 3: Nice!

server/seeders/20200518223022-browser.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
```
evmiguel, line 3: Loving these seeders!

rmeritz:
>  I added some more to the documentation for changing a flyway managed db to sequelize management. 

Your docs look really good. Thanks!

> Nonblocking question: what is server/migrations/.gitignore for?

It can be removed. It was added so that the folder would be tracked by git even though there was nothing it in. But then I put stuff in the folder and it became irrelevant. It had made since at some point though.

PR #105, opened 2020-05-22
Adding conflict handling

spectranaut:
Ok, to test this is a bit obnoxious

1. Make another user in the database:
`insert into users(username) values ("foobar");`
`insert into user_to_at(user_id, at_name_id) values (FOOBAR_ID, AT_NAME_ID);`

2. Add yourself and the new user to a test run of a test cycle you initiated.

3. Save some test results for that test run AS YOURSELF

4. Update those test results so that they belong to 'foodbar' in the db

`update test_results set (user_id=FOOBAR_USER_ID);`

5. Now go save tests as yourself

6. To see the conflicts, you will have to finish a test then go back to it. I should make it so that they show up when you click "next test" but I ran out of steam for now!


rwaldron:
When following your testing instructions, I get blanks where there should be some user's name. The reason for this is because your SQL only has a username and there are still places where `fullname` is used. To get around that, I added a user like this: 

```sql
WITH rows AS (
  INSERT INTO users (username, fullname) VALUES ('alice', 'Alice') RETURNING id
)
INSERT INTO user_to_at (user_id, at_name_id)
SELECT id, 1 FROM rows;
```

rwaldron:
> `update test_results set (user_id=FOOBAR_USER_ID);`

Should be: 
```
update test_result set user_id=FOOBAR_USER_ID;
```

client/components/DisplayTest/index.jsx
```diff
@@ -20,7 +20,17 @@ class DisplayTest extends Component {
     constructor(props) {
         super(props);
 
+        let {
+            test,
+            userId
+        } = this.props;
+
+        let conflicts = test.results
```
spectranaut, line 25: We need to calculate this for each new test, so I'm pretty sure you can't put this here

rwaldron: DisplayTest should be re-rendering every time it displays a new test.

PR #109, opened 2020-05-22
restores border with minor change to nav bar

jesdaigle:


rwaldron:
Thanks for sorting that out!!

PR #111, opened 2020-05-22
Edit test results in iframe, closes #103

### LINKED ISSUE #103: Support editing test results and saving of partial form values for aria-at tests
We need to support editing saved test results and saving of partial form values for embedded [w3c/aria-at](w3c/aria-at) tests. Since tests are loaded as iFrames, it is currently difficult to intercept the click handler on the "Review Results" button, *inside* the iFrame in order to accomplish the above task.

We need to update the [w3c/aria-at](w3c/aria-at) test harness to communicate with the parent window.

Then we need to prevent the test harness from showing results *inside* the iFrame in order to capture form values and show results *outside* the iFrame.

s3ththompson:
## Prerequisites

- Run `yarn db-migrate:dev`
- Import latest `aria-at` tests
- Run `yarn db-import-tests:dev -c 777bceeb7fded9dbafcb4a739e86e7b986372b1b `
- From your `aria-at-tests` directory: `git clone git@github.com:w3c/aria-at.git 777bceeb7fded9dbafcb4a739e86e7b986372b1b `
- Run `cd 777bceeb7fded9dbafcb4a739e86e7b986372b1b `
- Run `git reset --hard 777bceeb7fded9dbafcb4a739e86e7b986372b1b `

- Run `yarn`
- Run `yarn dev`

- Create a new test cycle with the commit referenced above. (Important!)

## Testing
This PR adds support for editing test results after submitting them. Navigate to the test queue, pick the new test cycle, and advance to a test that does not already have saved results.

Fill out a test result, noting your specific answers / test state. Submit results. And "Edit" button should appear. Click edit to be taken back to the form with your previous responses pre-populated.

- @rwaldron: you might specifically want to look at serialization code in `/client/lib/form-serialization.js`
- @spectranaut you might want to look at the new button handling logic in `/client/components/DisplayTest/` which I had to refactor given that saving results is now separate from nav buttons.

## Other Notes
- I returned the result of the `dispatch` from all of the redux actions in order to be able to await asynchronous redux actions (I did this to all actions for consistency even though I only needed it for `saveResult`)


client/lib/form-serialization.js
```diff
@@ -0,0 +1,84 @@
...
+            node.selected = nodeState.selected;
+        }
+        if (typeof nodeState.indeterminate !== 'undefined') {
+            node.indeterminate = nodeState.indeterminate;
+        }
+        if (typeof nodeState.checked !== 'undefined') {
+            node.checked = nodeState.checked;
+        }
+    }
+}
```
rwaldron, line 84: ❤️

server/migrations/20200521003201-create_table_test_issues.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
             }
         });
     },
-    down: (queryInterface) => {
+    down: queryInterface => {
```
rwaldron, line 33: Lol, I've been discarding this change all day, every time my linter fixes it, because I wasn't working on anything that would make sense to include this change 🤣

client/components/DisplayTest/index.jsx
```diff
@@ -42,27 +46,10 @@ class DisplayTest extends Component {
         this.handleModalCloseClick = this.handleModalCloseClick.bind(this);
         this.handleModalConfirmClick = this.handleModalConfirmClick.bind(this);
 
-        this.testIframe = React.createRef();
-    }
+        // iframe action
+        this.handleSubmit = this.handleSubmit.bind(this);
 
-    handleRaiseIssueClick() {
```
s3ththompson, line 48: @spectranaut I reordered a few things while rebasing in order to keep myself sane. this looks like a worse refactor than it is, but these are still here just under the other click handlers.

s3ththompson:
Looks like there is a non-Chrome issue to take a look at! I will fix in Firefox and test in other browsers in a followup PR.

PR #116, opened 2020-05-26
Add GitHub proxy to remove aria-at tests workaround

s3ththompson:
This PR changes the `/aria-at/*` route on the server and uses middleware from [rgrove/rawgit](https://github.com/rgrove/rawgit) to proxy requests to `githubusercontent` to serve test assets directly. All test assets are cached in memory for 7 days.

([rawgit.com](https://rawgit.com/), the hosted version, has been sunsetted, hence the need to pin the version we're importing to a specific commit immediately prior to the sunset.)

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -1,23 +1,5 @@
-const path = require('path');
-const cacheMiddleware = require('apicache').middleware;
-const proxyMiddleware = require('rawgit/lib/middleware');
-
-const BaseURL = 'https://raw.githubusercontent.com';
-
 const { listener } = require('./app');
 const port = process.env.PORT || 5000;
-
-const onlyStatus200 = (req, res) => res.statusCode === 200;
-
-listener.route('/aria-at/:branch/*').get(
-    cacheMiddleware('7 days', onlyStatus200),
-    (req, res, next) => {
-        req.url = path.join('w3c', req.url);
-        next();
-    },
-    proxyMiddleware.fileRedirect(BaseURL),
-    proxyMiddleware.proxyPath(BaseURL)
-);
 
 listener.listen(port, () => {
     // eslint-disable-next-line no-console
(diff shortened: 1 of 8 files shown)


server/server.js
```diff
@@ -1,13 +1,22 @@
-const express = require('express');
 const path = require('path');
+const cacheMiddleware = require('apicache').middleware;
```
evmiguel, line 2: While this works, the paradigm in this app is to put the middleware in the `app.js` file here: https://github.com/bocoup/aria-at-app/blob/main/server/app.js. 

My decision in this file to put `/aria-at` path in the `server.js` file was for the `express.static` functionality. For this,  though, I recommend doing as I said above.

s3ththompson: @evmiguel sure thing. could you clarify, if i move the middleware / path to the `app.js` file, the only thing in `server.js` file will be the `listen` right?

s3ththompson:
@evmiguel cc @spectranaut what is the best convention for loading interstitials for a11y? do you think the loading text be inside our outside the iFrame?

spectranaut:
@s3ththompson I'm not sure the best way to handle loading, I think we would need to research it. I think we should go with a follow up PR, but we should include this in the training -- that you might have to wait for the test to load.

spectranaut:
oh! I just looked at the loading video, and I realize that focus is automatically put on the test header when the test loads. This might be the a11y solution, because the focus will tell the user about the change, but we will find out during the pilot tests I suppose.

server/server.js
evmiguel: @s3ththompson yes!

s3ththompson:
@evmiguel fixed those changes and tested. let me know if there's anything else I should change!

evmiguel:
@s3ththompson looks good! merging.

PR #117, opened 2020-05-26
Rename test buttons for better UX

s3ththompson:
This PR makes the following changes:
- rename "Review Results" to "Submit Results"
- rename "Next Test" to "Skip Test" if results have not been submitted.

cc @isaacdurazo as a heads up for the current instructions

Note: The "correct" fix for the "Review Results" name change is to simply change the button name on the test harness side. To avoid that change today, the `TestIframe` patches the name after load. This code is not DRY, but will be removed in a follow-up PR after the test harness has been updated for the proper fix.


s3ththompson:
@spectranaut I also snuck in some HTML validation issues: https://validator.nu/?doc=http%3A%2F%2Faria-at.w3.org%2F

cc @zcorpan who suggested these changes be made before the pilot test

PR #118, opened 2020-05-26
Fix delete cycle foreign key constraint

spectranaut:
We should be able to delete cycles. Note, delete cycles deletes all test results. Before this, if there were test results or issues filed on tests, this delete did not work.

Do we want this change in before the pilot test? Seems potentially dangerous. Myabe we should not merge this and do it the right way later, with a popup that says "Are you sure you want to delete?"


spectranaut:
Nevermind, we should just fix this the right way when the time comes. Right now you can only delete cycles when there is no results or issues and I think tha tis fine.

PR #120, opened 2020-05-26
Fix assigning bugs

spectranaut:
Fixes:
https://app.asana.com/0/1140793737675474/1177519829083649
and 
https://app.asana.com/0/1140793737675474/1177519829083652

### DIFF
--- a/client/components/TestQueueRun/index.jsx
+++ b/client/components/TestQueueRun/index.jsx
@@ -66,10 +66,14 @@ class TestQueueRow extends Component {
             // You can only unassign if no tests results have been saved
             if (
                 testers.includes(tester.id) &&
-                this.testsCompletedByUser[tester.id]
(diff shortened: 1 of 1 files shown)


client/components/TestQueueRun/index.jsx
```diff
@@ -66,10 +66,14 @@ class TestQueueRow extends Component {
...
             if (
                 testers.includes(tester.id) &&
-                this.testsCompletedByUser[tester.id]
+                ! this.testsCompletedByUser[tester.id]
             ) {
                 canUnassignTesters.push(tester);
-            } else if (
+            }
+
+            // You can only unassign if the tester is not assigned alread
```
evmiguel, line 74: I think this comment should be:  **You can only _assign_ if the tester is not assigned already**

PR #125, opened 2020-05-28
Add run status

spectranaut:
To test, run:
```
yarn sequelize db:migrate 
yarn sequelize db:seed:all 
```

The purpose of this change is to support the ability to mark runs as "draft" and "final', see the "status" and "action" columns as described in the test queue: https://user-images.githubusercontent.com/1379244/79916778-23833080-83de-11ea-97ea-96edb644d4d3.png

### DIFF
--- a/server/migrations/20200528143624-add-run-status-to-run.js
+++ b/server/migrations/20200528143624-add-run-status-to-run.js
@@ -1,0 +1,18 @@
+'use strict';
+
+module.exports = {
+    up: (queryInterface, Sequelize) => {
+        return queryInterface.addColumn('run', 'run_status_id', {
+            type: Sequelize.INTEGER,
+            allowNull: true,
+            references: {
+                model: 'run_status',
+                key: 'id'
+            }
+        });
+    },
+
+    down: queryInterface => {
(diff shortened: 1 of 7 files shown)


server/migrations/20200528143624-add-run-status-to-run.js
```diff
@@ -0,0 +1,18 @@
...
+            allowNull: true,
+            references: {
+                model: 'run_status',
+                key: 'id'
+            }
+        });
+    },
+
+    down: queryInterface => {
+        return queryInterface.removeColumn('test_result', 'run_status');
```
evmiguel, line 16: This looks like a typo. Should it be:

```
queryInterface.removeColumn('run', 'run_status_id');
```

spectranaut:
I also forgot to update the models -- will do in this PR

server/migrations/20200528172804-add-run-status-to-run-data-view.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+
```
evmiguel, line 2: Ran this migration, and it works!

PR #132, opened 2020-05-29
Only admins can perform status changes

spectranaut:
This is an important bug fix on yesterdays work.

It also changes to pieces of logic:
1. You can mark results as "draft" or "final" as soon as one test is compete. I imagine tests that have bugs will want to be omitted from published test results, so we should allow publishing of incomplete sets of results.
2. I was incorrectly counting the number of tests that were complete (specifically, skipped tests were counted as completed tests)

### DIFF
--- a/client/components/TestQueueRun/index.jsx
+++ b/client/components/TestQueueRun/index.jsx
@@ -14,6 +14,23 @@ import {
 class TestQueueRow extends Component {
     constructor(props) {
         super(props);
+
+        const { totalConflicts, testsWithResults } = this.countCompleteTestsAndConflicts();
+        this.state = {
+            totalConflicts,
+            testsWithResults
+        };
+
+        this.handleUnassignSelfClick = this.handleUnassignSelfClick.bind(this);
+        this.handleAssignSelfClick = this.handleAssignSelfClick.bind(this);
+        this.assignTesterSelect = this.assignTesterSelect.bind(this);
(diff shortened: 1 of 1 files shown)


client/components/TestQueueRun/index.jsx
```diff
@@ -197,6 +198,31 @@ class TestQueueRow extends Component {
...
+        ) {
+            updateRunStatusButton = (
+                <Button onClick={() => this.updateRunStatus('draft')}>
+                    Mark as draft
+                </Button>
+            );
+        }
+        // If the results have been marked as draft and there is no conflict,
+        // they can be marked as "final"
+        else if (runStatus === 'draft' && this.state.totalConflicts === 0) {
```
evmiguel, line 218: I like the state conditional here!

client/components/TestQueueRun/index.jsx
```diff
@@ -347,28 +374,17 @@ class TestQueueRow extends Component {
...
-                updateRunStatusButton = (
-                    <Button onClick={() => this.updateRunStatus('final')}>
-                        Mark as final
-                    </Button>
-                );
-            } else if (runStatus === 'final') {
-                // To do: make this a link to published results
-                status = 'PUBLISHED RESULTS';
-            }
+        if (runStatus === 'draft') {
```
evmiguel, line 382: Verified that this works!

PR #136, opened 2020-06-03
Fixes to initiate test run

spectranaut:
We got a lot of feedback on the initiate run page, and I tried to implement the minimal amount of it. 

The logic on this page it complicated and I could appreciate some thorough testing of both the initiate test cycle page and the cycle summary page -- both you can get to from "Test Management".

You can now select only some design patterns, and you can delete rows.

### DIFF
--- a/client/components/ConfigureRunsForExample/index.jsx
+++ b/client/components/ConfigureRunsForExample/index.jsx
@@ -1,7 +1,6 @@
 import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { Form, Table } from 'react-bootstrap';
-import nextId from 'react-id-generator';
 
 class ConfigureRunsForExample extends Component {
     constructor(props) {
@@ -20,7 +19,6 @@ class ConfigureRunsForExample extends Component {
         const {
             assignTesters,
             removeAllTestersFromRun,
-            testersByRunId,
             example,
             usersById,
             runs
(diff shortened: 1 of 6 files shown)


client/components/ConfigureRunsForExample/index.jsx
```diff
@@ -95,12 +90,13 @@ class ConfigureRunsForExample extends Component {
             return b.at_id - a.at_id;
         });
 
-        let exampleTableTitle = example.name || example.directory;
-        let tableId = nextId('table_name_');
+        // If no runs are selected, disabled the assign drop down
+        let disableAssignDropdown = !Object.values(this.state.runSelected).find(
+            v => v
+        );
```
rwaldron, line 96: Would `!Object.entries(this.state.runSelected).length` be sufficient here?

client/components/ConfigureRunsForExample/index.jsx
```diff
@@ -115,15 +111,12 @@ class ConfigureRunsForExample extends Component {
...
-                                            let user = usersById[testerId];
-                                            return user.username;
-                                        }
-                                    );
-                                }
+                                let names = run.testers
+                                    ? run.testers.map(t => {
+                                          let user = usersById[t];
+                                          return user.username;
+                                      })
```
rwaldron, line 118:
```js
run.testers.map(t => usersById[t].username)
```

client/components/InitiateCycle/index.jsx
```diff
@@ -326,28 +478,72 @@ class InitiateCycle extends Component {
...
-                                usersById={usersById}
-                                assignTesters={this.assignTesters}
-                                removeAllTestersFromRun={
-                                    this.removeAllTestersFromRun
-                                }
-                                testersByRunId={
-                                    this.state.runTestersByExample[example.id]
-                                }
-                            />
+                            <>
```
rwaldron: You lost the key here, needs to be something like `<Fragment key={example.id}>`

PR #137, opened 2020-06-04
Data for test results page

spectranaut:


client/components/ResultsPage/index.jsx
```diff
@@ -0,0 +1,279 @@
...
+
+        // eslint-disable-next-line no-console
+        console.log('cyclesById', cyclesById);
+        // eslint-disable-next-line no-console
+        console.log('testsByRunId', testsByRunId);
+
+
+        for (let cycle of Object.values(cyclesById)) {
+            // eslint-disable-next-line no-console
+            console.log('Cycle: ', cycle.name);
```
spectranaut, line 87: This is the cycle name, @jesdaigle !

client/.babelrc
```diff
@@ -1,8 +1,11 @@
 {
     "presets": ["@babel/env", "@babel/preset-react"],
+          "plugins":
+          ["@babel/plugin-proposal-class-properties"],
```
s3ththompson, line 4: Since we are no longer using class properties in the ResultsPage component, let's not add this plugin.

PR #143, opened 2020-06-05
Add app screenshots to home page

isaacdurazo:


package.json
```diff
@@ -37,6 +37,7 @@
     "server"
   ],
   "devDependencies": {
+    "file-loader": "^6.0.0",
```
s3ththompson, line 40: I believe that this should actually be in `client/package.json` and if you add it via `yarn add file-loader` from the `client` folder, I don't think it will generate a `package-lock.json`.

s3ththompson:
see inline comment above, but I believe we should not in fact have a `package-lock.json` checked in as it's a sign of mixing `npm` with `yarn`. I think we need to `yarn add file-loader` from the `client` folder to fix.

s3ththompson:
@isaacdurazo if you don't mind me pushing a change to this PR, I can fix this when I update test content.

isaacdurazo:
Thanks you @s3ththompson! Feel free to push the fix :)

s3ththompson:
@zcorpan can you review the text content of the descriptions for these 4 screenshots?

client/components/Home/index.jsx
```diff
@@ -1,65 +1,147 @@
...
+                                <a href="https://www.w3.org/community/about/faq/#how-do-i-join-a-group">
+                                    how to join a W3C community group
+                                </a>
+                                )
+                            </li>
+                            <li>
+                                <a href="https://github.com/w3c/aria-at/issues/162">
+                                    Sign up as a tester
+                                </a>
+                                {' for the Pilot Test (May 27 - June 2)'}
```
zcorpan, line 49: Pilot test is over. We don't have anything yet to coordinate testers for the full cycle. Suggest remove for now

client/components/Home/index.jsx
```diff
@@ -1,65 +1,147 @@
...
+                                </a>
+                            </li>
+                        </ul>
+                    </section>
+                    <section className="app-screenshots">
+                        <div className="container">
+                            <div className="description">
+                                <h2>Test Management</h2>
+                                <p>
+                                    This app allows users to create Test Cycles
```
zcorpan, line 78: Maybe we should write to the reader with the assumption that they are the user. So, say "You can create Test Cycles" or similar

zcorpan:
I think the alt text should build on the prose and provide a bit more detail or something that is not present in the text next to the image. More like the wireframe descriptions instead of "Screenshot of foo".

For AT reading order, the heading and text should come before the image in the DOM order. Use CSS to get the layout right (e.g. with floats).

Test with mobile size viewport and make it single-column there

server/tests/integration/server.int.test.js
```diff
@@ -150,7 +150,9 @@ describe(atEndpoint, () => {
 
 describe(`${testEndpoint}/import`, () => {
     it(`POST ${testEndpoint}/import`, async () => {
-        const response = await request(listener).post(`${testEndpoint}/import`).send({git_hash: 1234});
+        const response = await request(listener)
+            .post(`${testEndpoint}/import`)
+            .send({ git_hash: 1234 });
         expect(response.statusCode).toBe(200);
```
rwaldron, line 156: These changes were made in https://github.com/bocoup/aria-at-app/pull/146/files#diff-15c43bcc258fd5c8ff9f71814f5658a1

isaacdurazo: @s3ththompson could you take care of this part?

s3ththompson: done!

s3ththompson:
@isaacdurazo rebased on top of master (which also addressed @rwaldron's comment). should be all set now

PR #146, opened 2020-06-08
Add a cron job to import tests from aria-at every 15 mins

spectranaut:
We could do it more or less often, this will just go and import HEAD of master.

I deploy to staging and the deploy was successful, but I'm otherwise not sure how to test this, as there has been no recent commits to master on staging...

### DIFF
--- a/deploy/roles/application/tasks/cron.yml
+++ b/deploy/roles/application/tasks/cron.yml
@@ -1,0 +1,7 @@
+---
+# Create a cron to import the most recent tests from aria-at
+- name: Set a cron job to build test results
+  cron:
+    name: "import latest aria-at tests"
+    minute: "15"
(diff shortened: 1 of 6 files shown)


deploy/roles/application/tasks/cron.yml
```diff
@@ -0,0 +1,7 @@
+---
+# Create a cron to import the most recent tests from aria-at
+- name: Set a cron job to build test results
+  cron:
+    name: "import latest aria-at tests"
+    minute: "15"
+    job: "curl -X POST https://{{fqdn}}/api/import/test"
```
spectranaut, line 7: Should we be saving the response somewhere, for debugging purposes?

rwaldron:
I test this by purging all test data from my local database (what a dance that was!): 

## Before import 
image
image

## After import
image
image

PR #160, opened 2020-09-15
Remove access for unlogged in user

spectranaut:
[asana issue](https://app.asana.com/0/1193055321453706/1172041722322605)

This makes the following changes:

The following pages are not only reachable if logged in by either tester or admin (otherwise page not found)
- `/account/settings`
- `/test-queue/:cycleId(\\d+)`
- `/test-queue`

The following pages are left exposed because as I recall, Matt might want to share them by link:
- `/results`
- `/cycles/:cycleId(\\d+)/run/:runId(\\d+)`

 On the execute test page, you cannot save results or submit an issue, the buttons silently do not work -- but it we want to leave a public facing version of the page, we should do something about that. We might want to make a part of Isaac's redesign of the tester workflow and we will need to find out the priority of this from Matt.

### DIFF
--- a/client/routes/index.js
+++ b/client/routes/index.js
@@ -33,7 +33,13 @@ export default [
             {
                 path: '/account/settings',
                 exact: true,
-                component: UserSettings
+                component: () => {
+                    return (
+                        <ConfirmAuth requiredPermission="tester">
+                            <Route component={UserSettings} />
+                        </ConfirmAuth>
+                    );
+                }
             },
             {
                 path: '/cycles',
@@ -48,12 +54,24 @@ export default [
             },
             {
                 path: '/test-queue/:cycleId(\\d+)',
-                component: TestQueue
+                component: () => {
+                    return (
+                        <ConfirmAuth requiredPermission="tester">
+                            <Route component={TestQueue} />
+                        </ConfirmAuth>
+                    );
+                }
             },
             {
                 path: '/test-queue',
                 exact: true,
-                component: TesterHome
+                component: () => {
+                    return (
+                        <ConfirmAuth requiredPermission="tester">
+                            <Route component={TesterHome} />
+                        </ConfirmAuth>
+                    );
+                }
             },
             {
                 path: '/results',


client/routes/index.js
```diff
@@ -33,7 +33,13 @@ export default [
             {
                 path: '/account/settings',
                 exact: true,
-                component: UserSettings
+                component: () => {
+                    return (
+                        <ConfirmAuth requiredPermission="tester">
```
spectranaut, line 38: the access of an admin is everything the tester has access to + specially admin pages. Should we make that clearer on this page? And do something like `requiredPermission="any"` or something

rmeritz: I think this seems fine and how one would normally reason about things.

rmeritz:
> On the execute test page, you cannot save results or submit an issue, the buttons silently do not work -- but it we want to leave a public facing version of the page, we should do something about that. We might want to make a part of Isaac's redesign of the tester workflow and we will need to find out the priority of this from Matt.

I think hiding the buttons for users who shouldn't be able to access them should be considered part of this bug fix...

spectranaut:
> > On the execute test page, you cannot save results or submit an issue, the buttons silently do not work -- but it we want to leave a public facing version of the page, we should do something about that. We might want to make a part of Isaac's redesign of the tester workflow and we will need to find out the priority of this from Matt.
> 
> I think hiding the buttons for users who shouldn't be able to access them should be considered part of this

spectranaut:
> > On the execute test page, you cannot save results or submit an issue, the buttons silently do not work -- but it we want to leave a public facing version of the page, we should do something about that. We might want to make a part of Isaac's redesign of the tester workflow and we will need to find out the priority of this from Matt.
> 
> I think hiding the buttons for users who shouldn't be able to access them should be considered part of this bug fix...

On second thought, especially after today's meeting, I'm not sure whether this page needs to be public -- today we talked about having the test plans reviewable and linked to the actual test, which would essentially invalidate the need for this page to be public. We need to check with Matt somehow.

PR #162, opened 2020-09-17
Remove global mocks

spectranaut:


rmeritz:
I'm so stoked you finished getting this working. I didn't do a really deep review. But I'm so happy. And would be happy if you merged it.

PR #163, opened 2020-09-17
Prepare for removal of TestCycle

rmeritz:
- Added the new models/column change as planned out in this design sketch: https://docs.google.com/document/d/1RhRBDl4Go-pM7DWUQBf14ogXRslV2CVZB5KQ8lCEzw0/edit

- Added the controller, routes and for new BE API endpoints and fleshed out a description of the API

- Added shells of services to do the actual work. I'm currently working on a separate PR that actually fleshes out the adding of the services. This is the majority of the remaining BE work.

### DIFF
--- a/server/controllers/RunController.js
+++ b/server/controllers/RunController.js
@@ -1,0 +1,41 @@
+const RunService = require('../services/RunService');
+
+async function configureRuns(req, res) {
+    const cycle = req.body.data;
+    try {
+        const savedRun = await RunService.configureRuns(cycle);
+        res.status(201).json(savedRun);
+    } catch (error) {
+        res.status(400);
+        res.end();
+        console.error(`Error caught in RunController: ${error}`);
+    }
+}
+
+async function getActiveRuns(req, res) {
+    try {
+        const cycles = await RunService.getActiveRuns();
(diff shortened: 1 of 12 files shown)


server/controllers/RunController.js
```diff
@@ -0,0 +1,41 @@
+const RunService = require('../services/RunService');
+
+async function configureRuns(req, res) {
+    const cycle = req.body.data;
```
spectranaut, line 4: I'm guessing we won't want to use this word "cycle" -- but we are sending configuration data and then getting back all active runs right? So maybe it's just "config". This is a nit pick and maybe we just don't need to make this change now.

server/controllers/RunController.js
```diff
@@ -0,0 +1,41 @@
...
+    try {
+        const savedRun = await RunService.configureRuns(cycle);
```
spectranaut, line 6: perhaps this is "activeRuns"

spectranaut: or maybe what we will return is an object that explains the changes that will be made with the new configuration.

spectranaut (approved):
This looks great only minor comments that will certainly be fixed when that function is revisited

PR #169, opened 2020-09-29
Add configure runs

spectranaut:
To test, you will need to migrate and import tests.
```
yarn sequelize:test db:migrate
yarn db-import-tests:test -c '826d2c2a6a9c5b9c7045d6fa7e1b575ce4fb9762'
yarn db-import-tests:test -c '413183a5f6909940f163a433931bd17ab6b48886'
```

Things to do:
- [x] add importing tests to travis, right now in travis test will fail because there is no test import
- [x] @rmeritz fix getActiveRuns test that were broken by my change
- [x] @rmeritz add migration to add "active" to test version table and make sure that ApgExample and Runs associate with the old test don't need to be updated
- [x] @rmeritz add the active/inactivating test version based on test version in configuration
- [x] @spectranaut Persisting ApgExample active/inactive changes
- [x] @spectranaut update tests to test whether or not the ApgExample, TestVersion, BrowserVersionToAtVersion rows have been active/de-active


server/services/RunService.js
```diff
@@ -38,13 +38,281 @@ const db = require('../models/index');
...
 
+
 /* eslint-disable no-unused-vars */
 async function configureRuns({
     test_version_id,
     apg_example_ids,
     at_browser_pairs
 }) {
+    // if (!t) {
+    //     t = await db.sequelize.transaction();
```
rmeritz, line 49: I don't know what is going on here. 

But I also think probably the whole thing should be wrapped in a transaction. Because we don't want to update half the things and fail. Or have a problematic race condition.

server/services/RunService.js
```diff
@@ -1,8 +1,8 @@
 const db = require('../models/index');
+const { Op } = require("sequelize");
```
rmeritz, line 2: I think you should be able to access this from the `db`

server/services/RunService.js
```diff
@@ -38,13 +38,281 @@ const db = require('../models/index');
...
+                    at_id,                                    // maybe eventually will remove column
+                    browser_version_to_at_versions_id: r.browser_version_to_at_version_id,
+                    apg_example_id: r.apg_example_id,
+                    test_version_id,
+                    active: true,
+                    run_status_id: runStatus.id
+                };
+            });
+
+            await db.Run.bulkCreate(dbRuns);
```
rmeritz, line 313: This is the only time you do anything in bulk. But you have a lot of loops above where I think you could/should be doing bulk actions.

server/tests/unit/services/RunService.test.js
```diff
@@ -10,37 +10,194 @@ afterAll(async done => {
...
+                    apg_example_name: apgExample.name,
+                    apg_example_id: apgExample.id,
+                    run_status_id: runStatus.id,
+                    run_status: runStatus.name,
+                    test_version_id: testVersion.id,
+                    testers: []
                 });
+            });
+        });
+        it('Activate deactived run', async () => {
```
rmeritz, line 102: What is this test testing... Can you name it more clearly? 

What is the other test you want to add. 

I'm confused what you are including in your naive version of this implementation.

server/services/RunService.js
```diff
@@ -44,7 +43,269 @@ async function configureRuns({
...
+        //         mark all the runs from the old test version as inactive
+        //         mark all the old apg_examples as inactive
+
+        // Activate/deactive APGExample rows
+
+        const apgExamples = await db.ApgExample.findAll({
+            where: {
+                test_version_id
+            }
+        });
```
spectranaut, line 61: Look at this findAll, @rmeritz  -- if the config causes a change to the test version, we DO need to deactivate all the APGExamples associated with that old test version, because we aren't going to deactivate them in this logic. Same with the runs.

spectranaut: we could change this to where active=true or where test_version_id

PR #170, opened 2020-09-29
Fix null constraint bug on created_at and updated_at columns in latest migration

evmiguel:
I was able to reproduce this bug by populating my database with the staging database data:
```
# On staging server
ssh root@aria-at-staging.w3.org
pg_dump -U atr -h localhost aria_at_report > dump.sql

# On local machine
scp root@aria-at-staging.w3.org:dump.sql .
psql -d aria_at_report < dump.sql
```

## Changes in this PR:

- Changed `createdAt` and `updatedAt` to `created_at` and `updated_at` in `run` table
- Set a default value in the migration, which causes Sequelize to auto-populate the rows with null in those columns
- Updated Run model to include the `createdAt` and `updatedAt` attributes.

## Updating database
```
# Rollback to the migration before the latest one
yarn sequelize db:migrate:undo

# Run the migrations
yarn sequelize db:migrate

## This is outdated. There will be nulls.
# Verify that the rows were auto-populated with -infinity
psql -d aria_at_report -c "select id, created_at, updated_at from run;"
```

### DIFF
--- a/server/migrations/20200915203422-AddColumnsToTestRun.js
+++ b/server/migrations/20200915203422-AddColumnsToTestRun.js
@@ -26,19 +26,21 @@ module.exports = {
                 ),
                 queryInterface.addColumn(
                     'run',
-                    'createdAt',
+                    'created_at',
                     {
                         type: Sequelize.DATE,
-                        allowNull: false
+                        allowNull: false,
+                        defaultValue: Number.NEGATIVE_INFINITY
                     },
                     { transaction: t }
                 ),
                 queryInterface.addColumn(
                     'run',
-                    'updatedAt',
+                    'updated_at',
                     {
                         type: Sequelize.DATE,
-                        allowNull: false
+                        allowNull: false,
+                        defaultValue: Number.NEGATIVE_INFINITY
                     },
                     { transaction: t }
                 )
@@ -58,10 +60,10 @@ module.exports = {
                 queryInterface.removeColumn('run', 'test_version_id', {
                     transaction: t
                 }),
-                queryInterface.removeColumn('run', 'createdAt', {
+                queryInterface.removeColumn('run', 'created_at', {
                     transaction: t
                 }),
-                queryInterface.removeColumn('run', 'updatedAt', {
+                queryInterface.removeColumn('run', 'updated_at', {
                     transaction: t
                 })
             ]);
--- a/server/migrations/20200915203422-AddColumnsToTestRun.js
+++ b/server/migrations/20200915203422-AddColumnsToTestRun.js
@@ -29,8 +29,7 @@ module.exports = {
                     'created_at',
                     {
                         type: Sequelize.DATE,
-                        allowNull: false,
-                        defaultValue: Number.NEGATIVE_INFINITY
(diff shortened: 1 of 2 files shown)


server/migrations/20200915203422-AddColumnsToTestRun.js
```diff
@@ -26,19 +26,21 @@ module.exports = {
                 ),
                 queryInterface.addColumn(
                     'run',
-                    'createdAt',
+                    'created_at',
```
evmiguel, line 29: The queryInterface does not have the capability to use `underscored: true`

server/migrations/20200915203422-AddColumnsToTestRun.js
```diff
@@ -26,19 +26,21 @@ module.exports = {
...
                     {
                         type: Sequelize.DATE,
-                        allowNull: false
+                        allowNull: false,
+                        defaultValue: Number.NEGATIVE_INFINITY
```
spectranaut, line 33: unfortunately, new entries now have negative infinity for this value as well!

spectranaut (changes requested):

server/migrations/20200915203422-AddColumnsToTestRun.js
evmiguel: @spectranaut oh dear! Can you show me how you tested what happens with new entries?

spectranaut: yes def! I test by running the migration, then running "yarn dev", then adding a test cycle through the front end, then inspecting the `run` table in the database.

spectranaut: I just tested out of curiosity, changing "defaultValue" to "Sequelize.Now" didn't work --- you still got the null constraints on the migration. I think we have no choice but to allow null. However, I think we also need to add Sequelize.Now to get those values populated. This means, in the migration:
```
                queryInterface.addColumn(
                    'run',
                    'created_at',
                    {
                        type: Sequelize.DATE,
                        defaultValue: Sequelize.NOW
                    },
                    { transaction: t }
                ),
                queryInterface.addColumn(
                    'run',
                    'updated_at',
                    {
                        type: Sequelize.DATE,
                        defaultValue: Sequelize.NOW
                    },
                    { transaction: t }
                )
```
and in the model:
```
            createdAt: {
                type: DataTypes.DATE,
                allowNull: true,
                defaultValue: sequelize.NOW
            },
            updatedAt: {
                type: DataTypes.DATE,
                allowNull: true,
                defaultValue: sequelize.NOW
            }
```

spectranaut:
@evmiguel I just pushed the change I found worked can you run it to test it too? If you're good with it we can merge :)

server/migrations/20200915203422-AddColumnsToTestRun.js
evmiguel: @spectranaut You commit works! I suppose having nulls is just as informative as having something like a negative infinity take the place of a null.

spectranaut (approved):

PR #171, opened 2020-10-05
Complete Published Runs API Endpoint

rmeritz:
- The controller/routing already existed
- The published Runs return the same `Run` object as the active runs endpoint so abstract out a function to convert runs from the DB to the run objects returned to the FE.

### DIFF
--- a/server/services/RunService.js
+++ b/server/services/RunService.js
@@ -312,13 +312,48 @@ async function configureRuns({
     }
 }
 
+async function sequelizeRunsToJsonRuns(sequelizeRuns) {
+    let ats = {};
+    if (sequelizeRuns.length > 0) {
+        ats = await db.At.findAll({
+            where: {
+                test_version_id: sequelizeRuns[0].test_version_id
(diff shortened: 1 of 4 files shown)


server/services/RunService.js
```diff
@@ -337,46 +372,7 @@ async function getActiveRuns() {
...
-                apg_example_name: activeRun.ApgExample.name,
-                apg_example_id: activeRun.apg_example_id,
-                run_status_id: activeRun.run_status_id,
-                run_status: activeRun.RunStatus.name,
-                test_version_id: activeRun.test_version_id,
-                testers: activeRun.Users.map(u => u.id)
-            };
-            return acc;
-        }, {});
+        return await sequelizeRunsToJsonRuns(activeRuns);
```
evmiguel, line 375: Nice! Very clean.

evmiguel (approved):
Nice abstraction! I have thoughts about the tests, but I think they can be addressed in my work this week.

PR #173, opened 2020-10-05
Add new `GET /api/run/config` and `GET /api/test-versions` endpoints 

spectranaut:
- Added to new API endpoints following the jdoc specs laid out by @spectranaut 
- `GET /api/run/config` is used to populate the current settings on the edit active run configuration page
- `GET /api/test-versions` is used to populate the possible test versions on the edit active run configuration page
- Both endpoints use services in the `RunService`. Even though `getNewTestVersions` doesn't clearly do anything with runs I left it in the run service because it shares a lot of code with the the `getActiveRunsConfiguration` method that definately belongs in `RunService` because they both return `TestSuiteVersion` objects
- I added through unit testing of the `RunService`, but didn't test the controllers/routes because they are just boilerplate

### DIFF
--- a/server/services/RunService.js
+++ b/server/services/RunService.js
@@ -2,9 +2,10 @@ const db = require('../models/index');
 /**
  * @typedef AtBrowserPairing
  * @type {object}
- * @property at_version
- * @property browser_id
- * @property browser_version
+ * @property {number} at_name_id
+ * @property {string} at_version
+ * @property {number} browser_id
+ * @property {string} browser_version
  *
  * @typedef Run
  * @type {object}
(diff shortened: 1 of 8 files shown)


server/services/RunService.js
```diff
@@ -23,6 +24,43 @@ const db = require('../models/index');
...
+ * @typedef TestSuiteVersion
+ * @type {object}
+ * @property {number} id
+ * @property {string} git_repo
+ * @property {string} git_tag
+ * @property {string} git_hash
+ * @property {string} git_commit_msg
+ * @property {Object.Date} date
+ * @property {Array.<At>}          supported_ats
+ * @property {Array.<ApgExamples>} apg_examples
```
spectranaut, line 37: This object should be there same as this object returned by "getAllTestVersion", see documentation of that function:  https://github.com/w3c/aria-at-app/blob/28e80d62c43e612e064ee22bedec1bc1230e89d1/server/services/CycleService.js#L224

spectranaut: This whole file will be deleted, and the two functions I propose here replace that one. The first will get the current "test version" and run configuration information -- which is used various places in the app, and the second end point is just for the "Test Run Configuration" page.

evmiguel (approved):
Verified that this works for me at `localhost:5000/api/run/config` and `localhost:5000/api/test-versions`. Unit tests are also working.

I agree to keep the `getNewTestVersions()` in the RunService as well. Nice job!

PR #174, opened 2020-10-05
Configure runs unit tests

evmiguel:


server/tests/unit/services/RunService.test.js
```diff
@@ -236,6 +237,85 @@ describe('RunService', () => {
...
+                        }
+                    ]
+                });
+
+                // There are three apg examples
+                const keys = Object.keys(activeRuns);
+                expect(keys.length).toEqual(3);
+
+                // TODO: is there a better way to assert this?
+                Object.values(activeRuns).forEach((run, i) => {
```
rmeritz, line 286: I know it is super verbose, but its best not to be using higher-order functions like `forEach` in tests. 

The code is the correct place for this kind of abstraction. But not the tests itself. 

https://mtlynch.io/good-developers-bad-tests/#dare-to-violate-dry

Also if you want to ignore the Id instead of deleting the id you could use `objectContaining`: https://jestjs.io/docs/en/expect#expectobjectcontainingobject

server/tests/unit/services/RunService.test.js
```diff
@@ -236,6 +237,85 @@ describe('RunService', () => {
...
+                        apg_example_id: apgExample.id,
+                        run_status_id: runStatus.id,
+                        run_status: db.RunStatus.RAW,
+                        test_version_id: testVersion.id,
+                        testers: []
+                    }
+
+                    expect(run).toEqual(result);
+                });
+            });
```
rmeritz, line 308: You are not checking the active status of anything. Which I think you want to do.

server/tests/unit/services/RunService.test.js
```diff
@@ -236,6 +237,85 @@ describe('RunService', () => {
...
+        });
+        it('creates several runs from a list of tech pairs', async () => {
+            //TODO
+        });
+        it('creates several runs from a list of apg examples and a list of tech pairs', async () => {
+            //TODO
+        });
+        it('creates several runs from a list of apg examples when there are existing apg examples in the database', async () => {
+            //TODO
+        });
```
rmeritz, line 318: I don't think this is the correct list of tests that need to be written. (Not sure if is meant to be the full list or not). 

It doesn't add much to do a list of several things verses one of the things. In most cases that isn't a huge problem. I would probably only add tests like that if there was a specific bug. Or maybe one that doesn't cover all cases and just make sure that I have one test that covered all cahses. 

I think it would take me a while to sit down and think of all the cases that need to be covered. But the thing that doens't seem to be covered by these test case descriptions is the change of state... I don't think the creation of new state is probably that hard or a thing I'm really worried about catching bugs in but the thing I'm really nervous about and want throughly tested is when when some things already exist in the DB from the previously active run or the previously previouslu active run and we need to copy over some of thata state.

rmeritz (changes requested):
I left some feedback to help guide the future work.

server/tests/unit/services/RunService.test.js
```diff
@@ -224,18 +225,220 @@ describe('RunService', () => {
...
+                );
+                expect(activeCreatedBrowserVersion[0].version).toBe(browserVersionNumber2);
+
+                const activeCreatedAtVersion = await db.AtVersion.findAll(
+                    { where: { id: activeBrowserVersionToAtVersions[0].at_version_id }}
+                );
+                expect(activeCreatedAtVersion[0].version).toBe(atVersionNumber2);
+            });
+        });
+        it('sets an existing inactive run to active', async () => {
```
evmiguel, line 337: Reviewer: here is a test.

server/tests/unit/services/RunService.test.js
```diff
@@ -224,18 +225,220 @@ describe('RunService', () => {
...
-                // After this another test should be writtent hat is like
-                // excalty the same but the active tech also changes.
-                //
-                // Ideally another test maybe were you kept one and add a new
-                // one...
                 await tech.reload();
                 expect(tech.active).toBe(true);
             });
         });
+        it('update active tech pairs when there is existing tech pair data', async () => {
```
evmiguel, line 232: Reviewer: here is a test.

server/tests/unit/services/RunService.test.js
```diff
@@ -224,18 +224,323 @@ describe('RunService', () => {
...
+                    test_version_id: testVersion.id,
+                    testers: []
+                });
+
+                // Verify that pre-existing run is now active
+                await previouslyInactiveRun.reload();
+                expect(previouslyInactiveRun.active).toBe(true);
+            });
+        });
+        it('sets one run inactive, one run active, and creates a new run when the apg examples change', async () => {
```
evmiguel, line 441: Reviewer: Here is a test.

server/tests/unit/services/RunService.test.js
```diff
@@ -224,18 +224,421 @@ describe('RunService', () => {
...
+                );
+                expect(activeCreatedBrowserVersion[0].version).toBe(
+                    browserVersionNumber2
+                );
+
+                const activeCreatedAtVersion = await db.AtVersion.findAll({
+                    where: {
+                        id: activeBrowserVersionToAtVersions[0].at_version_id
+                    }
+                });
```
rmeritz, line 344: If you are looking for one object just use `findOne` or in this case `findByPk`.

And in this case, as you are looking for a related association you should be preloading it with `include` and then then referencing it directly like:

 ```
                const activeBrowserVersionToAtVersions = await db.BrowserVersionToAtVersion.findAll(
                    { where: { active: true } },
                    { include: [db.AtVersion]}
                );
activeBrowserVersionToAtVersions.AtVersion
```

File: server/tests/unit/services/RunService.test.js [hunk 8]
```diff
@@ -224,18 +224,421 @@ describe('RunService', () => {
...
+                const runStatus = await db.RunStatus.findOne({
+                    where: { name: db.RunStatus.RAW }
+                });
+
+                let techPairDeactivate = await db.BrowserVersionToAtVersion.create(
+                    {
+                        at_version_id: atVersion.id,
+                        browser_version_id: browserVersion.id,
+                        active: true,
+                        run_status_id: runStatus.id
```
rmeritz, line 272: This is not an attribute on the tech pair.

server/tests/unit/services/RunService.test.js
```diff
@@ -224,18 +224,421 @@ describe('RunService', () => {
...
+                    active: false,
+                    run_status_id: runStatus.id
+                });
+
+                let previouslyActiveRun = await db.Run.findOne({
+                    where: { id: previouslyActiveRunEntry.id }
+                });
+                let previouslyInactiveRun = await db.Run.findOne({
+                    where: { id: previouslyInactiveRunEntry.id }
+                });
```
rmeritz, line 610: I'm confused about why you need to `findOne` object you just created. You should have it already if you just created it.

server/tests/unit/services/RunService.test.js
```diff
@@ -224,18 +224,421 @@ describe('RunService', () => {
...
+
+                await previouslyInactiveRun.reload();
+                expect(previouslyInactiveRun.active).toBe(true);
+
+                const newlyCreatedRun = Object.values(activeRuns).filter(
+                    run =>
+                        run.id !== previouslyActiveRun.id &&
+                        run.id !== previouslyInactiveRun.id
+                );
+                expect(newlyCreatedRun.length).toBe(1);
```
rmeritz, line 639: Can you actually test that they the correct values.

rmeritz (changes requested):
Just some random detail oriented comments.

server/tests/unit/services/RunService.test.js
evmiguel: Thank you! Good catch

spectranaut:
Thx for all this! the one additional test I'd like to see is: 
```
const activeRuns = await RunService.configureRuns({
    test_version_id: testVersion.id,
    apg_example_ids: [apgExample.id],
    at_browser_pairs: [
        {
            at_name_id: at.at_name_id,
            at_version: atVersionNumber,
            browser_id: browser.id,
            browser_version: browserVersionNumber
        }
        {
            at_name_id: at.at_name_id,
            // The only difference between this at_browser_pair and the one above is that the atVersionNumber is different!
            // We want to allow two different runs that are exactly the same other than a version number, in case there are still
            // "runs" for an older AT that need to be moved from "draft" to "final" while a new version of the AT is being tested.
            at_version: atVersionNumber2,
            browser_id: browser.id,
            browser_version: browserVersionNumber
        }

    ],
    run_status_id: runStatus.id
});
```

server/tests/unit/services/RunService.test.js
evmiguel: Good catch! Thank you

server/tests/unit/services/RunService.test.js
evmiguel: Good catch! I honestly have no idea why I did that 😅

server/tests/unit/services/RunService.test.js
evmiguel: https://github.com/w3c/aria-at-app/pull/174/commits/2c11bfd661be27d3010bcaf51c94e68e6aa3ad3a#diff-3e074bfb22ba34ca6d7b9412c4ae20eeR626

server/tests/unit/services/RunService.test.js
```diff
@@ -230,6 +228,139 @@ describe('RunService', () => {
...
+                    at_id: at.id,
+                    test_cycle_id: testCycle.id,
+                    browser_version_to_at_versions_id: tech.id,
+                    apg_example_id: apgExample.id,
+                    test_version_id: testVersion.id,
+                    active: false,
+                    run_status_id: runStatus.id
+                });
+
+                const activeRuns = await RunService.configureRuns({
```
evmiguel, line 296: @spectranaut here's the test!

spectranaut (approved):
It all looks good to me!

PR #175, opened 2020-10-07
Remove cycle specific pages and use new runs endpoint for test queue

spectranaut:
This PR does the following:
- Removes the "Manage Cycles" page
- Renames "Initiate Cycle" to "Configure Runs"
    - changes the "Initiate Cycle" button to "Update Active Run Configuration" and uses the new API end point to update active run configuration
- Removes the intermediary "test queue" page which lists all of the test cycles
- Makes the "test queue" page only show the active test via the ActiveRuns api end point

How to test:
- run the dev server
- go to the new "configure active runs" page
- make some runs
- After you click "Update Active Run Configuration", the screen will be blank -- refresh. This bug is related to the test results being in the cycle reducer -- will be fixed when we add results to the ActiveRuns
- check the database to see that only the new runs have the active flag (select * from runs)
- go to the run queue, see all the active runs
- change the active runs via the "configure active runs" page
- go back tot eh run queue, see all the active runs
- assign and unassign yourself to the run

What is left to do in DIFFERENT prs:
- [ ] You cannot run tests, that pages needs to be de-cycleid. To do this you need to (1) send the tests/test results up with the ActiveRun data, then use those test results in the TestQueue page and (2) move the save result endpoint from cycles in the backend and the cycle reducer in the front end.
- [ ] The "configure active runs" page also needs to be updated.


client/reducers/runs.js
```diff
@@ -0,0 +1,117 @@
...
+                }
+                return {
+                    ...state,
+                    activeRunsById: updatedActiveRunsById,
+                    publishedRunsById: updatedPublishedRunsById
+                };
+            }
+        }
+        // falls through
+        case SAVE_USERS_TO_RUNS: {
```
evmiguel, line 103: Known bug?

evmiguel (approved):
@spectranaut I ran the branch localyl, and it worked In my review, I looked at the following:

- Structural changes and additions to the Redux infrastructure
- Behavior of the reducers
- Changes to the front end components
- Changes to the routes on the front end
- Changes to the back end services

I approve this knowing that there are known bugs that will be improved later:
- The Test Queue screen isn't automatically refreshed
- Tests can't be run yet
- Configure active runs page has not been updated

client/reducers/runs.js
```diff
@@ -0,0 +1,117 @@
...
+                    updatedActiveRunsById = {
+                        ...this.state.activeRunsById,
+                        [runId]: updatedRun
+                    };
+                }
+                return {
+                    ...state,
+                    activeRunsById: updatedActiveRunsById,
+                    publishedRunsById: updatedPublishedRunsById
+                };
```
evmiguel, line 99: I'm thinking there should be a default case here for a clean return:

```
return { ...state }
```

client/components/TestQueueRun/index.jsx
```diff
@@ -9,7 +9,7 @@ import {
     deleteUsersFromRun,
```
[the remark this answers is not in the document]
spectranaut: right lol! shows cycles was really an add-on

client/reducers/runs.js
spectranaut: good idea

PR #176, opened 2020-10-07
Run configuration now shows existing active runs

spectranaut:
How to test:
- run the dev server
- go to the new "configure active runs" page
- make some runs
- click "save run configuration" and see the popup saying a save is happening
- go to the run queue, see all the active runs
- go back to the "configure active runs" page and hard refresh, make sure configuration reloads
- Change the configuration, check that the test queue reflects it, and hard refresh to make sure the changes maintain.


evmiguel (changes requested):
The changes around the ConfigureActiveRuns page and the related Redux infrastructure are good! The only issue is in the TestQueue. When I add an AT/Browser combination that has a different AT version but same browser version, the filtering in the TestQueue squashes the runs under only one AT/browser pair.

Here's where I believe the error starts:
https://github.com/w3c/aria-at-app/blob/main/client/components/TestQueue/index.jsx#L154

Here's the video:
https://www.loom.com/share/43e491e4ec414a0fa82984016f80e105

spectranaut:
Fixed @evmiguel ! Thanks so much for catch that bug.

evmiguel:
Great! Looks good.

PR #177, opened 2020-10-08
Add tests and their results as return values from run API calls

rmeritz:
See jsdoc for more information on the spec followed. It was largely cribbed from the existing `getTestsForRunsForCycle` cycle service

### DIFF
--- a/server/services/RunService.js
+++ b/server/services/RunService.js
@@ -486,7 +486,16 @@ async function getPublishedRuns() {
             where: {},
             include: [
                 { model: db.RunStatus, where: { name: db.RunStatus.FINAL } },
-                db.ApgExample,
+                {
+                    model: db.ApgExample,
+                    include: {
+                        model: db.Test,
+                        include: {
+                            model: db.TestResult,
+                            include: db.TestStatus
+                        }
+                    }
+                },
                 {
                     model: db.BrowserVersionToAtVersion,
                     include: [
--- a/server/services/RunService.js
+++ b/server/services/RunService.js
@@ -411,7 +411,11 @@ async function sequelizeRunsToJsonRuns(sequelizeRuns) {
             run_status: run.RunStatus.name,
             test_version_id: run.test_version_id,
             testers: run.Users.map(u => u.id),
-            tests: run.ApgExample.Tests.reduce((acc, test) => {
+            tests: run.ApgExample.Tests.filter(test =>
+                (test.TestToAts || []).some(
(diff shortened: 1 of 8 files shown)


server/services/RunService.js
```diff
@@ -393,7 +410,27 @@ async function sequelizeRunsToJsonRuns(sequelizeRuns) {
             run_status_id: run.run_status_id,
             run_status: run.RunStatus.name,
             test_version_id: run.test_version_id,
-            testers: run.Users.map(u => u.id)
+            testers: run.Users.map(u => u.id),
+            tests: run.ApgExample.Tests.reduce((acc, test) => {
```
spectranaut, line 414: getting the tests for a run is a little more complicated that this -- you need to use TestToAt to make sure that any test applies to a run with a specific AT. Not all test files for an APG pattern apply to all ATs.

For example, of all Checkbox APG example tests, 2 of them do not apply to VoiceOver, so those test need to be skipped in all the runs of VoiceOver.

server/services/RunService.js
```diff
@@ -393,7 +410,27 @@ async function sequelizeRunsToJsonRuns(sequelizeRuns) {
...
+                acc.push({
+                    id: test.id,
+                    file: test.file,
+                    name: test.name,
+                    execution_order: test.execution_order,
+                    results: test.TestResults.filter(
+                        testResult => testResult.run_id == run.id
```
spectranaut, line 421: this might take care for that for results? but you still will see a test with no results.

spectranaut (changes requested):
One more detail of the tests needs to be accounted for! Otherwise looking really good.

evmiguel (commented):
Reviewed, waiting for Val's requests to be fulfilled.

rmeritz:
@spectranaut || @evmiguel - I fixed the issue with non-relavant tests being included in the results. See the last commit.

server/services/RunService.js
```diff
@@ -393,7 +410,31 @@ async function sequelizeRunsToJsonRuns(sequelizeRuns) {
             run_status_id: run.run_status_id,
             run_status: run.RunStatus.name,
             test_version_id: run.test_version_id,
-            testers: run.Users.map(u => u.id)
+            testers: run.Users.map(u => u.id),
+            tests: run.ApgExample.Tests.filter(test =>
+                (test.TestToAts || []).some(
```
evmiguel, line 415: This change should handle the change request @spectranaut

evmiguel (approved):

PR #179, opened 2020-10-13
Wrap configureRuns() transaction

evmiguel:


client/components/TestQueue/index.jsx
```diff
@@ -142,7 +142,13 @@ class TestQueue extends Component {
         const atBrowserRunSets = [];
```
evmiguel, line 142: This came in with lint fixes.

server/services/RunService.js
```diff
@@ -105,17 +105,24 @@ const db = require('../models/index');
...
-    test_version_id,
-    apg_example_ids,
-    at_browser_pairs
-}) {
+async function configureRuns(
+    { test_version_id, apg_example_ids, at_browser_pairs },
+    commit = true
+) {
     try {
+        await db.sequelize.query('BEGIN;');
```
evmiguel, line 113: This was the only way I could create a transaction that was compatible with the `ROLLBACK` in `dbCleaner`

spectranaut: Unless we want to change the way we do dbCleaner, I think we might as well run with this -- but I agree it's unsatisfying!

rmeritz:
I think this is a legit solution! LGTM.

PR #180, opened 2020-10-13
Remove cycle from test run

spectranaut:
You should be able to executes tests and save results and raise issues and see conflicts! All functionality on test queue and test execution page should work.


evmiguel (approved):
Lgtm!

PR #181, opened 2020-10-14
Copy testers from older runs that are not final in configureRuns()

evmiguel:
Asana issue: https://app.asana.com/0/1193055321453706/1197883166761503

To test:

- Create an initial active run configuration
- Go to the test queue and assign some testers to different runs
- Select one run to be marked as Published by running one test, saving the results, then clicking "Mark as Draft", refresh, then click "Mark as final". Refresh
- Go back to the Configure Active Runs page and change the "Git Commit of Tests" to a more recent version
- Go back to the test queue to confirm that the testers in the runs that aren't final are copied over to the new runs

### DIFF
--- a/server/services/RunService.js
+++ b/server/services/RunService.js
@@ -275,6 +275,12 @@ async function configureRuns(
             ]
         });
 
+        // Capture value of currently active runs for copying testers
+        const currentActiveRuns = await db.Run.findAll({
+            where: { active: true },
+            include: [db.Users, db.ApgExample]
+        });
+
         await db.Run.update(
             { active: false },
             {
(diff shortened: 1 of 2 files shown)


spectranaut:
I'm actually seeing a bug, it's a little tricky to set up the database to an older test version now that we can only move forward, but the way I did it was this:
```
update run set active='f';
update run set active='t' where run.test_version_id = OLDVERSION and run.browser_version_to_at_versions_id in (select id from browser_version_to_at_versions where active='t');

update test_version set active='f';
update test_version set active='t' where id = OLDVERISION;
```

When I did this, then made sure testers were assigned to some runs, and then updated the test version, I didn't see those testers.

server/services/RunService.js
```diff
@@ -376,7 +382,37 @@ async function configureRuns(
                 };
             });
 
-            await db.Run.bulkCreate(dbRuns);
+            const newRuns = await db.Run.bulkCreate(dbRuns);
+
+            // Add tester to the runs that match on ApgExample, ATVersion, and BrowserVersion
```
spectranaut, line 387: We only want to copy over the testers if the test_version changes, otherwise we run through all this code unnecessarially because you will never find a matchIndex (because if you aren't changed the test_version, there is a uniqueness to the "browser_version_to_at_version.id" and "apg_example.name").

So maybe should should block off all this code based on whether there is a test_version change?

spectranaut (changes requested):
This looks good and works good, just one suggestion!

server/services/RunService.js
evmiguel: Good idea! I'll implement this.

server/services/RunService.js
```diff
@@ -376,7 +382,40 @@ async function configureRuns(
                 };
             });
 
-            await db.Run.bulkCreate(dbRuns);
+            const newRuns = await db.Run.bulkCreate(dbRuns);
+
+            // Copy over testers on test version change
+            if (currentActiveRuns.length > 0 &&  currentActiveRuns[0].test_version_id !== test_version_id) {
```
evmiguel, line 388: @spectranaut here's the condition that copies testers on a version change

PR #182, opened 2020-10-15
Fix Test Reports Pages to work without Cycles

rmeritz:
### Summary of changes
- Add new reducer for `getPublishedRuns`
- Remove use of Cycle reducers from the the `ResultsPage` and `RunResultsPage` components
- Adjust `getNewTestVersions` service to be a `getTestVersions` service and do the filtering to only show new test versions in the `ConfigureActiveRuns` component

### QA Instructions:
- Ensure you have publshed tests locally (Create active runs, runs some runs, marks some runs as final (you don't need to actually complete the runs))
- Go to "Test Reports" page. Does it work? Do all the links on it work?


evmiguel:
@rmeritz can you include QA instructions?

evmiguel:
@rmeritz the change works, but when I go to the page to refresh, the page data gets lost: https://www.loom.com/share/3b2539206cfb4d23bb99f66ae319058a

I think it has something to do with the data in the initial state and how it's loaded.

server/services/RunService.js
```diff
@@ -631,20 +631,10 @@ async function getActiveRunsConfiguration() {
...
-        });
-        let where = {};
-        if (activeTestVersion) {
-            where = {
-                datetime: { [db.Sequelize.Op.gte]: activeTestVersion.datetime }
-            };
-        }
         return await db.TestVersion.findAll({
-            where: where,
+            where: {},
```
evmiguel, line 637: Curious as to why the filtering was removed here and moved to the frontend?

client/components/ConfigureActiveRuns/index.jsx
```diff
@@ -244,16 +244,22 @@ class ConfigureActiveRuns extends Component {
...
-                        <option key={version.id} value={version.id}>
-                            {version.git_hash.slice(0, 7) +
-                                ' - ' +
-                                version.git_commit_msg.slice(0, 30) +
-                                '...'}
-                        </option>
-                    );
-                })}
+                {testVersions
+                    .filter(
```
evmiguel, line 248: Same question but place in front end code. Why are we filtering here, just curious

server/services/RunService.js
rmeritz: Because we need the old tests too in order to show historical versions on the test reports pages.

client/components/ConfigureActiveRuns/index.jsx
rmeritz: Because now we are passing all the test versions. The other option was to add a second endpoint but we decided against it.

rmeritz:
> @rmeritz the change works, but when I go to the page to refresh, the page data gets lost:

Good catch. I fixed it.

evmiguel (approved):
LGTM!

PR #190, opened 2020-10-19
Set latest test version at this point in time to active

evmiguel:


spectranaut (approved):
This migration looks good! If you want to go with this instead of handling the lack of test_version on the front end, I think that will be fine.

rmeritz:
Using the other solution instead.

PR #192, opened 2020-10-21
Add Test Navigator and Pagination to the Test Runner Page

isaacdurazo:
- Adding Test Navigator markup and styling to Test Run page
- Add bootstrap pagination to test run page
- Track when changes are dirty to be able to distinguish "skipped" tests from "in progress" tests

:exclamation: I spent a day trying to wire up these new buttons and the buttons in the top nav to be able to save partial results. I attempted to do this in the `TestIframe` on `componentWillUnmount` and via the `react-router` but I ran into issues with both approaches. I found a bug already in `main` that provides the warning "Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method." It seems like by the time we are trying to save the state is already gone. Fixing the saving partial results will need to be done in a follow up PR.

:notebook: Improving the styling of the bottom pagination is also incomplete because @isaacdurazo is planning on rethinking this entire section in a follow up ticket.

### DIFF
--- a/client/components/TestRun/TestRun.css
+++ b/client/components/TestRun/TestRun.css
@@ -1,5 +1,9 @@
 /* Test Navigator */
 .test-navigator {
+    /* display: none; */
+}
+
+button.test-navigator-toggle.show {
     display: none;
 }
 
--- a/client/components/TestRun/TestRun.css
+++ b/client/components/TestRun/TestRun.css
@@ -17,12 +17,12 @@ button.test-navigator-toggle.hide {
 }
 
 button.test-navigator-toggle.hide:before {
-    content: "<<";
+    content: '<<';
     position: absolute;
     left: 0;
-    font-size: .75em;
+    font-size: 0.75em;
     font-weight: 700;
-    top: .25em;
+    top: 0.25em;
 }
 
 button.test-navigator-toggle.show {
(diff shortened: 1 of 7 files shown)


client/components/TestRun/index.jsx
```diff
@@ -192,14 +209,97 @@ class TestRun extends Component {
...
+                                    onClick={this.toggleTestNavigator}
+                                    className="test-navigator-toggle"
+                                >
+                                    Show Test Navigator
+                                </button>
+                            )}
+                        </aside>
+                        <button
+                            onClick={this.toggleTestNavigator}
+                            className="test-navigator-toggle show"
```
rmeritz, line 292: @isaacdurazo - The way I did "show" looks weird. I cannot tell if it is the `css` or my implentation.

client/components/TestRun/index.jsx
```diff
@@ -192,14 +209,97 @@ class TestRun extends Component {
...
+                                                ) {
+                                                    resultClassName =
+                                                        'complete';
+                                                }
+                                            } else if (
+                                                this.state.currentTestIndex -
+                                                    1 >
+                                                i
+                                            ) {
+                                                resultClassName = 'skipped';
```
rmeritz, line 258: @isaacdurazo - I'm not sure how we should define what "skipped" means. I wonder if it should show up separately at all. 

I took a stab at it but we should discuss.

rmeritz:
@isaacdurazo - I'm not sure how the behavior of this and the new pagination buttons are supposed to interact with the "Partial Results" work @evmiguel has been working on. Let's discuss this.

client/components/TestRun/index.jsx
rmeritz: Remove skipped.

client/components/TestRun/index.jsx
```diff
@@ -192,14 +209,98 @@ class TestRun extends Component {
                 </Helmet>
                 <Container fluid>
                     <Row>
-                        <Col>{heading}</Col>
+                        <aside className="col-md-3 test-navigator">
```
rmeritz, line 212: @spectranaut - Do you know the best way with `aria` to markup the table of contents?

I think we might need something for the `aside` and/or `ol` I definitely want to add a way to represent the result state for each `li` which is currently on in the icon.  

Do if you don't "know" is there a recommended way you would look something like this up?

rmeritz:
This is now done except for the open questions on the PR and implementing auto-saving partial results when you navigate to another PR which depends on the work @evmiguel is doing now.

client/components/TestRun/TestRun.css
```diff
@@ -1,13 +1,166 @@
...
+    width: 2px;
+    height: 15px;
+    left: 4px;
+    top: 2px;
+    background: white;
+}
+
+.show {
+    width: 2em;
+}
```
spectranaut, line 68: hey @isaacdurazo , do you know why this is here? it's causing the toast to look like this: 
problem-with-toast

spectranaut (approved):
Added commit to fix a few things

PR #193, opened 2020-10-21
Save partial results

evmiguel:
This change is for saving partial results from the ARIA-AT App side. The following changes have been made:

- The removal of modals on Previous and Skip that warn about saving. Since there are partial results being saved, these modals are no longer relevant
- Saving of serialized form of iframe for skipped tests
- React state in `DisplayTest` modal that waits for the skipped test results to be saved
- Unit tests for `TestService`

To Test:
~~- Import the tests in PR branch of the `aria-at`: `yarn workspace server db-import-tests:dev -c c13791257b7bcb5258a2df6ecf24953898777295`~~
- Run the dev ARIA-AT App server and run a test from the test queue
- Fill in some data on the test iframe, but do not fill it out completely.
- Hit 'Skip test'.
- Hit 'Previous'
- Verify that the partial data entered previously is there

### DIFF
--- a/client/components/DisplayTest/index.jsx
+++ b/client/components/DisplayTest/index.jsx
@@ -81,6 +81,8 @@ class DisplayTest extends Component {
                 break;
             }
             case 'goToNextTest': {
+                // Save the serialized form of iframe
+                await this.iframe.current.processResults(null);
                 displayNextTest();
                 break;
             }
@@ -120,19 +122,11 @@ class DisplayTest extends Component {
     }
 
     handleNextTestClick() {
-        if (this.testHasResult) {
-            this.performButtonAction('goToNextTest');
-        } else {
-            this.confirm('goToNextTest');
-        }
+        this.performButtonAction('goToNextTest');
     }
 
     handlePreviousTestClick() {
-        if (this.testHasResult) {
-            this.performButtonAction('goToPreviousTest');
-        } else {
-            this.confirm('goToPreviousTest');
-        }
+        this.performButtonAction('goToPreviousTest');
     }
 
     handleRedoClick() {
@@ -193,14 +187,6 @@ class DisplayTest extends Component {
         if (this.buttonAction === 'closeTest') {
             modalTitle = 'Close';
             action = `You are about to leave this test run. ${cannotSave}`;
-        }
-        if (this.buttonAction === 'goToNextTest') {
-            modalTitle = 'Skip Test';
-            action = `You are about to move to the next test. ${cannotSave}`;
-        }
-        if (this.buttonAction === 'goToPreviousTest') {
-            modalTitle = 'Previous Test';
-            action = `You are about to move to the previous test. ${cannotSave}`;
         }
         if (this.buttonAction === 'redoTest') {
             modalTitle = 'Start Over';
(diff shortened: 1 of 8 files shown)


client/components/DisplayTest/index.jsx
```diff
@@ -81,7 +82,7 @@ class DisplayTest extends Component {
                 break;
             }
             case 'goToNextTest': {
-                displayNextTest();
+                this.iframe.current.triggerSkip();
```
spectranaut, line 85: I actually think this is probably not a good thing to add -- because we don't need any information from the iframe at all, we really just want to grab the serialized form and send save a "incomplete" result in the `test_result` table. I can see how you are trying to hook into the same code for saving that is used later on, but I think it would be simplier to just have a different code path that calls `SaveResult({.... status: "incomplete"})`

spectranaut (changes requested):
Hey Erika, looks good except three things:

1. I think we really shouldn't do the windowListener convo with the iframe in this case -- I closed the PR on aria-at and left an explanation why: https://github.com/w3c/aria-at/pull/308#pullrequestreview-517107976. Could you update this PR to not use it?

client/components/DisplayTest/index.jsx
```diff
@@ -120,19 +122,11 @@ class DisplayTest extends Component {
...
+        this.performButtonAction('goToNextTest');
     }
 
     handlePreviousTestClick() {
-        if (this.testHasResult) {
-            this.performButtonAction('goToPreviousTest');
-        } else {
-            this.confirm('goToPreviousTest');
-        }
+        this.performButtonAction('goToPreviousTest');
```
spectranaut, line 129: In a follow up, "goToPreviousTest" should call triggger the "this.iframe.current.processResults"

client/components/DisplayTest/index.jsx
```diff
@@ -367,8 +355,20 @@ class DisplayTest extends Component {
...
+        } else if (this.testResultsSkipped) {
+            testContent = (
+                <TestIframe
+                    git_hash={git_hash}
+                    file={test.file}
+                    at_key={at_key}
+                    onResults={this.handleResults}
+                    ref={this.iframe}
+                    serializedForm={test.results[testerId].serialized_form}
+                    onSave={this.onSave}
```
spectranaut, line 369: I think you no long have this function, right?

client/components/TestIframe/index.jsx
```diff
@@ -122,5 +129,7 @@ TestIframe.propTypes = {
     onResults: PropTypes.func,
     git_hash: PropTypes.string,
     file: PropTypes.string,
-    at_key: PropTypes.string
+    at_key: PropTypes.string,
+    onSave: PropTypes.func,
```
spectranaut, line 133: I think this can be remove here too?

spectranaut (changes requested):
Two minor things, otherwise works great!

client/components/DisplayTest/index.jsx
evmiguel: oh right! removing

client/components/TestIframe/index.jsx
evmiguel: Yes!

spectranaut (approved):

PR #195, opened 2020-10-27
Save partial results other buttons

evmiguel:
Partial results should now be saved when the following buttons are clicked:

- Skip
- Previous
- Close
- Next

### DIFF
--- a/client/components/DisplayTest/index.jsx
+++ b/client/components/DisplayTest/index.jsx
@@ -77,25 +77,34 @@ class DisplayTest extends Component {
                 break;
             }
             case 'closeTest': {
+                // Save the serialized form of iframe
+                if (this.iframe.current) {
+                    await this.iframe.current.processResults(null);
+                }
                 history.push(`/test-queue`);
                 break;
             }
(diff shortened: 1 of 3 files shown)


rmeritz:
@evmiguel - This should also work for other clicks off the page. i.e. all the buttons in the top nav. We can make another ticket for that if you prefer. 

I tested that the other buttons are working though!

client/components/DisplayTest/index.jsx
```diff
@@ -182,11 +191,11 @@ class DisplayTest extends Component {
         } = this.props;
 
         let modalTitle, action;
-        let cannotSave = `Test ${testIndex} has not been completed in full and your progress on this test won’t be saved.`;
+        let modalWarning = `Test ${testIndex} has not been completed in full. Your progress on this test will be saved.`;
```
rmeritz, line 194: I think this modal should just be removed. I asked in Slack.

rmeritz:
Looks good. I see you made another ticket for the other followup! Thanks.

PR #196, opened 2020-10-27
Change configure AT/Browser rows to not have form controls.

evmiguel:
Asana issue: https://app.asana.com/0/1193055321453706/1198879584985510

This change:
- Sets a row with a configured AT/Browser pair to have no inputs
- Changes the "Delete" button to "Remove"
- Changes the "Delete" column to "Action"

Test:
- Go to the Configure Active Runs page
- Verify that the configured AT/Browser pairs are rows without inputs
- Configure a new AT/Browser pair and save
- Verify that the new pair is a row without inputs

### DIFF
--- a/client/components/ConfigureActiveRuns/index.jsx
+++ b/client/components/ConfigureActiveRuns/index.jsx
@@ -40,7 +40,8 @@ function getDefaultsTechCombinations(testVersion, activeRunConfiguration) {
                 at_id,
                 at_version: combo.at_version,
                 browser_id: combo.browser_id,
-                browser_version: combo.browser_version
+                browser_version: combo.browser_version,
+                editable: false
             });
         }
     }
@@ -227,7 +228,7 @@ class ConfigureActiveRuns extends Component {
 
     addTechnologyRow() {
         let newRunTechnologies = [...this.state.runTechnologyRows];
-        newRunTechnologies.push({});
+        newRunTechnologies.push({editable: true});
         this.setState({
             runTechnologyRows: newRunTechnologies
         });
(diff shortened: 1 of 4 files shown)


spectranaut (changes requested):
One user feedback: This looks great! But the I noticed if you fill in a row completely, the row becomes uneditable. It should not become uneditable because the purpose is to be able to tell the difference between existing run technology pairs and newly configure runs. When you click "save", everything should become uneditable, but not until then.

client/components/ConfigureActiveRuns/index.jsx
```diff
@@ -393,6 +395,9 @@ class ConfigureActiveRuns extends Component {
                                                     deleteTechnologyRow={
                                                         this.deleteTechnologyRow
                                                     }
+                                                    editingRow={
+                                                        this.state.editingRow
```
spectranaut, line 399: Actually it's not quite this simple -- the purpose of having some rows look uneditable is because those rows are part of the previous configuration. We want the user to be able to tell which rows they added and which already exist up until the point they save.

For the code, this means the rows that come directly from the "activeRunConfiguration" global state should always look uneditable, though they can be deleted. Any new rows a user adds (before they click save) should remain editable until they click "save".

One way to implement this is to add a new flag called "editable" to all of the objects in `this.state.runTechnologyRows`. This attribute would set to false for all the at/browser pairs returned by  `getDefaultsTechCombinations`, and set to true for all new rows added (see the "push" in `handleTechnologyRowChange`). Then you can pass that value in here (`runTech.editable`).

spectranaut (changes requested):

client/components/ConfigureActiveRuns/index.jsx
evmiguel: Good catch! I didn't see that the uneditable rows become editable. I'm on it!

PR #197, opened 2020-10-28
Delete tests for user and run

spectranaut:
Adds functionality to the "delete" buttons on the test queue.

Test as both and admin and as a tester.


server/models/.#UserToAt.js
```diff
@@ -0,0 +1 @@
+spectranaut@val-bocoup.11049:1603728167
```
evmiguel, line 1: Can we delete this file?

evmiguel (approved):
Works for me! Only things are to fix the build and maybe delete the file I commented on.

PR #198, opened 2020-10-28
Configure active runs should respect apg_examples selection

spectranaut:
On the configure run page, select and unselect APG patterns then click "save" -- it should work now!


evmiguel (commented):
LGTM

PR #202, opened 2020-11-04
Show run results that will be deleted on configuration changes

evmiguel:
I wasn't quite sure how to describe QA instructions. Here is a Loom for reference:

https://www.loom.com/share/8b19cddf105c4544bf14241d8dcde2c2

### DIFF
--- a/client/components/ConfigurationModal/index.jsx
+++ b/client/components/ConfigurationModal/index.jsx
@@ -1,0 +1,68 @@
+import React, { Component } from 'react';
+import { Button, Modal } from 'react-bootstrap';
+import PropTypes from 'prop-types';
+
+class ConfigurationModal extends Component {
+    render() {
+        const {
+            show,
+            handleClose,
+            saveRunConfiguration,
+            configurationChanges,
+            resultsDeleted
+        } = this.props;
+        return (
+            <Modal show={show} onHide={handleClose}>
+                <Modal.Header closeButton>
+                    <Modal.Title>Update Active Test Runs</Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    {resultsDeleted ? (
+                        <React.Fragment>
+                            <p>
+                                <b>
+                                    The draft results for the following test
+                                    plans will be lost:
+                                </b>
+                            </p>
+                            <ul>
+                                {configurationChanges.map(runDeleted => (
+                                    <li
+                                        key={runDeleted.id}
+                                    >{`${runDeleted.apg_example_name} - ${runDeleted.at_name} ${runDeleted.at_version} with ${runDeleted.browser_name} ${runDeleted.browser_version}`}</li>
+                                ))}
+                            </ul>
+                        </React.Fragment>
+                    ) : (
+                        <React.Fragment>
+                            <p>
+                                This new configuration does not remove any runs
+                                that are in a draft state, but it might delete
+                                runs that have test results. Are you sure you
+                                want to make this change?
+                            </p>
+                        </React.Fragment>
+                    )}
+                </Modal.Body>
+                <Modal.Footer>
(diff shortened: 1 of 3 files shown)


client/components/ConfigureActiveRuns/index.jsx
```diff
@@ -203,7 +341,34 @@ class ConfigureActiveRuns extends Component {
...
+            );
+        }
+    }
+
+    saveNewConfiguration() {
+        this.props.dispatch(saveRunConfiguration(this.state.newConfiguration));
+
+        this.closeChanges();
 
         alert('You saved results! (...what should we do after saving?)');
```
spectranaut, line 373: We should delete this alert, it was just a place holder

client/components/ConfigureActiveRuns/index.jsx
```diff
@@ -135,8 +162,118 @@ class ConfigureActiveRuns extends Component {
...
+            showChangeModal: false
+        });
+    }
+
+    /** 
+     * This is a naive algorithm for finding all the configuration changes. 
+     * It can be vastly improved, especially with smarter logic on handling
+     * duplicate changes.
+     * */
+    setConfigChanges(newConfig) {
```
spectranaut, line 182: this should be called "getConfigChanges" or "calculateConfigChanges" because it doesn't set any thing, it just does some data munging and returns the result!

client/components/ConfigureActiveRuns/index.jsx
```diff
@@ -203,7 +341,34 @@ class ConfigureActiveRuns extends Component {
...
+            this.setState(
+                {
+                    configurationChanges,
+                    newConfiguration: config
+                },
+                () => {
+                    this.showChanges();
+                }
+            );
+        } else {
```
spectranaut, line 356: If there are no changes, we should have a popup that says there are no changes -- rather than just try to save, right? I feel a bit strange about this else here.

I think we should only save if you click the confirmation on the modal.

spectranaut (changes requested):
Like I mentioned, there is a bug preventing the modal from ever appear for me that I'm pretty sure is in setConfigChanges -- but also looking at the code there is a few other things I think should go in this PR, like we shouldn't except after confirming in the modal.

Edit: no bug, it only warns you about runs that are in the "DRAFT" state

client/components/ConfigurationModal/index.jsx
```diff
@@ -0,0 +1,53 @@
...
+                        {configurationChanges.map(runDeleted => (
+                            <li
+                                key={runDeleted.id}
+                            >{`${runDeleted.apg_example_name} - ${runDeleted.at_name} ${runDeleted.at_version} with ${runDeleted.browser_name} ${runDeleted.browser_version}`}</li>
+                        ))}
+                    </ul>
+                </Modal.Body>
+                <Modal.Footer>
+                    <Button variant="secondary" onClick={handleClose}>
+                        Close
```
spectranaut, line 35: This should be "Cancel" not "Close"

client/components/ConfigureTechnologyRow/index.jsx
```diff
@@ -78,116 +81,124 @@ class ConfigureTechnologyRow extends Component {
...
+                        aria-label={`Delete at/browser combination ${index +
+                            1}`}
+                        onClick={this.deleteRun}
+                        disabled={deleted === true}
+                    >
+                        Remove
+                    </Button>
+                    <Button
+                        disabled={deleted === false}
+                        onClick={this.undoDelete}
```
spectranaut, line 129: we need an "aria-label" here that says "Undo delete", otherwise there is no way for a screen reader user ot know what this button does!

spectranaut (changes requested):

spectranaut (approved):
awesome! works great :)

PR #205, opened 2020-11-05
Fixes for partial results (do not save when no data, save when navigating to different tests)

spectranaut:
This PR does three things mostly:
1. Only saves serialized data when data has been added to form
2. When you click "redo", actually sets the test status to "skipped" so it shows up with an empty circle in the test navigation
3. When you click to a random test from either test navigation, the serialized form is saved just like with "next test" and "previous test"

To do (3), I felt like we need to refactor the TestRun and DisplayTest components. It doesn't really make sense the way they are broken up. I decided to try to combine them as a first step (and later break things off into reasonable components). The problem is that the test nav needs functionality only in it's child, DisplayTest (that is, to save partial results). So I got rid of Display Test. 

To test:
Try editing and saving and starting over and redoing. Essentially try every single button and way of filling things out, saving, and navigating around the tests.


client/components/TestIframe/index.jsx
```diff
@@ -98,6 +103,11 @@ class TestIframe extends Component {
         // inside the iframe
         window.addEventListener('message', this.handleResultsMessage);
 
+        // Before hydrating, save local copy of serialized state with no results
+        // This is a short cut to tell whether the form has been edited
+        await this.waitForTestHarnessReload();
+        this.emptyForm = JSON.stringify(this.serializeForm());
```
spectranaut, line 109: This is not always the "emptyForm" of the test you currently see in the iframe. The problem is we re-render when we have a new "testFile" coming in from the props, but maybe componentDidMount is not called again? But in this case, the tests are not be rehydrated. I'm not sure how I broke this in the refactor but I've run out of time to look into it now :(

evmiguel (commented):
Looks good to me, so far! I confirm that there is a bug with the "Edit Results". Let me know when that is ready

evmiguel (approved):

s3ththompson:
@spectranaut didn't get a chance to look until now, but i think this approach makes sense! 👍

PR #213, opened 2020-11-20
Handle errors when there is no configured data

evmiguel:
On a fresh database with no active runs, there were errors in rendering the "Configure Active Runs" page. This can be implemented better, but @dawnwages needs this to get up and running.

### DIFF
--- a/client/components/ConfigureActiveRuns/index.jsx
+++ b/client/components/ConfigureActiveRuns/index.jsx
@@ -18,19 +18,26 @@ import './ConfigureActiveRuns.css';
 
 function selectExamples(testVersion, activeRunConfiguration) {
     let exampleSelected = {};
-    for (let example of testVersion.apg_examples) {
-        let selected = true;
-        // If we are looking at the currently configured test version, select
-        // only the currently configured apg examples, otherwise select all by default.
-        if (
-            testVersion.id === activeRunConfiguration.active_test_version.id &&
-            activeRunConfiguration.active_apg_examples.indexOf(example.id) ===
-                -1
-        ) {
-            selected = false;
+
+    // Select examples if there is a configured test version
+    if (Object.keys(testVersion).length) {
+        for (let example of testVersion.apg_examples) {
+            let selected = true;
+            // If we are looking at the currently configured test version, select
+            // only the currently configured apg examples, otherwise select all by default.
+            if (
+                testVersion.id ===
+                    activeRunConfiguration.active_test_version.id &&
+                activeRunConfiguration.active_apg_examples.indexOf(
+                    example.id
+                ) === -1
+            ) {
+                selected = false;
+            }
+            exampleSelected[example.id] = selected;
         }
-        exampleSelected[example.id] = selected;
     }
+
     return exampleSelected;
 }
 
@@ -54,7 +61,7 @@ function getDefaultsTechCombinations(testVersion, activeRunConfiguration) {
         }
     }
 
-    return initialRunRows.length ? initialRunRows : [{}];
+    return initialRunRows.length ? initialRunRows : [];
 }
 
 class ConfigureActiveRuns extends Component {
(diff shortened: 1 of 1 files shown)


server/services/UsersService.js
```diff
@@ -249,7 +249,7 @@ async function saveUserAndRoles(options) {
             }
```
evmiguel, line 249: Lint fix

rmeritz:
@evmiguel - I think since @dawnwages isn't working on this now. I think we should not add anything to the code with "this is a hack". I think it would be better to just have @dawnwages do a better fix. 

What do you think the better solution is?

evmiguel:
> @evmiguel - I think since @dawnwages isn't working on this now. I think we should not add anything to the code with "this is a hack". I think it would be better to just have @dawnwages (or one of us) do a better fix.
> 
> What do you think the better solution is?

In the [initial state](https://github.com/w3c/aria-at-app/pull/213/files#diff-bf6af43baaa40d6e4b093436ee6862061595542cf09f26c00a3ae2b1f0a96a51R78) for `runTechnologyRows`, it should just be an empty array instead of an array with one empty object.

The empty object yields checking for the emptiness of the object with `Object.keys(this.state.runTechnologyRows[0]).length`, which could be cleaner if just checking for length. Then, I could remove [this line](https://github.com/w3c/aria-at-app/pull/213/files#diff-bf6af43baaa40d6e4b093436ee6862061595542cf09f26c00a3ae2b1f0a96a51R572) and just check for the length of the `runTechnologyRows` in [this line](https://github.com/w3c/aria-at-app/pull/213/files#diff-bf6af43baaa40d6e4b093436ee6862061595542cf09f26c00a3ae2b1f0a96a51R653).

I can fix it, since it will be quick.

client/components/App/App.css
```diff
@@ -52,9 +52,10 @@ nav.navbar {
     background: #cfcece;
```
evmiguel, line 52: Lint fix

client/components/ConfigureActiveRuns/ConfigureActiveRuns.css
```diff
@@ -6,19 +6,20 @@
 }
```
evmiguel, line 6: Lint fix

client/components/ConfigureActiveRuns/index.jsx
```diff
@@ -632,45 +698,22 @@ class ConfigureActiveRuns extends Component {
...
-                                                        .undoDeleteTechnologyRow
-                                                }
-                                                editable={runTech.editable}
-                                                deleted={runTech.deleted}
-                                            />
-                                        );
-                                    }
-                                )}
-                        </tbody>
+                        {technologyRows.length ? (
```
evmiguel, line 701: Hack fixed

client/components/ConfigureTechnologyRow/ConfigureTechnologyRow.css
```diff
@@ -1,23 +1,24 @@
-
```
evmiguel, line 1: Lint fix

client/components/ConfigureTechnologyRow/index.jsx
```diff
@@ -138,7 +138,7 @@ class ConfigureTechnologyRow extends Component {
                 <td>{runTechnologies.browser_version}</td>
```
evmiguel, line 138: Lint fix

client/components/TestQueue/TestQueue.css
```diff
@@ -1,4 +1,4 @@
-main.container-fluid table .row,
```
evmiguel, line 1: Lint fix

client/components/TestQueueRun/TestQueueRun.css
```diff
@@ -28,7 +28,7 @@ li {
 }
```
evmiguel, line 28: Lint fix

client/components/TestRun/index.jsx
```diff
@@ -440,7 +440,11 @@ class TestRun extends Component {
```
evmiguel, line 440: Lint fix

client/components/TestRun/TestRun.css
```diff
@@ -165,8 +165,7 @@ button.test-navigator-toggle.show:after {
     padding: 0;
```
evmiguel, line 165: Lint fix

client/components/TestQueueRun/index.jsx
```diff
@@ -1,7 +1,11 @@
 import React, { Component, Fragment } from 'react';
```
evmiguel, line 1: Lint fix

evmiguel:
@rmeritz all set!

PR #214, opened 2020-11-20
Visual polish

isaacdurazo:
Hi folks, here's my PR for the visual polish. Unfortunately, I couldn't finish the Test Queue. It was just too much work, but I think it's in a good state for the usability study.


rmeritz:
It looks really sharp!!!!!!

PR #216, opened 2020-11-25
New reports page implementation.

evmiguel:
It's missing the full styling. I'll follow up the file tree view styling in another PR.

Testing instructions:
- Create test results in as many test runs as you'd like
- Publish the results
- Go to `localhost:3000/reports` to see a table that sort of resembles [the wireframes](https://github.com/w3c/aria-at/issues/317)

Here's mine:
Screen Shot 2020-11-25 at 11 02 13 AM

### DIFF
--- a/client/components/ReportsPage/index.jsx
+++ b/client/components/ReportsPage/index.jsx
@@ -1,0 +1,297 @@
+import React, { Component, Fragment } from 'react';
+import { Helmet } from 'react-helmet';
+import { Table } from 'react-bootstrap';
+import { connect } from 'react-redux';
+import { getPublishedRuns } from '../../actions/runs';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faFolderOpen, faFolder } from '@fortawesome/free-solid-svg-icons';
+import { ProgressBar } from 'react-bootstrap';
+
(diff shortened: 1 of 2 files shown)


client/components/ReportsPage/index.jsx
```diff
@@ -0,0 +1,297 @@
...
+    dispatch: PropTypes.func,
+    publishedRunsById: PropTypes.object
+};
+
+const mapStateToProps = state => {
+    const { publishedRunsById } = state.runs;
+    return { publishedRunsById };
+};
+
+export default connect(mapStateToProps)(ReportsPage);
```
rmeritz, line 297: I feel really overwhelemed by the amount of (un-tested) logic in this component. And it is only slated to grow as we add more to this page. 

You are doing a lot of munging here that I think really belongs in a utitly and ought to be unit tested. 

I think we have the time aloted to get the reports page correct. And we are plan for this page to go through many iternations so I think its worth taking the time to do this now.

rmeritz:
Despite by above comment I think this is looking really good and I think we should get it merged to we can share it with Matt for feedback ASAP to to unblock other reports page work.

client/components/ReportsPage/index.jsx
evmiguel: +1 to this! I'll add tests in a follow up!

PR #218, opened 2020-11-25
Jest tests and refactor for Reports Page

evmiguel:


client/package.json
```diff
@@ -89,6 +90,9 @@
       "^@components(.*)": "<rootDir>/components/$1",
       "\\.(css|less)$": "<rootDir>/tests/__mocks__/styleMock.js",
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/tests/__mocks__/fileMock.js"
-    }
+    },
+    "setupFiles": [
+      "core-js"
```
evmiguel, line 95: This is because jest doesn't understand `Array.flat()`

rmeritz:
This is a huge improvement!!! 

Would it be possible to also move the calculating of the percentage to the matrix state? I think it makes sense to unit test that and not have it mixed in with the view logic if possible?

evmiguel:
> This is a huge improvement!!!
> 
> Would it be possible to also move the calculating of the percentage to the matrix state? I think it makes sense to unit test that and not have it mixed in with the view logic if possible?

Yup, I can do that!

PR #219, opened 2020-11-25
Add ability to filter by Tech Pair to reports page

rmeritz:
Firefox_Screenshot_2020-11-25T21-23-22 809Z
Firefox_Screenshot_2020-11-25T21-23-03 860Z

PR #220, opened 2020-11-27
Add ability to filter by Tech Pair to reports page (Take 2)

rmeritz:
This work was orignally implemented in: https://github.com/w3c/aria-at-app/pull/219. But two bugs were found one where the page crashed. This was caused by col and rows being switched which was easy to fix. The other bug was more serious and rows with `-` (without data) never were hidden. This was a fundamental design flaw. 

In this PR instead of iterating over the `techMatrix` add two new objects `techPairs` and `apgExamples` which a much similar data objects and use those to generate the table. This allows the rows to generated in one iteration and for them to be hidden without classes which much similar than the orginal.

### DIFF
--- a/client/components/ReportsPage/index.jsx
+++ b/client/components/ReportsPage/index.jsx
@@ -12,7 +12,6 @@ import {
     generateTechPairs,
     generateApgExamples,
     calculateTotalObjectPercentage,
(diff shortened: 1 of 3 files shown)


evmiguel (approved):
Lovely! LGTM.

PR #222, opened 2020-12-02
Update backend with apg links

evmiguel:
This PR includes:

- Migration to add two columns: `example`, `practice_guide` 
- Update `ApgExample` model to include the above columns
- Update import script (naive solution) to add values to the above columns


rmeritz:
Please do not add any new code that does SQL directly. Please use the Sequalize `ApgExample` orm to insert into the DB. 

Also, can you have more error handling if the `refernce.csv` file isn't there or formatting issues in it?

evmiguel:
@rmeritz Your feedback has been addressed!

spectranaut (approved):
This looks good!

PR #223, opened 2020-12-02
Add abilty to collapse/expand test data for each example

rmeritz:
- Only expand/collapse the APG examples and not the top-level because multiple levels of collapsing is confusing for screen readers and not that useful to sighted users
- Use the bootstrap-react `Collapse` utility
- Add a new `open` state to the `apgExamples` state
- Move the open/closed folder icons into a separate row and use them to control the expand collapse, but style them for visual users so they look like they are in the same column. We think having the controls in their own left-most column is best for screen-reader-users.

See it working here: https://www.loom.com/share/0f45d93b30244686bba872af0b4a3e75

### DIFF
--- a/client/components/ReportsPage/index.jsx
+++ b/client/components/ReportsPage/index.jsx
@@ -174,7 +174,7 @@ class ReportsPage extends Component {
                 let exampleRow = [];
                 exampleRow.push(
                     <td key={`example-${exampleName}`} className="border-bottom-0 border-right-0 pr-0">
-                        <span className="ml-3" onClick={() => this.setOpenExample(exampleIndex)} aria-expanded={open}>
+                        <span className="ml-3" onClick={() => this.setOpenExample(exampleIndex)} aria-expanded={open} aria-label={`${open ? 'Collapse' : 'Open'} ${exampleName} tests`}>
                             <FontAwesomeIcon icon={open ? faFolderOpen : faFolder} />
                         </span>
                     </td>
--- a/client/components/ReportsPage/index.jsx
+++ b/client/components/ReportsPage/index.jsx
@@ -1,6 +1,6 @@
 import React, { Component, Fragment } from 'react';
 import { Helmet } from 'react-helmet';
-import { Table, Button, Collapse } from 'react-bootstrap';
+import { Table, Collapse } from 'react-bootstrap';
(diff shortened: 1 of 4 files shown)


client/components/ReportsPage/index.jsx
```diff
@@ -65,6 +70,18 @@ class ReportsPage extends Component {
         });
     }
 
+    setOpenExample(i) {
+        this.setState({
```
evmiguel, line 74: This isn't a blocker, but the more React-y way to do this is to use the [`updater` function](https://reactjs.org/docs/react-component.html#setstate).

This example would then look something like:

```
this.setState(prevState => {
    return {
        apgExamples: [
               ...prevState.apgExamples.slice(0, i),
                Object.assign(prevState.apgExamples[i], {
                    open: !prevState.apgExamples[i]['open']
                }),
                ...prevState.apgExamples.slice(i + 1) 
        ]
  }
});
```

client/components/ReportsPage/index.jsx
```diff
@@ -205,7 +247,15 @@ class ReportsPage extends Component {
...
                     .filter(pair => pair.active)
-                    .forEach(({ browser, at }) => {
+                    .forEach(({ browser, at }, techPairIndex) => {
+                        if (techPairIndex === 0) {
+                            testStatHeaderRow.push(
+                                <td className="border-top-0 border-right-0"></td>
+                            );
+                            testStatHeaderRow.push(
+                                <td className="border-top-0 border-left-0"></td>
+                            );
```
evmiguel, line 257: Tested this with a screen reader, and this is causing some usability issues.

evmiguel (commented):
@rmeritz 

From a visual standpoint, this looks great! One non-blocking feedback on the visual front is that it is kind of when collapsing to close, but I think we can fix this later.

From a screen reader point of view, I think there are some accessibility issues. The table is not so navigable because of the new columns added. I was getting confused about which column the the data is in. https://www.loom.com/share/2c19d508ed6a43d6917bd827c09c2ab4

I'm ok with merging this in, but I think we should have a larger conversation about the accessibility of this table. Let me know your thoughts, and then we can merge!

dawnwages (approved):

PR #224, opened 2020-12-03
Add link to test details

rmeritz:
- Link to the existing details tables on the pages of the existing report
- Modify the existing anchors to use `execution_order` instead of `index` so they will definitely work across pages when if the order the test `tests` being iterated over is different
- Manually scroll to the anchor with the `componentDIdUpdate` use a solution heavily inspired by https://github.com/filippoitaliano/react-snorty-tips/blob/master/react/anchor-link.md

See it working here: https://www.loom.com/share/6972e49a8fea4956b54862f00cb1c6fc

### DIFF
--- a/client/components/RunResultsPage/index.jsx
+++ b/client/components/RunResultsPage/index.jsx
@@ -19,7 +19,7 @@ class RunResultsPage extends Component {
 
         this.handleRaiseIssueClick = this.handleRaiseIssueClick.bind(this);
 
-      this.$refs = {};
+        this.$refs = {};
     }
 
     async componentDidMount() {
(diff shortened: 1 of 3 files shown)


client/components/RunResultsPage/index.jsx
```diff
@@ -30,6 +32,18 @@ class RunResultsPage extends Component {
         }
     }
 
+    async componentDidUpdate() {
+        if (this.$refs && location.href.includes('#test-')) {
+            const anchor = location.href.split('#')[1];
+            if (anchor !== undefined) {
+                const ref = this.$refs[anchor];
+                if (ref !== undefined) {
+                    ref.scrollIntoView();
```
evmiguel, line 41: Nice, this is clever!

evmiguel (approved):
@rmeritz Looks good! Again, there are some accessibility things, but I'll make an Asana task for that

dawnwages (approved):

PR #225, opened 2020-12-03
Fix configure active runs

dawnwages:
combines two tasks regarding the commit drop down on ConfigureActiveRuns component
- Fix test version formatting on Active Run Configuration page: https://app.asana.com/0/1193055321453706/1199200817992352/f
- BUG: "Select a different commit" dropdown on the Configure Active Runs Page : https://app.asana.com/0/1193055321453706/1199342698891221/f

### DIFF
--- a/client/components/ConfigureActiveRuns/index.jsx
+++ b/client/components/ConfigureActiveRuns/index.jsx
@@ -393,18 +393,20 @@ class ConfigureActiveRuns extends Component {
             version => version.id === parseInt(event.currentTarget.value)
         )[0];
 
-        this.setState({
-            selectedVersion: versionData.id,
-            assignedTesters: [],
-            exampleSelected: selectExamples(
-                versionData,
-                activeRunConfiguration
-            ),
-            runTechnologyRows: getDefaultsTechCombinations(
-                versionData,
-                activeRunConfiguration
-            )
-        });
+        if (versionData) {
+            this.setState({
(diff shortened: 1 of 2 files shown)


evmiguel (changes requested):
@dawnwages good start! One thing to make this PR more complete is to make sure that the `Select a different commit` prompt only appear if there are new versions.

We don't want this:
Screen Shot 2020-12-03 at 4 47 31 PM

client/components/ConfigureActiveRuns/index.jsx
```diff
@@ -466,37 +468,46 @@ class ConfigureActiveRuns extends Component {
...
                 <option key={version.id} value={version.id}>
-                    {version.git_hash.slice(0, 7) +
+                    {moment(version.date).format('MMMM Do YYYY') +
                         ' - ' +
                         version.git_commit_msg.slice(0, 80) +
-                        '...'}
+                        '...' +
+                        ' - ' +
+                        version.git_hash.slice(0, 7) +
+                        ' - '}
```
spectranaut, line 477: This formatting seems a little strange to me but it's a minor thing we can fix later if someone knows what to change it too

spectranaut (approved):
looks good! I'm not sure if the formatting in the string in the drop down is the best but also I don't have a suggestion, we can have isaac look at it at some later point.

PR #226, opened 2020-12-04
Link to example and pattern

evmiguel:
- This PR is dependent on #222.
- Adds links to Design Pattern and APG Example on the top level APG Example Rows

Screen Shot 2020-12-04 at 1 40 44 PM


rmeritz:
@evmiguel - I merged to the reports page so you can follow up with this now and just add it to the top of the test plan pages for now.

evmiguel:
@rmeritz I rebased this on top of main. I'm not sure where the links will go on the new reports page. This PR is just surfacing the links in the API response for now.

PR #230, opened 2020-12-09
Restrict publish results to only when there are results

evmiguel:
Asana issue: https://app.asana.com/0/1193055321453706/1199380126013637


rmeritz:
@evmiguel - I still was able to publish no results in at least one corner case. 

1. Add results for at least one test for one tester
2. Delete those results
3. Publish

evmiguel:
> @evmiguel - I still was able to publish no results in at least one corner case.
> 
> 1. Add results for at least one test for one tester
> 2. Delete those results
> 3. Publish

Good catch! I'll handle that case

PR #233, opened 2020-12-10
Add a new simple reports

rmeritz:
#### Description
- This a new simplified version of the new reports page. 
- Based on these ["wireframes"](https://docs.google.com/spreadsheets/u/1/d/1jsvhue5xRxO03vvVxoqfljd_eihZcxKcz6HkhbSioD8/edit#gid=0) created after out discussion about redesigning this page with PAC.
- I reused and stripped down most of the code from the original, complex new reports page. The most important thing I preserved the marshalled data structure that we use to iterate over to show results. In the medium term I think it is a good solution, but it the long term it is not going to performant to be loading so much data that we don't need on any one page. 
- Remove the old main test report page `/results` but left the by run reports pages in place
- Fixed a bug so that runs with `updated_at = NULL` don't always show up as the most recently published version.
- This has been tested with NVDA on Windows. 

#### Follow up work
Tickets have already been for the following: 
- Adding breadcrumbs
- Handling invalid testReportIds in the URL and redirecting to a 404 page
- Adding the current test version info to the pages
- Adding links to the design page and examples to the example reports pages

#### QA Instructions
- Ensure you have some published test results
- Click on the "Test Reports" in the header or you can go there directly without signing in via by going to `/reports`.


rmeritz:
@jscholes & @sinabahram - Thanks again so much for taking the time to talk through our design ideas for the reports page. It would great to have your input to make sure we correctly reflected the keys aspects of your input in our design.  I've deployed this branch to staging and you can see the reports page [here](https://aria-at-staging.w3.org/reports).

Overall, Is there anywhere in the implementation that we missed the mark? Also, but less importantly I have a list of follow up questions below:

- In the summary table the visual design shows progress bars. There is an aria role progress bar, but when I tested in it read out "Progress Bar 90" for every cell. I just removed it now the cell just reads the value "90%". I think it makes more sense. Is that the correct approach?

- For the labels, of the table headers it more helpful to be succient like "Required" because everything is read over and over again or better to be clear like "Required Assertions" even Required Assertions (Passing/Total)"?

- For the "Assertions" columns, "9/10" get read as "9 slash 10" instead of "9 out of 10". Is that changeable without changing it visually? Is it appropriate to change?

- For skipped tests and skipped assertions, I still included the row, so you can see that is was skipped. I added a dash instead of in those cells so you see there is no result. NVDA isn't reading anything for though cells even though the `td` has an `aria-label` "No Results". Is it okay to leave in rows with all skipped values? Or does that make it too hard to navigate? Is there a better way to demarcate that? Would it work better if I just wrote "No results" in every cell for visual users as well, even though it is a bit visually overwhelming? I think we already discussed it would be problematic for navigation to make the `colspan` 3 for those rows and just right "no results" one time.

- Are we giving the appropriate level of detail in the captions? Sometimes NVDA is reading the caption of a table as multiple captions and I'm not sure why or what to do about it if anything.

- When moving from the test example pages to the detailed test results, we open up an anchor on a new page. I set JS `focus` but I'm not sure if there is a better way to work with anchors.

- I would like to style the table header with more than one style, I tried adding `span` or other elements inside the `th` but when I do that NVDA only reads part of the header when naming the column. Is there another way around that?

Thanks again for any input.

rmeritz:
@s3ththompson - I ended up not using slugs, and just ids on the reports page for test plans because we didn't have support yet for slugs. (Nothing saved in the DB and no guarantee of unique names.) I think in the long run for usability sharing links and SEO we should switch to slugs but I didn't think it was urgent to do in the next two weeks. If you think I should we could make a follow-up ticket for it.

rmeritz:
@isaacdurazo - We don't include the AT or Browser Version anywhere on the pages? Do you think it is appropriate to do so? And if so where? I was thinking of adding it everywhere, but it makes the table headers too long esp for SRs esp. on the summary page? I think we definitely could and should add it to the headers on the test plan reports page and it already on the old runs page.

jscholes:
@rmeritz Thanks for a quick turn-around on this!  Overall it's looking really promising.

I'll report back here with more detailed feedback, but in the meantime I have an immediate point of concern about the top-level summary table.

It was my understanding that in the initial designs, the highest level columns would only show the aggregate percentage of passing required tests across all AT/browser combinations, and that users would need to dig down to determine data for a specific combination.  But in this implementation, the summary table has columns for JAWS/Firefox and VoiceOver/Safari.

The screen reader speech is already quite verbose because of the wording of the column headers.  For example, "JAWS with Firefox Required Tests Passing" is 11 syllables before users hear the data.  If NVDA is the AT under test, that immediately adds an additional three syllables even though the length of the text is the same.

This is also problematic for braille users.  The JAWS/Firefox example would just about fit a 40-cell display if users opt for contracted output.  But:

* that doesn't guarantee that the data will be available without scrolling/panning;
* in uncontracted braille, users will likely need to scroll/pan to even read the header;
* the screen reader will add its own information to the output (such as cell coordinates); and
* many people use braille displays with less than 40 cells.

Much of the frustration for braille users will likely also be experienced by those relying on screen magnification.  This is quite difficult to read with two combinations.  The more that are added, the more cognitive overhead will be imposed.

Appreciate any thoughts here.  The mark-up is semantically accessible, e.g. the column headers are correctly associated with the cells underneath.  It's just impossible to scan quickly.

isaacdurazo:
@rmeritz re: versions, I agree with you, I think that requires a little more time from me to think about it. Can we leave it as is until I design a solution for this?

rmeritz:
@rmeritz re: versions, I agree with you, I think that requires a little more time from me to think about it. Can we leave it as is until I design a solution for this?

> We can definitely do it as part of a later design pass. I'll make a separate ticket of a design pass and note this.

rmeritz:
@jscholes - Thanks for highlighting your concerns about the top-level summary table. 

For now, I shortened the label for the AT/Browser Pairs to be "AT/Browser Passing Tests". Is this enough to address your problem? Would it be enough if I potentially shortened it further or differently? Just "AT/Browser" and description in the caption? "AT/Browser % Passing"?

I suspect that you might reply that the density of information is still too high or clarity too low. I have some other ideas to run by you below:

1. Replace the existing summary table with two new ones. Both would only two columns with the second column header being "Passing Required Tests". In one, the first table the first column header would be "Design Patterns" and in the other, the first column would be "At/Browser Pairs". Both tables would allow people to drill down more by clicking on the names in the first column. I drew out quick examples below:

    Design Patterns | Passing Required Tests
    ------| ----------------------
    All Design Patterns | 90%
    Checkbox | 100%
    Combobox | 80%

    AT/Browser | Passing Required Tests
    -----------|-----------------------
    All At/Browser Pairs | 90%
    NVDA/Chrome | 90%
    JAWS/Firefox | 90%

2. The above solution (1) feels like it addresses the needs of SR users, but it feels like there is a real loss for fully-sighted users not to be able to have access to the current version of the table that has more information density. Could we potentially just add the two new tables and keep the existing summary table in place primarily for sighted users? Or add a button under the first table "Expand for more AT/Browser info" that would turn the first table into the current summary table? Or a toggle switch to swap out these two summary table views for the existing one? Or is there another approach that you would recommend that would allow fully sighted users to still have a more dense view?

spectranaut (approved):
Approving the technical changes!

jscholes:
@rmeritz I'm Diving into these changes to provide more feedback, and want to start by addressing some of your questions.  Some of these have turned out quite long, so I'll add separate comments.  I'll then circle back to the summary table column header issue.

> In the summary table the visual design shows progress bars. There is an aria role progress bar, but when I tested in it read out "Progress Bar 90" for every cell. I just removed it now the cell just reads the value "90%". I think it makes more sense. Is that the correct approach?

In my opinion, yes.  The progress bar role is useful when a value will be updated during a running task because a screen reader can announce the changes automatically.  But here it would just add additional speech verbosity with no benefit.

jscholes:
> For the "Assertions" columns, "9/10" get read as "9 slash 10" instead of "9 out of 10". Is that changeable without changing it visually? Is it appropriate to change?

It's certainly possible via off-screen text.  That is, you can create a CSS class for screen-reader-only text, put an alternative representation of the data in a container with that class applied and then `aria-hidden="true"` the visible version.  For example:

```
<td>
  <span class="sr-only">18 of 18</span>
  <span aria-hidden="true">18 / 18</span>
</td>
```

I think this would be a nice usability touch.

Here are some resources related to off-screen text:

* From WebAIM: [CSS in Action, Invisible Content Just for Screen Reader Users](https://webaim.org/techniques/css/invisiblecontent/)
* [An explanation of the `screen-reader-text` CSS class used by WordPress](https://make.wordpress.org/accessibility/handbook/markup/the-css-class-screen-reader-text/)

jscholes:
> For skipped tests and skipped assertions, I still included the row, so you can see that is was skipped. I added a dash instead of in those cells so you see there is no result. NVDA isn't reading anything for though cells even though the td has an `aria-label` "No Results".

This is expected; screen readers don't have good support for accessible names applied to non-interactive containers.  For more context, you can take a look at [Woe-ARIA: The Surprisingly but Ridiculously Complicated World of aria-label/ledby](https://www.jantrid.net/2015/12/woe-aria-surprisingly-but-ridiculously.html).

> Is it okay to leave in rows with all skipped values? Or does that make it too hard to navigate? Is there a better way to demarcate that? Would it work better if I just wrote "No results" in every cell for visual users as well, even though it is a bit visually overwhelming?

I definitely think it's appropriate to surface the presence of no results, and just showing some text for all users would be accessible to some user groups.  But I don't want to ignore aspects of visual design here.  It does sound overwhelming to have the words "No results", "No data", whatever, visually repeated all over the place.  The problems I foresee though are:

* If we had off-screen text for screen reader users and left the dash visually, how would sighted users know that a dash represented "no results"?  It's sort of implied, I guess, but not super inclusive.
* On the other hand, if we had a legend to explain the meaning of the dash, it would be useless to most screen reader users because now they'd have text instead.  But it couldn't be `aria-hidden`, because screen reader users may need to explain the legend to somebody.
* If we do use a legend, where should it go?  Before each table where a dash might be used?  That also sounds overwhelming.

I think it would be ideal to find a solution that works for everyone here, rather than different treatment for different user groups.  But there are options depending on how we can make the visuals work.

jscholes:
> Are we giving the appropriate level of detail in the captions? Sometimes NVDA is reading the caption of a table as multiple captions and I'm not sure why or what to do about it if anything.

If anything, I think the captions are too long at present.  But to address your latter point first, NVDA isn't reading multiple captions.  It's just announcing a caption which spans across multiple lines in NVDA's virtual buffer, and announcing the fact that each line is part of a caption.

In terms of caption length: the text of the caption is mapped as the table's accessible name.  So when I move through the page by table, I hear the caption text, which at present is fairly long and not semantically prioritised.  For many users, the most convenient information will be inside the heading which precedes the table, but this isn't connected to the table in any way.

Looking at these in practical terms, for the top-level summary table, the caption reads:

> This table shows the percentage of required passing tests for all finalized test results for the most recently tested version of the ARIA-AT tests. The first row is a summary of results from all examples. All other rows are the summary of results for a single example.

The word "test" is used four times in the first sentence which is difficult to parse.  It also mentions the fact that this is a table, which is already visually apparent and announced by the screen reader anyway.  The latter two sentences only provide information which should be obvious from the data itself.

For an individual AT/browser result table:

> This table shows the number of passing required assertions, the number of passing optional assertions and the number of unexpected behaviors for results from the most recently tested version of the ARIA-AT tests. The first row is a summary of results from all tests run on Firefox with JAWS. All other rows are data from a single test on Firefox with JAWS. The test name in the first column is a link to the detailed results of a specific test.

Again, this only repeats information which is available elsewhere.  The table column/row headers should, and in fact do, adequately explain the meaning of the data.  The combination, which here is JAWS/Firefox, is available in the heading before the table.

Given the above:

* We need to decide which information we need right now.  There is currently no concept of versioned results available for browsing, for example, so do we need to indicate that the data represents the latest version of something?  This could be a future enhancement.
* Once we've determined which information we need from the captions, it would be great to extract it out into either the heading before a given table, or provide it as a brief note between the heading and table.
* Each table caption can then be removed, and replaced with an `aria-labelledby` attribute on the `<table>` elements referencing the ID of the associated heading.
* If a note is needed, it can be placed in a container (like a `<p>` or `<span>`) and then attached to the table via `aria-describedby`.  I don't think this is required for the individual results tables, and ideally not for the summary table either.

jscholes:
> When moving from the test example pages to the detailed test results, we open up an anchor on a new page. I set JS focus but I'm not sure if there is a better way to work with anchors.

Anchors should be avoided for setting focus.  They are usually blank, so a screen reader either announces a blank element or tries to guess an element close by to announce instead.  If you give an anchor an accessible name, it becomes a named control with no purpose and just adds noise.

Give the `<h1>` on the new page a `tabindex` of `"-1"` and an ID, and then focus that instead.  I should note that:

* this should happen for every page transition including summary -> example, not just example -> details; and
* when users navigate backwards, it would be ideal to place focus on the link representing the source page.  For example when moving from "Checkbox Example (Two State) Report" back to the main reports page, place focus on the "Checkbox Example (Two State)" link.

jscholes:
> I would like to style the table header with more than one style, I tried adding span or other elements inside the th but when I do that NVDA only reads part of the header when naming the column. Is there another way around that?

Could you describe what you're trying to achieve in more detail?  For example, what sort of styling are you trying to apply to which parts, and what sort of mark-up did you test?

jscholes:
@rmeritz 

> For now, I shortened the label for the AT/Browser Pairs to be "AT/Browser Passing Tests". Is this enough to address your problem?

This only changes the problem statement, because now the text is still quite long without indicating that only required assertions are considered in the percentage.  

> Replace the existing summary table with two new ones.

I think you've already highlighted the problems with this approach.  It feels confusing and would require a new page layout for when users click on a specific combination.

An approach I'm thinking through right now would be to change the table to only have two columns: example in the first, and "Required Passing Assertions" in the second.  Then, for each cell in the second column, include a list like:

* JAWS/Chrome: 100%
* JAWS/Firefox: 90%
* ... etc.

The issue with this is that it doesn't allow easy comparisons between a single AT/browser combination across rows.  For example, if I wanted to determine the support for two checkbox patterns with macOS VoiceOver and Safari, and they were at the bottom of these new lists, I would have to find the last item in the list each time.  Appreciate any thoughts here, although note that the list could be horizontal or vertical.

One thing about the current header for the first row.  It reads, "Design Pattern Examples", but:

* Should it be "examples", plural?  Other than the first one, each row only relates to a single pattern, and the text "All Design Pattern Examples" in the first row is pretty clear.
* Should we be using the word "example" at all?  On the test queue, project wiki, etc., these are always described as "test plans".

evmiguel (approved):

rmeritz:
@jscholes - Thanks a million for taking the time to provide some in-depth feedback and context. I'm working on incorporating it now.

rmeritz:
I've fixed a couple of small issues: the "No results" cells, the use of example in names, and the "19 of 20" issues. 

I've made follow up tickets for adjusting the captions, adjusting the anchors, breadcrumbs, design pass etc. 

I think we will open up an issue to continue to discuss the summary table format.

PR #235, opened 2020-12-10
Add AT Alert

spectranaut:
This PR does three things:
1. Adds an "ATAlert" component, which you can use to add a invisible alert for an AT. I tested in NVDA and firefox. In this case, you will here "You have been assigned to this test run" after you click "Assign yourself". Not sure if this is the best announcement? @isaacdurazo ?
2. Hides the "start testing" button instead of disabling it
3. Renames the "TestQueueRun" component within the file (it was named "TestQueueRow" which did not match the folder)

### DIFF
--- a/client/components/ATAlert/ATAlert.css
+++ b/client/components/ATAlert/ATAlert.css
@@ -1,0 +1,8 @@
+.at-alert {
+    clip: rect(1px, 1px, 1px, 1px);
+    height: 1px;
+    width: 1px;
+    overflow: hidden;
+    position: absolute;
+    white-space: nowrap;
+}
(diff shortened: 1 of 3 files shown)


evmiguel (changes requested):
The changes look semantically correct, but after trying it on VoiceOver and NVDA on Chrome, the screen reader reads "Continue testing" right away. I think the focus states are interfering with this alert.

Can you investigate this again for safe measure?

rmeritz:
@evmiguel - Can you possibly take this over because Val is off the project for the rest of the year?

evmiguel (approved):
Works on VoiceOver / Safari

PR #236, opened 2020-12-11
Add aria-current to test nav

spectranaut:
Followed instructions on this asana issue: https://app.asana.com/0/1193055321453706/1199542482316214

Tested with NVDA and Firefox, "current" was announced!

Also @isaacdurazo, in order to see which test you are in the test nav, I made the link bold (before you could tell where you were because it wasn't a link).

### DIFF
--- a/client/components/TestRun/TestRun.css
+++ b/client/components/TestRun/TestRun.css
@@ -219,3 +219,7 @@ main.container-fluid .test-iframe-contaner > .row {
     border-radius: 50%;
     display: inline-block;
 }
+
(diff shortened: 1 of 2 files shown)


evmiguel (approved):
Works well on NVDA/Edge and VoiceOver/Safari

jscholes:
@evmiguel One thing to note: Edge, although powered by Chromium, isn't guaranteed to behave exactly the same as Chrome itself, and isn't one of the browsers tested by the ARIA-AT project.  Where possible, it's good to test with actual Chrome.

PR #238, opened 2020-12-11
Test run page UI improvements

dawnwages:
Talked to @isaacdurazo about the changes and he's comfortable with taking the styling from here: small padding and background change.


isaacdurazo (approved):
This looks great, @dawnwages! I ended up pushing a commit to fix the layout issues after your changes because I realized it was a pretty simple fix

PR #249, opened 2020-12-17
heading/test-info-wrapper placement underneath H1 inside content object

dawnwages:
fixed the `heading`  / `test-info-wrapper` placement, but it's not an elegant solution.
better/future solution: restructure the order of rendering in `TestRun` so `heading` doesn't need to be a global variable.

https://app.asana.com/0/1193055321453706/1199152269152973/f

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -26,6 +26,7 @@ import checkForConflict from '../../utils/checkForConflict';
 import './TestRun.css';
 
 const PROGRESS_SAVED = 'Progress has been saved.';
+let heading = null;
 
 class TestRun extends Component {
     constructor(props) {
@@ -557,6 +558,7 @@ class TestRun extends Component {
                     <span className="task-label">Testing task:</span>{' '}
                     {`${this.state.currentTestIndex}.`} {test.name}
                 </h1>
+                <span>{ heading }</span>
                 <StatusBar key={nextId()} {...statusProps} />
                 <Row>
                     <Col md={9} className="test-iframe-contaner">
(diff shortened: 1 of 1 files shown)


isaacdurazo (approved):

isaacdurazo:
I tested it and it looks perfect, Thanks @dawnwages!

I'll let someone on the engineering side review the technical implementation

client/components/TestRun/index.jsx
```diff
@@ -389,7 +389,7 @@ class TestRun extends Component {
         );
     }
 
-    renderTest({ run, test, git_hash, at_key, testIndex, testerId }) {
+    renderTest({ run, test, git_hash, at_key, testIndex, testerId }, heading) {
```
evmiguel, line 392: @dawnwages I didn't like the idea of having a global heading variable while the component takes on lifecycle methods, so I just passed the heading as a param to renderTest(). It's not perfect, but it takes care of the global var code smell.

client/components/TestRun/index.jsx
```diff
@@ -557,6 +557,7 @@ class TestRun extends Component {
                     <span className="task-label">Testing task:</span>{' '}
                     {`${this.state.currentTestIndex}.`} {test.name}
                 </h1>
+                <span>{heading}</span>
```
evmiguel, line 560: Putting the header here made sense.

client/components/TestRun/index.jsx
```diff
@@ -671,18 +672,26 @@ class TestRun extends Component {
...
+                        userId,
+                        testerId: openAsUser || userId
+                    },
+                    heading
+                );
             } else {
-                content = <div>Tests are complete.</div>;
+                content = (
+                    <div>
+                        {heading}
```
evmiguel, line 691: We were losing the heading when the tests were completed, so I added it to this content section.

evmiguel (approved):

PR #250, opened 2020-12-17
Fix bug where admin is unable to assign testers

evmiguel:
Asana issue: https://app.asana.com/0/1193055321453706/1199634503793974

To test:
- Create a new user in the database
```
insert into users(fullname, username, email) values ('Test User', 'testuser', 'test@user.com') returning id;
insert into user_to_role(user_id, role_id) values (<id returned>,2);
insert into user_to_at(at_name_id, user_id, active) values (1,<id returned>,true);
insert into user_to_at(at_name_id, user_id, active) values (2,<id returned>,true);
insert into user_to_at(at_name_id, user_id, active) values (3,<id returned>,true);
```
- Go to the Test Queue, and verify that the entries in the Assign Users dropdown are not disabled
- Verify that you can assign the new user

### DIFF
--- a/client/components/TestQueueRun/index.jsx
+++ b/client/components/TestQueueRun/index.jsx
@@ -287,9 +287,7 @@ class TestQueueRun extends Component {
         return canAssignTesters;
     }
 
-    renderAssignMenu(admin) {
-        const { userId } = this.props;
-
+    renderAssignMenu() {
         const canAssignTesters = this.generateAssignableTesters();
 
         return (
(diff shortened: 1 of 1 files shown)


rmeritz:
@evmiguel - Doesn't your fix mean now that anyone can assign anyone? Shouldn't it only be admins can assign anyone and everyone else can only assign themselves? Or am I miss re-reading the code?

evmiguel:
@rmeritz [This line](https://github.com/w3c/aria-at-app/blob/d7bf8888e18bd284c7dac6bea01fa50f8e9bff75/client/components/TestQueueRun/index.jsx#L453) renders the assign menu for admins. The line that I removed was making the menu such that the admin could only assign themselves, while the rest of the users appeared as disabled. The tester functionality is still the same.

PR #251, opened 2020-12-17
Cosmetic change to Test Queue run states

evmiguel:
Asana task: https://app.asana.com/0/1193055321453706/1199634503793965

### DIFF
--- a/client/components/TestQueueRun/index.jsx
+++ b/client/components/TestQueueRun/index.jsx
@@ -93,8 +93,12 @@ class TestQueueRun extends Component {
     }
 
     updateRunStatus(status) {
+        let statusMap = {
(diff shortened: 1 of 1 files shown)


client/components/TestQueueRun/index.jsx
```diff
@@ -93,8 +93,12 @@ class TestQueueRun extends Component {
     }
 
     updateRunStatus(status) {
+        let statusMap = {
+            'In Review': 'draft',
+            Published: 'final'
```
rmeritz, line 98: Please change "Published Back to "Final" and then it should be good to go.

evmiguel: @rmeritz all set!

PR #252, opened 2020-12-18
Fix "In-Review" Reports

rmeritz:
- Ensure that not only "Final" but also "In Review" reports are viewable on all reports pages.
- Add back the updates status termst to the run view page.
- Fix "Report Issue" button for logged out user.


evmiguel (approved):
LGTM! And also, thanks for fixing the raise issue but with this PR

PR #253, opened 2020-12-18
Home page redesign

isaacdurazo:


s3ththompson:
@isaacdurazo this looks GREAT!

I updated the hero text to be a bit more beginner-friendly.

Perhaps we can avoid linking to the ARIA-AT project / APG in the intro paragraphs (given the usability feedback about users trying to find Call-to-Action links...) so perhaps we could just add APG to the "Get Involved" icons? Maybe it could follow "Join the community group" and then we could remove "Review results" to make space? What do you think?

s3ththompson:
@isaacdurazo oops i was trying to review and i think i actually unmarked as draft...

client/components/Home/index.jsx
```diff
@@ -1,76 +1,87 @@
...
+                        <Button variant="primary">Sign Up to Run Tests</Button>
+                        <Button variant="secondary">Browse Test Reports</Button>
+                    </div>
+                </section>
+                <section className="get-involved">
+                    <div className="container">
+                        <h2>Get Involved</h2>
+                        <div className="resources">
+                            <article>
+                                <img src={iconJoinCommunity} alt=""/>
```
s3ththompson, line 51: just a reminder to add alt text to all of the images...

client/components/App/App.css
```diff
@@ -352,6 +356,64 @@ img {
...
+.get-involved .resources article {
+    background: white;
+    padding: 2em 1em;
+    box-shadow: 0px 6px 11px 0px rgba(0, 0, 0, 0.15);
+    border-radius: 5px;
+    text-align: center;
+}
+
+.get-involved .resources img {
+    width: 100px;
```
s3ththompson, line 391: given that `.resources` has `grid` display based on percentages, when the screen size is narrow, the `article` elements are shrinking to less than 100px width...

this is making the images squeezed since sizing of image elements is a little weird. i think you could either fix by changing the height to be `auto` or the `object-fit` of the image to be `contain`

client/components/Home/index.jsx
isaacdurazo: I was thinking that since these images are abstractions and purely decorative, we could "hide" them for Screen Reders by having empty `alt` attributes

client/components/App/App.css
isaacdurazo: Good catch, Seth. I just fixed this issue

client/components/Home/index.jsx
s3ththompson: TIL, thank you for explaining!

s3ththompson (approved):

PR #263, opened 2020-12-23
Replace flashing text for saving with a success alert

evmiguel:
Asana issue: https://app.asana.com/0/1193055321453706/1199633858943639
Loom: https://www.loom.com/share/ffe5f07a04a14551840f36746efbf460

This is a naive solution to tackle the issue of notifying the user on save. This PR does the following:
- Set a save interval of 7.5 seconds
- Replace the plain text of "Progress has been saved" to an alert
- Removes the interval when component unmounts

Known issues:
- Screen reader users will hear the alert every 7.5 seconds. This could be annoying.
- On first load, the save alert appears the slowest. Subsequent tests alert fine.


rmeritz:
This seems really annoying. I think we are making it worse. I think at most we want to do it once per page. 

I don't really want to merge it as is because I thinks its probably better not to know than to know every 7.5 secs.

evmiguel:
> This seems really annoying. I think we are making it worse. I think at most we want to do it once per page.
> 
> I don't really want to merge it as is because I thinks its probably better not to know than to know every 7.5 secs.

I agree. @s3ththompson , what should be done?

alflennik:
This PR will conflict with the new test renderer, so I think it's time to close this one! As for the problem the PR addresses, we will have to reevaluate whether saving is causing problems when we start testing the UX again in the next few weeks.

PR #264, opened 2020-12-23
Change page title in test run when the test is in a review state

evmiguel:
Asana issue: https://app.asana.com/0/1193055321453706/1199542482316236
Loom: https://www.loom.com/share/39a25fda860b42538a66f8a013796778


rmeritz:
@evmiguel - The title now always says reviewing tests even when you go to test that doesn't start as reviewing tests.

evmiguel:
> @evmiguel - The title now always says reviewing tests even when you go to test that doesn't start as reviewing tests.

It changes for me. What are the conditions that you were testing with? Such as, "Open run as" or something else?

rmeritz:
Hmm. I cannot recreate either...

PR #265, opened 2020-12-24
reorder nav and change Configure Active Run to Test Configuration

s3ththompson:
This is a purely cosmetic change. I reordered the buttons in the nav, renamed "Configure Active Run" to "Test Configuration," and updated the home page to reflect the new changes.

FWIW The git diff for the nav bar looks weird, but I reordered each item as an entire block (e.g. `<Nav.Link></Nav.Link>` or `{isAdmin && (<Nav.Link></Nav.Link>)}`.


s3ththompson:
@rmeritz I still don't have the full app up-and-running locally, so you might want to check this out locally just to make sure it looks okay. I couldn't preview it.

PR #267, opened 2021-01-08
Swagger file and readme

evmiguel:
I added a folder,`server/swagger`, which houses the Swagger API file. I added documentation on how to run Swagger editor and UI in a markdown file.

PR #268, opened 2021-01-12
Swagger run endpoints

evmiguel:
This is dependent on #267.


alflennik:
Closing since we are firmly out of the RESTful land into GraphQL land now!

PR #269, opened 2021-02-23
Fix aria-describedby / aria-labelledby values on Reports page

s3ththompson:
On Reports page, `aria-describedby` and `aria-labelledby` values referred to `#div-id` rather than just `div-id`.


jesdaigle (approved):

jesdaigle:
Is there a reason this wasn't merged @s3ththompson ?

PR #270, opened 2021-03-01
Fix typo on index

alflennik:
I noticed the start page has typo of the word assertion on it. Clicking around the app, I think deleting that misspelled word is a good solution.

Take a look @isaacdurazo ?


isaacdurazo (approved):
🌟

PR #273, opened 2021-03-19
Add .nvmrc file

alflennik:
Since this project recommends using Node 12 with [Node Version Manager](https://github.com/nvm-sh/nvm), I think it would be helpful to follow the common practice of including an `.nvmrc` file.

An `.nvmrc` file allows for nvm users to [automatically switch to the correct Node version](https://github.com/nvm-sh/nvm#deeper-shell-integration) when `cd`ing into the project directory. And it could also allow other automated tools to understand which Node version is in use.

Without it, developers may have to manually switch the nvm Node version every time they open a new terminal window.


jscholes:
@alflennik Given that Node 12.x will be end-of-life at the end of April next year, are there plans to update to a newer LTS release?  I think the `.nvmrc` file is a good idea, although I don't believe it will work with the Windows version of NVM so standardising on 14.x would help with project interop (for me at least).

alflennik:
@jscholes I think updating Node to 14 would be great. To do this I will have to investigate what is keeping us on the old version, and figure out how to upgrade Node in the staging and production environments.

It might be nice to set up this `.nvmrc` file as the source of truth for all environments about what Node version to use. But I have to admit that this might end up more trouble than it's worth.

alflennik:
Closed since I think the solution here to remove the dependency on Node 12. I know from more time with the app that we can easily support Node 14 and higher just by updating the npm packages.

PR #283, opened 2021-04-29
Add quick role switching to development environments

alflennik:
Currently changing roles requires leaving and joining GitHub teams which makes testing difficult, especially since rejoining the team requires a team admin to re-add you, requiring multiple people to cooperate.

This change will allow developers to quickly switch roles by adding a parameter to the URL bar when they log in, which will persist during the session.

It also has the benefit that other developers will be able to access the logged-in parts of the app without needing to create their own GitHub teams specifically for that purpose.

I tested these scenarios:

1. Works as tester when I am normally an admin
2. Works as admin when I am normally a tester
3. Works as tester or admin when I normally have no access
4. Reverts to normal after signing out
5. Requires a environment variable flag to enable the functionality


jesdaigle (approved):
This is great. I think we should merge it into the develop branch before it goes to main tho. :)

PR #289, opened 2021-05-12
Implementation of new data model definitions

howard-e:
Following the drafting up of a new database schema, this aims to provide new Sequelize models to support that migration. 

Note that this introduces breaking changes to how models are referenced on the server, queries and existing import scripts.

~~NB. This was branched off of `entity-tests` so certain files may be duplicated here as well as tasks that may have come out of the discussion for that related PR #287.~~

If your database has not been initialized, run:
```bash
    yarn db-init:dev;
else
fi;

yarn sequelize db:migrate;
```

Otherwise, you should be able to safely migrate up with:
```bash
yarn sequelize db:migrate
```


patches/sequelize-test-helpers+1.3.3.patch
```diff
@@ -0,0 +1,314 @@
...
+--- /dev/null
++++ b/node_modules/sequelize-test-helpers/src/utils/checkIsJestRunner.js
+@@ -0,0 +1,7 @@
++const isJestRunner = process.env.SEQUELIZE_TEST_HELPERS_JEST_RUNNER === 'true' || process.env.SEQUELIZE_TEST_HELPERS_JEST_RUNNER === true
++
++module.exports = {
++  isJestRunner,
++  expect: isJestRunner ? expect : require('chai').expect,
++  sinon: isJestRunner ? () => {} : require('sinon')
++}
```
alflennik, line 314: Why is this file here after merging the entity-tests branch??? I saw a couple other inconsistencies as well. Is that possible to fix?

howard-e: The base branch wasn't merged in after approval the approval for #287. Fixed.

jesdaigle (approved):

PR #292, opened 2021-05-18
Support deploying to "sandbox" environment

jugglinmike:
Re-order the Roles in the Ansible Playbook to reflect their
interdependency. Ensure that privilege escalation is used for Tasks
which require it. Introduce additional Tasks to support using
password-based "sudo" when the environment requires it.

---

I haven't been able to verify this patch because the `develop` branch cannot be deployed at the moment.


deploy/inventory/sandbox.yml
```diff
@@ -0,0 +1,29 @@
+---
+# The difference between this environment and aria-at-staging.w3.org is that
+# this environment is Bocoup-internal, and used for testing and sharing the
+# more drastic or long-term changes we're making.
+#
+# - The sandbox environment will track the dev branch, while staging and
```
alflennik, line 6: Good idea to add this documentation!

Maybe change to "develop" branch since that's the literal branch name?

deploy/roles/application/tasks/upload-source-code.yml
```diff
@@ -0,0 +1,53 @@
...
+      become: yes
+  always:
+    - name: Revoke temporary privilege
+      lineinfile:
+        path: /etc/sudoers
+        state: absent
+        line: '%sudo ALL=(ALL) NOPASSWD:/usr/bin/rsync'
+        validate: 'visudo -cf %s'
+      become: yes
+  when: deployment_mode != 'development'
```
alflennik, line 53: I really like the way you solved it! And you explained it really clearly.

alflennik (approved):
I'm really excited to have a sandbox environment we can actually use. It's also cool to see how you solved the messy problems which were taking me so long to figure out.

I had one extremely minor suggestion which is optional, but yeah, I'm happily approving this.

deploy/inventory/sandbox.yml
jugglinmike: Sure. I copied this text from [Bocoup's internal design documentation](https://docs.google.com/document/d/1DEDMgrKim7ABg7gIbZ9ZGc6ITkVU_bXnJS6KuzmaVVw/edit), so you may want to update that as well.

alflennik: Good idea - I updated it there too.

alflennik:
@jugglinmike I meant to ask, we have an outstanding issue about updating the Node version to 14. Do you think we could include that?

jugglinmike:
Since that's an orthogonal issue, it would be best to address it with a dedicated patch. We should probably start by making a task for that in our issue tracker.

alflennik:
Additions to the PR:
- Updates Node to 14
- Fixes some inaccuracies in the docs
- Fixes bugs in the new environment

jesdaigle (approved):

PR #293, opened 2021-05-18
GraphQL Resolvers

alflennik:
- Creates a file structure for resolvers.
- Connects the existing auth system to GraphQL.
- Adds the ability to write GraphQL tests with TDD.
- Should match up well with Howard's model service.


server/tests/util/graphql-test-utilities.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Returns the data from a given query, useful for situations where errors are
+ * not expected and no query variables are needed.
+ * @param {String} gql - GraphQL query from a gql template string
+ * @example gql`query { me { username } }`
+ * @param {Object} options
+ * @property {Object} user - Replace the default user or set it to null
+ * @returns {Any} Data matching the query
+ */
```
alflennik, line 40: @howard-e Any tips about improving the JSDoc?

server/graphql-context.js
```diff
@@ -0,0 +1,6 @@
+const getGraphQLContext = ({ req }) => {
+    const user = req.session?.user ?? null;
+    return { user };
+};
```
alflennik, line 4: Assumes Mike will add Node 14 support. If he doesn't I'll update it to use the Node 12 syntax here.

server/tests/util/graphql-test-utilities.js
howard-e: Sure!

1. So you typically use the `@property` directive when defining a `@typedef` or a namespace. So you'd use the `@property` directive to declare a property of that defined `@typedef`. I don't think that's what you want here with `user` so I think you could safely change that to using `@param` or create a `@typedef` for the `options` object and make it a property of that.
2. If you do choose to just make it a `@param` however, make sure you use `options.user` so the IDE could identify `user` as actually being a part of `options`
3. This is a nitpick, but usually, the pattern you'd see when representing primitives vs Classes/Types/Objects types is with the latter being your typical PascalCase (if that's how they're defined) and the former being lowercase which is what I think all of these types could be (primitives). All this isn't relevant of course if the project has already been using their own type standard when defining the JSDocs but could look into a best practice around it

Otherwise, it's looking good and descriptive enough. And I'm always a big fan of the `@example` being used so that's great!

alflennik: Ah I'm glad I asked, thanks for the help.

Here's what I ended up with:

```js
/**
 * Returns the data from a given query, useful for situations where errors are
 * not expected and no query variables are needed.
 * @param {GraphQLSyntaxTree} gql - GraphQL query from a gql template string
 * @example gql`query { me { username } }`
 * @param {object=} options
 * @param {object=} options.user - Replace the default user or set it to null
 * to simulate being logged out.
 * @returns {any} Data matching the query
 */
```

howard-e: Nice. Looks good!

server/graphql-context.js
alflennik: He wants to do it in a separate PR, not the upcoming one. So I reverted the syntax.

howard-e (approved):

jesdaigle (approved):

PR #290, opened 2021-05-12
Introduce GraphQL

alflennik:
This PR adds GraphQL to the project!

Since the business logic and Sequelize are both under construction, I put GraphQL into mock mode, so no resolvers are needed and all the data it returns is faked. This makes it possible to write queries and verify that the schema supports them.

Below are the queries I tested, which are examples of what using GraphQL from the client side would be like. Please don't be shy about giving feedback around those queries, which I think are the most important way to judge the schema. Do they seem clear? Are they confusing or tangled in some way? Can you think of any use-cases I didn't include?

My hope is that this schema will support 90% of what the current app can do.

```gql
# Test Reports
# https://aria-at.w3.org/reports
query {
  testPlanReports {
    id
    coveragePercent
    testPlan {
      id
      title
    }
    testPlanTarget {
      id
      title
    }
  }
}
```

```gql
# Test Report > Test Plan
# https://aria-at.w3.org/reports/test-plans/4
query {
  testPlanReports(testPlan: 123) {
    testPlanTarget {
      title
    }
    canonicalRun {
      testResults {
        title
        passedAssertions
        totalAssertions
      }
    }
  }
}
```

```gql
# Test Report > Results
# https://aria-at.w3.org/results/run/4
query {
  testPlanReport(id: 123) {
    testPlan {
      title
    }
    testPlanTarget {
      title
    }
    canonicalRun {
      testResults {
        title
        passedAssertions
        totalAssertions
        # More here
      }
    }
  }
}
```

```gql
# Test Queue
# https://aria-at-staging.w3.org/test-queue
query {
  testPlanReports {
    id
    publishStatus
    testPlan {
      id
      title
      sourceGitCommit
      testCount
    }
    testPlanTarget {
      id
      title
    }
    testPlanRuns {
      id
      tester {
        username
      }
      testResultCount
    }
  }
}
```

```gql
# Test Queue Mutations
mutation AssignTester {
  testPlanReport(id: 123) {
    assignTester(user: 445)
    resultingTestPlanReport {
      testPlanRuns {
        tester {
          username
        }
      }
    }
  }
}

mutation RemoveReportTester {
  testPlanReport(id: 123) {
    deleteTestPlanRun(user: 445)
    resultingTestPlanReport {
      testPlanRuns {
        tester {
          username
        }
      }
    }
  }
}

mutation UpdateReportStatus {
  testPlanReport(id: 123) {
    updateStatus(status: FINALIZED)
    resultingTestPlanReport {
      coveragePercent
    }
[...]

PR #294, opened 2021-05-19
Model service layer for new database schema

howard-e:
This provides a standard to be set up on and followed when creating general CRUD handlers and custom functions for operations related to manipulating and retrieving data from the database instance.

#### Files to note:
1. `/server/models/services/ModelService.js`
    - Provides a base module for the other model handlers to utilize
    - Provides typical CRUD + Sequelize specific operations
      - getById
      - getByQuery
      - get (pagination + sort + filter supported; custom search implicitly through the custom handling at the specific Model's Layer)
      - create
      - update
      - removeById
      - removeByQuery
      - rawQuery
2. `/server/models/services/helpers.js`
    - Used to pull the attributes for the models which is useful when conditionally requesting attributes for a model and/or it's associations in a request

Ideally, the functionality here also covers the requirements of the queries and mutations dictated by #290 and #293.

#### Example Usage:
**NB. Please see `/server/models/services/UserService.js` for more context of what's happening at the UserService level**

**Retrieving All Users**
```javascript
const UserService = require('path/to/models/services/UserService');
const ModelServiceHelper = require('path/to/models/services/helpers');

// retrieving all users
const result = await UserService.getUsers(
  'search condition', // query that would be handled at the UserService level to then be passed to ModelService's where clause
  { }, // implied filter object, eg. finding all records with status = active is { status: 'active' }
  ['username'], // specify exactly which attributes to be returned for the user object
  ModelServiceHelper.ROLE_ATTRIBUTES, // get all attributes for associated role
  [], // don't return the test plan run association
  { limit: 3, page: 1, order: [ ['username', 'desc'] ], enable: true } // pagination support; if enable is set to false, the result is just what the data attribute containing all results that matches the query
);
```

```
// expected result
{
    "page": 1, // page requested
    "pageSize": 3, // limit
    "resultsCount": 3, // results counted for the current page
    "totalResultsCount": 5, // how many results in all would match this query
    "pagesCount": 2, // how much 'pages' of results exist
    "data": [
        {
            // "id": 2, - not included; not a part of the user attributes
            "username": "foobar-tester",
            // "createdAt": "2021-05-14T18:57:16.232Z", - not included; not a part of the user attributes
            // "updatedAt": "2021-05-14T18:57:20.473Z", - not included; not a part of the user attributes
            "roles": [
                {
                    "name": "tester" // ModelServiceHelper.ROLE_ATTRIBUTES only found 'name'
                }
            ]
            // "testPlanRuns": [ { ... } ] // not included
        },
        ...
    ]
}
```

**Retrieving a User by id**
```javascript
const UserService = require('path/to/models/services/UserService');
const ModelServiceHelper = require('path/to/models/services/helpers');

// retrieving user
const result = await UserService.getUserById(
  1, // id of the user we're looking for
  ['username'], // specify exactly which attributes to be returned for the user object
  // ModelServiceHelper.ROLE_ATTRIBUTES, // if attributes dataset is omitted, all attributes for that model/association will be returned
  // ModelServiceHelper.TEST_PLAN_RUN_ATTRIBUTES // if attributes dataset is omitted, all attributes for that model/association will be returned
);
```

```
// expected result
{
    "username": "foobar-admin",
    "roles": [
        {
            "name": "admin"
        },
        ...
    ],
    "testPlanRuns": [
        {
            "id": 1,
            "isManuallyTested": true,
            "tester": 1,
            "testPlanReport": 1
        },
        ...
    ]
}
```

The other methods (update, remove, create, etc) follow the same pattern closely enough in terms of parameters to be passed.


server/models/services/TestPlanReportService.js
```diff
@@ -0,0 +1,491 @@
...
+                testPlanRunAttributes,
+                userAttributes,
+                testResultAttributes
+            ),
+            testPlanAssociation(testPlanAttributes),
+            testPlanTargetAssociation(testPlanTargetAttributes)
+        ]
+    );
+
+    if (canonicalRunKeyFound) {
```
howard-e, line 127: Would definitely like some special focus on the handling of `canonicalRun` @alflennik and what it could mean for the format coming from the `TestResult.data`. Haven't fully locked down how it's handled, whether at this layer (which is ok too) or as some virtual property of the TestPlanReport model so was considering to remove support for it until we've confirmed the `TestResult.data`. Thoughts?

server/models/services/ModelService.js
```diff
@@ -0,0 +1,216 @@
...
+ * @param {Model} model - Sequelize Model instance to query for
+ * @param {object} filter - values to be used to search Sequelize Model
+ * @param {any[]} attributes - attributes of the Sequelize Model to be returned
+ * @param {any[]} include - information on Sequelize Model relationships
+ * @param {object} pagination - pagination options for query
+ * @param {number} [pagination.page=0] - page to be queried in the pagination result (affected by {@param pagination.enable})
+ * @param {number} [pagination.limit=10] - amount of results to be returned per page (affected by {@param pagination.enable})
+ * @param {string[][]} [pagination.order=[]] - expects a Sequelize structured input dataset for sorting the Sequelize Model results (NOT affected by {@param pagination.enable}). See {@link https://sequelize.org/v5/manual/querying.html#ordering} and {@example [ [ 'username', 'DESC' ], [..., ...], ... ]}
+ * @param {boolean} [pagination.enable=false] - use to enable pagination for a query result as well useful values. Data for all items matching query if not enabled
```
alflennik, line 78: I think this enabled property isn't needed - not including the pagination object is enough to know there won't be any pagination!

server/models/services/ModelService.js
```diff
@@ -0,0 +1,216 @@
...
+
+    // search and filtering options
+    let where = { ...filter };
+
+    // pagination and sorting options
+    let { page = 0, limit = 10, order = [], enable = false } = pagination; // page 0->1, 1->2; manage through middleware
+    // 'order' structure eg. [ [ 'username', 'DESC' ], [..., ...], ... ]
+    if (page < 0) page = 0;
+    if (limit < 0 || !enable) limit = null;
+    const offset = limit < 0 ? 0 : page * limit; // skip (1 * 10 results) = 10 to get get to page 2
```
alflennik, line 98: I don't think we need to check for negative numbers - if someone passes in a negative number and it explodes, it seems ok to me. That might make the code here slightly cleaner too! If you do think negative numbers are possible I think throwing an error would be appropriate.

server/models/services/ModelService.js
```diff
@@ -0,0 +1,216 @@
...
+ * @param {any[]} include - information on Sequelize Model relationships
+ * @returns {Promise<Model>} - Sequelize Model
+ */
+const getById = async (model, id, attributes = [], include = []) => {
+    if (!model) throw new Error('Model not defined'); // TODO: pass through custom APIError
+
+    // findByPk
+    return await model.findOne({
+        where: { id },
+        attributes,
```
alflennik, line 41: I actually don't think we need the ability to request specific attributes! GraphQL handles removing unused fields internally so it wouldn't be convenient (it would be difficult) to pull the request attributes out a GraphQL request.

When I say this, I'm not talking about associations, I'm specifically referring to all the attributes which are part of the model's table. I think this service should load all of them, every time, no need to provide a specific list.

server/models/services/BrowserService.js
```diff
@@ -0,0 +1,261 @@
+const ModelService = require('./ModelService');
+const { BROWSER_ATTRIBUTES, BROWSER_VERSION_ATTRIBUTES } = require('./helpers');
+const { Sequelize, Browser, BrowserVersion } = require('../');
+const { Op } = Sequelize;
+
+// Section :- association helpers to be included with Models' results
```
alflennik, line 6: What are these! `Section :-`? I don't think I've seen them before.

server/models/services/helpers.js
```diff
@@ -0,0 +1,42 @@
...
+    BROWSER_VERSION_ATTRIBUTES: getSequelizeModelAttributes(BrowserVersion),
+    ROLE_ATTRIBUTES: getSequelizeModelAttributes(Role),
+    TEST_PLAN_ATTRIBUTES: getSequelizeModelAttributes(TestPlan),
+    TEST_PLAN_REPORT_ATTRIBUTES: getSequelizeModelAttributes(TestPlanReport),
+    TEST_PLAN_RUN_ATTRIBUTES: getSequelizeModelAttributes(TestPlanRun),
+    TEST_PLAN_TARGET_ATTRIBUTES: getSequelizeModelAttributes(TestPlanTarget),
+    TEST_RESULT_ATTRIBUTES: getSequelizeModelAttributes(TestResult),
+    USER_ATTRIBUTES: getSequelizeModelAttributes(User),
+    USER_ROLES_ATTRIBUTES: getSequelizeModelAttributes(UserRoles)
+};
```
alflennik, line 42: Based on what I said elsewhere about the attribute filtering not being needed, it might be possible to remove this file!

server/models/services/AtService.js
```diff
@@ -0,0 +1,404 @@
...
+ */
+const atModeAssociation = atModeAttributes => ({
+    association: 'modes',
+    attributes: atModeAttributes
+});
+
+// At
+
+/**
+ * NB. You can pass any of the attribute arrays as '[]' to exclude that related association
```
alflennik, line 42: What's `NB.`?

alflennik (commented):
I'm definitely keeping the same enthusiasm I had back when I got the first glimpse of this last week. Having the ability to wrap any of the models and then to add features is a really good architecture and is the kind of thing that can pay dividends for a long time. I imagine adding a feature like soft delete to all our models like this. Or adding functionality for a cascading delete where we delete models which are related to each other. Pagination and search is just the beginning. So I really think it's super cool. And I'm excited to use it 🤠

The biggest piece of feedback is that there's one thing it does that we might not actually need - accepting a list of attributes to query for. I explained a bit more in a comment but we can talk 1-on-1 about it as well if that helps. This could be good because I think the attributes are a source of complexity, so removing that feature might even clean up the code.

The other idea I have, and this is mainly caused by my really liking this is - do you think in the future we can hide the Sequelize models and only make these services available? Basically forcing the GraphQL world to use these services. I really like this architecture and I want to remove any chance of undermining it by reaching down to the lower level of Sequelize which it's abstracting.

This is awesome, I really like it. And I'm looking forward to using it!

server/models/services/TestPlanReportService.js
alflennik: @howard-e It can be any of the runs, maybe the first one, or a random one. When the report is finalized that means all conflicts between the results have been resolved, so they're all identical.

alflennik: There's no need for an additional key.

server/models/services/ModelService.js
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * @link {https://sequelize.org/v5/manual/models-usage.html#-code-findandcountall--code----search-for-multiple-elements-in-the-database--returns-both-data-and-total-count}
+ * @param {Model} model - Sequelize Model instance to query for
+ * @param {object} filter - values to be used to search Sequelize Model
+ * @param {any[]} attributes - attributes of the Sequelize Model to be returned
+ * @param {any[]} include - information on Sequelize Model relationships
+ * @param {object} pagination - pagination options for query
+ * @param {number} [pagination.page=0] - page to be queried in the pagination result (affected by {@param pagination.enable})
+ * @param {number} [pagination.limit=10] - amount of results to be returned per page (affected by {@param pagination.enable})
+ * @param {string[][]} [pagination.order=[]] - expects a Sequelize structured input dataset for sorting the Sequelize Model results (NOT affected by {@param pagination.enable}). See {@link https://sequelize.org/v5/manual/querying.html#ordering} and {@example [ [ 'username', 'DESC' ], [..., ...], ... ]}
```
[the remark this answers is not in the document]
howard-e: Yep, multiline @param directive is supported. We'll just have to scroll down a little bit further ha. Although you believe that warrants a global change, there are prettier configuration options you can look into for JSDoc

server/models/services/ModelService.js
```diff
@@ -0,0 +1,216 @@
...
+ * @param {boolean} [pagination.enable=false] - use to enable pagination for a query result as well useful values. Data for all items matching query if not enabled
+ * @returns {Promise<*>} - collection of queried Sequelize Models or paginated structure if pagination flag is enabled
+ */
+const get = async (
+    model,
+    filter = {}, // pass to 'where' for top level model
```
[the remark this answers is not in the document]
howard-e: Correct. At this level for the `get` method, that parameter truly is just the `where` object so I could make that change.

For the others, they aren't exactly the entire `where` object. Handled just so slightly different but the distinction between `queryParams` and `updateParams` for with the `update` method for example is easier to understand at a glance I believe vs `where` and `updateParams`.

server/models/services/ModelService.js
howard-e: But it shouldn't explode! If it were to blow up, that's a potentially bad experience for the user and I support it because I'd rather default the calling method to the correct approach but throwing an error is ok as well.

server/models/services/BrowserService.js
howard-e: Those prefixes should have been removed. Missed them 🙂 - but depending on your IDE's configuration, you can use it to quickly navigate around a file so I use it sometimes as a file grows. It's basically bookmarks most IDEs support.

server/models/services/AtService.js
howard-e: Note briefly

server/models/services/TestPlanReportService.js
howard-e: That means this can just be easily handled in the resolver then? To just move up any object found in the linked `testPlanRuns` array as the `canonicalRun`

server/models/services/ModelService.js
alflennik: I get this - some tolerance to failure at such an important and frontend-facing feature makes sense.

alflennik:
> The biggest piece of feedback is that there's one thing it does that we might not actually need - accepting a list of attributes to query for.

Howard and I talked this one over and he explained that it the fact that GraphQL may not need this feature doesn't mean it's not going to be handy, so we decided to keep it in!

alflennik (approved):
Changes look awesome! Thanks!

jesdaigle (approved):

PR #295, opened 2021-05-24
Server Tests Refactoring and Cleanup

howard-e:
To align with the migration efforts, this aims to provide new tests for the server application for functionality dependent on the new database schema.
Outdated tests reliant on the previous structure have also been removed and the ones that have remained may slowly be refactored over time.

Generally, basic CRUD functionality and some business logic operations are being tested for the various Model Service Layers currently implemented.


server/models/services/TestPlanRunService.js
```diff
@@ -114,6 +114,20 @@ const createTestPlanRun = async (
...
+        {
+            tester,
+            testPlanReport
+        },
+        testPlanRunAttributes,
+        testPlanReportAttributes,
+        testResultAttributes,
+        userAttributes
+    );
+    if (existingTestPlanRuns.length) return existingTestPlanRuns[0];
```
alflennik, line 129: We'll need to implement this rule (no duplicate test plan runs) at the GraphQL level, so probably just throwing an error is better at this level. I'm a fan of throwing errors 😈

server/models/services/TestPlanTargetService.js
```diff
@@ -61,25 +63,47 @@ const createTestPlanTarget = async (
...
+        browser = browserResults.length ? browserResults[0].id : 0;
+    } else {
+        // assuming type is number; get name to construct title if required
+        const browserResult = await BrowserService.getBrowserById(browser);
+        browserName = browserResult ? browserResult.name : '';
+    }
+
+    // Construct title as <at> <atVersion> with <browser> <browserVersion>
+    if (!title)
+        title = `${atName} ${atVersion} with ${browserName} ${browserVersion}`;
```
alflennik, line 93: I actually think this attribute makes more sense to be in GraphQL only. However, since I'm about to migrate the data, I think it's easier to leave it in for now and let me do it in a day or two!

Let me know if you agree with that?

server/models/services/TestPlanTargetService.js
```diff
@@ -61,25 +63,47 @@ const createTestPlanTarget = async (
...
+        browserName = browser;
+        const browserResults = await BrowserService.getBrowsers(null, {
+            name: browser
+        });
+        browser = browserResults.length ? browserResults[0].id : 0;
+    } else {
+        // assuming type is number; get name to construct title if required
+        const browserResult = await BrowserService.getBrowserById(browser);
+        browserName = browserResult ? browserResult.name : '';
+    }
```
alflennik, line 89: Thanks for thinking this through!! Looks like it will work great.

server/scripts/import-tests/populate-test-data.js
```diff
@@ -0,0 +1,19 @@
...
+const populateTestDatabase = async () => {
+    const testDataScript = fs.readFileSync(
+        __dirname + '/pg_dump_2021_05_test_data.sql',
+        'utf-8'
+    );
+    console.info('test.data.dump', testDataScript);
+    return await db.sequelize.query(testDataScript);
+};
+
+if (require.main === module)
```
alflennik, line 14: Could you assign `require.main === module` to a variable like `isCliEnvironment` in case someone (me 🤪) doesn't know what it means?

server/scripts/import-tests/populate-test-data.js
```diff
@@ -0,0 +1,19 @@
...
+    populateTestDatabase()
+        .then(() => console.info('Successfully Populated'))
+        .finally(() => console.info('Test Data Populated'));
```
alflennik, line 17: I don't think we need this `.finally` (?) And the fact it will say "test data populated" when errors occurred seems like it might be confusing!

server/scripts/import-tests/pg_dump_2021_05_test_data.sql
```diff
@@ -0,0 +1,145 @@
+--
```
alflennik, line 1: What is this file doing? Is it a backup? Do you think it should be committed or not?

server/scripts/import-tests/index.js
```diff
@@ -241,8 +241,8 @@ const ariaAtImport = {
 
         // checking to see if unique testPlan row
         const testPlanResult = await client.query(
-            'SELECT id, "sourceGitCommitHash" FROM "TestPlan" WHERE "sourceGitCommitHash"=$1 and "exampleUrl"=$2',
-            [commitHash, exampleUrl]
+            'SELECT id, "sourceGitCommitHash" FROM "TestPlan" WHERE "sourceGitCommitHash"=$1 and parsed ->> \'directory\'=$2',
+            [commitHash, exampleDir]
```
alflennik, line 245: I'm assuming this is the bug fix you mentioned? This definitely deserves a comment!

server/tests/models/services/AtService.test.js
```diff
@@ -0,0 +1,591 @@
+const { sequelize } = require('../../../models');
+const AtService = require('../../../models/services/AtService');
+const randomStringGenerator = require('../../util/random-character-generator');
+const { dbCleaner } = require('../../util/db-cleaner');
+
+afterAll(async () => {
+    await sequelize.close(); // close connection to database
```
alflennik, line 7: Why do we need to explicitly close the database? Maybe you could update the comments to explain why?

alflennik (commented):
That green checkmark looks pretty awesome!!

I enjoy reading the tests - they're quite readable even when there's a lot of them. I've found tests often to be hard to read but that's definitely not the case here.

Comments are all peripheral stuff but for the main meat of the PR I have basically no feedback besides "looks great."

server/scripts/import-tests/pg_dump_2021_05_test_data.sql
howard-e: It's what is being used to populate the test database! See `scripts/import-tests/populate-test-data.js`

server/tests/models/services/AtService.test.js
howard-e: It's generally a safe practice to explicitly close your db connections if no longer needed especially in a testing environment. It will free up space and lessen the chance of any transactional locks.
Also, given that this is a Postgres instance, there's a ~100 max connections pool size so you wouldn't want to unintentionally hit that (although it should never happen) while the build server is attempting to run many test suites that depend on those connections back-to-back

server/scripts/import-tests/populate-test-data.js
howard-e: Yea, could be done. The plan is to move all these environment config level settings through a single module (`process.env` variables, etc) so that's coming

server/models/services/TestPlanRunService.js
howard-e: Yep, will definitely lock that into a future migration!

server/tests/models/services/AtService.test.js
alflennik: Gotcha - maybe remove the comments then!

server/scripts/import-tests/populate-test-data.js
alflennik: Ok, no need to comment if it's temporary code.

server/models/services/TestPlanTargetService.js
howard-e: So I think it's better managed here just in case there is an instance where a custom title is required to be passed or having the title manipulated in some other way (which this supports) and having that persisted. so I don't think that should necessarily be dependent on the GraphQL query. I could see instances where definitely having a saved `title` would be ideal and also cases against but I'm definitely up for discussing it further

howard-e: Having a "defaulted" title persisted on the model and the GraphQL having it's own variation based on something specific for its query works as well.

server/scripts/import-tests/populate-test-data.js
howard-e: Good point!

server/models/services/TestPlanTargetService.js
alflennik: Sure, let's keep it then!

jesdaigle:
Nice work @howard-e  :)

jesdaigle (approved):

alflennik (approved):

PR #297, opened 2021-05-25
Test queue migrations

alflennik:
This is one where a speedy review will help! These 600 lines changed could cause 601 merge conflicts if we don't merge it speedily.

This PR synchronizes everything we've learned about the models since we created the diagram, which is a lot.

- I misunderstood how Sequelize represents associations which resulted in the need to link some models with awkward association names like 'testPlanReportObject'. Now the associations are named 'testPlanReport' as you would expect.
- Updates column names to match the way we speak - no one says "publish status" so I changed it to "status". Changes like that.
- Removes columns that should be derived: for example, isManuallyTested can be derived from whether a tester is associated and coveragePercent can be derived from the results.
- Updates table name: TestPlans were actually TestPlanVersions all along.
- Removes TestResult table: the results should be stored in JSONB in the TestPlanRun table, which is symmetrical with the parsed JSONB field in the TestPlanVersion table.
- Updates the string representation of statuses to be all caps, which matches how GraphQL will expect it.

When merged, it will require you to rebuild your local database.

I doubt we'll need more migrations of this scale in the immediate future!

This wouldn't have been possible without the tests Howard wrote. Thank you!!

~~We should merge https://github.com/w3c/aria-at-app/pull/296 first so we can confirm this doesn't break the build.~~ Build is now fixed.


server/models/TestPlanRun.js
```diff
@@ -10,9 +10,9 @@ module.exports = function(sequelize, DataTypes) {
                 primaryKey: true,
                 autoIncrement: true
             },
-            isManuallyTested: { type: DataTypes.BOOLEAN, defaultValue: false },
-            tester: { type: DataTypes.INTEGER, allowNull: true },
-            testPlanReport: { type: DataTypes.INTEGER }
+            testerUserId: { type: DataTypes.INTEGER, allowNull: true },
+            testPlanReportId: { type: DataTypes.INTEGER },
+            results: { type: DataTypes.JSONB }
```
howard-e, line 15: Previous comment in the migration script on the `results` attribute type applies here as well.

server/migrations/20210524210332-TestQueueChanges.js
```diff
@@ -0,0 +1,206 @@
...
+            );
+            await queryInterface.removeColumn(
+                'TestPlanRun',
+                'isManuallyTested',
+                { transaction }
+            );
+            await queryInterface.addColumn(
+                'TestPlanRun',
+                'results',
+                Sequelize.DataTypes.JSONB,
```
howard-e, line 79: May want to use a JSONB array here instead? To take advantage of the built in postgres handlers that will come with querying a collection of a json dataset.
Unless `results` will have some other shape beyond just being an array of test results to warrant that.

server/models/services/BrowserService.js
```diff
@@ -86,8 +86,8 @@ const createBrowser = async (
     browserAttributes = BROWSER_ATTRIBUTES,
     browserVersionAttributes = BROWSER_VERSION_ATTRIBUTES
 ) => {
-    const atResult = await ModelService.create(Browser, { name });
-    const { id } = atResult;
+    const browserResult = await ModelService.create(Browser, { name });
```
howard-e, line 89: Good catch on correcting the variable name here!

server/migrations/20210524210332-TestQueueChanges.js
alflennik: Ah, good idea! I didn't know about that.

server/models/services/TestPlanVersionService.js
```diff
@@ -0,0 +1,237 @@
...
+    testPlanVersionAttributes = TEST_PLAN_VERSION_ATTRIBUTES,
+    testPlanReportAttributes = TEST_PLAN_REPORT_ATTRIBUTES,
+    testPlanTargetAttributes = TEST_PLAN_TARGET_ATTRIBUTES,
+    testPlanRunAttributes = TEST_PLAN_RUN_ATTRIBUTES,
+    userAttributes = USER_ATTRIBUTES
+) => {
+    await ModelService.update(
+        TestPlanVersion,
+        { id },
+        { title, status, gitSha, gitMessage, exampleUrl, updatedAt, parsed }
```
howard-e, line 218: `parsed` attribute here for `TestPlanVersion` should be unsupported now

howard-e (commented):
Looks good! I also think it was a really good move to break `tests` and `metadata` out of the `TestPlanVersion.parsed` attribute.

Other than the `parsed` attribute still being referenced in the update method, looks good to me.

howard-e (approved):

jesdaigle (approved):

PR #299, opened 2021-05-26
GraphQL Test Queue Backend

alflennik:
Adds GraphQL functionality to support the Test Queue page as it stands today, including supporting fetching the data, assigning testers, deleting runs and updating the report status.


server/models/services/TestPlanReportService.js
```diff
@@ -159,121 +158,9 @@ const updateTestPlanReport = async (
     );
 };
 
-// Custom Functions
```
alflennik, line 162: I moved these business-logic-y functions into the resolvers!

server/resolvers/TestPlanReportOperations/assignTesterResolver.js
```diff
@@ -0,0 +1,22 @@
...
+const {
+    createTestPlanRun
+} = require('../../models/services/TestPlanRunService');
+
+const assignTesterResolver = async (
+    { parentContext: { id: testPlanReportId } },
+    { user: testerUserId },
+    { user }
+) => {
+    if (!user.roles.includes('ADMIN')) {
```
howard-e, line 11: nit-pick: Could probably have some global-level constants defined for the roles just as some typing safety, chance of any to the role name (unlikely), etc - if they'll be referenced often. Could also be defined directly on the `User` model

howard-e (approved):
Looks good. Certainly seems to provide everything we'd need. Also, viewing and running the GraphQL tests certainly helped with understanding some of these derived associations and properties so thanks for that.

I left follow up questions just for my understanding on 2 currently unused attributes so they block nothing for me at the moment so otherwise, this all looks good to me!

server/graphql-schema.js
```diff
@@ -65,40 +65,108 @@ const graphqlSchema = gql`
...
+        passed: Boolean!
+    }
+
+    type UnexpectedBehavior {
+        id: ID!
+        description: String!
+    }
+
+    type TestResultConflict {
+        breadcrumbs: String! # TBD
```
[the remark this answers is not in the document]
alflennik: It's actually pretty common to need to refer / point at deeply nested places in the graph, so I was thinking about introducing a type that would allow that. It's a way to say, "I want you to look at this test plan, this version, this report and this run."

It's not actually a string, that's just a placeholder.

My current best name for it is "Traversal" and it's starting to take form in my current branch. But definitely still very TBD. I can show more on Monday ... assuming I don't decide it's all a horrible idea, haha.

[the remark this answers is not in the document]
howard-e: Ok I follow you. Great idea there
And I hope it's the former for Monday, rather than the latter 😄

server/graphql-schema.js
```diff
@@ -65,40 +65,108 @@ const graphqlSchema = gql`
...
+    interface Assertion {
+        # TODO: account for at-specific assertions
+        # TODO: account for optional assertions
+        # TODO: account for automation
+        manualAssertion: String!
+    }
+
+    interface PassThrough {
+        atMode: String!
+        nthCommand: Int!
```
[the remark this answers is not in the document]
alflennik: The test authors have a way to specify a group of commands which should be interchangeable within a certain test. In practice it seems that pretty much every interaction we're testing can be accomplished in at least a few different ways.

So in order to completely run a test, manual testers have to re-run it once for each available and interchangeable command. This could mean three, four, eight or more "pass throughs."

nthCommand, then, is used to cycle through all the available commands, and it would apply to all the groups of commands which are used within the test.

It's possible the "nthCommand" part might get replaced with an actual command type in the future... it is still a tiny bit blurry!

server/resolvers/TestPlanReportOperations/assignTesterResolver.js
alflennik: So something like `user.roles.includes(User.ADMIN)`? Yes, I'd be willing to try that!

howard-e: Yep, exactly that. Cool!

server/graphql-schema.js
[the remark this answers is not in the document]
howard-e: Got it! Thanks for the explanation.

jesdaigle (approved):

server/resolvers/TestPlanReportOperations/assignTesterResolver.js
alflennik: This will be in my next PR.

PR #301, opened 2021-05-28
Finalizing implementation of new models

howard-e:
This includes changes to finalize the implementation of our models for the new database schema which were not covered in previous PRs:

Changes include:
* Adding a model service for the `Role` model
* Adding test assertions for the `RoleService`
* Refactoring and adding a few more assertions to the model services' tests to verify different structures and the existence of associated attributes which can be controlled through `ModelService`


alflennik (approved):
These tests have been so useful at validating whether or not my changes are work or not, so bulking up that system is awesome. Nothing looks out of place 👍

jesdaigle (approved):

PR #302, opened 2021-06-04
New test queue mutations

alflennik:
Ready for review. ~~But: there are still three tests left to write for features I added to the model service. All the tests related to the GraphQL mutations are complete and passing.~~ All tests have been added.

I fixed a couple problems with the database schema that I missed in the previous migration. It will require rebuilding your local test database.

I further improved the model service, adding a few more associations, a few more functions to individual models, and added a new getOrCreate function that can be used by any model.

The biggest change to the schema is the locationOfData type which contains the ids which point to a given record. For example, `{ testPlanRunId: 3 }` could be one such locationOfData. Using that locationOfData, you can populate out all the data which is known just from that one id: the TestPlan, TestPlanVersion, TestPlanReport, Tester, TestPlanTarget, At, Browser, At Version and Browser Version. It's already proven useful! I think it's pretty cool :)


server/resolvers/PopulatedData/index.js
```diff
@@ -0,0 +1,103 @@
...
+        throw new Error(
+            'LocationOfData without an associated test plan version is not ' +
+                'yet implemented. Note you may see this error if the data ' +
+                'you requested could not be found.'
+        );
+    }
+    testPlan = { id: testPlanVersion.metadata.directory };
+
+    const idsContradict = (provided, found) => {
+        return provided && provided.toString() !== found.id.toString();
```
howard-e, line 70: nit: any safety check for `found` needs to be done here as well?

server/graphql-schema.js
```diff
@@ -197,46 +222,91 @@ const graphqlSchema = gql`
...
+        browserId: ID
+        browserVersion: String
+        atId: ID
+        atVersion: String
+        testPlanRunId: ID
+        testResultIndex: Int
+        passThroughResultIndex: Int
+    }
+
+    type PopulatedData {
```
howard-e, line 260: nit: `PopulatedData`'s name may make its purpose a bit too ambiguous without looking through the attributes and where it's used as a return type. Maybe just add an additional comment here?

server/models/services/ModelService.js
```diff
@@ -132,28 +150,149 @@ const get = async (
...
+ *     [atVersion, isNewAtVersion],
+ *     [browserVersion, isNewBrowserVersion],
+ *     [testPlanTarget, isNewTestPlanTarget],
+ *     [testPlanReport, isNewTestPlanReport],
+ * ] = results;
+ *
+ * @param {[object|function]} getOptionsArray
+ * @returns {Promise<[[*,Boolean]]>}
+ */
+const nestedGetOrCreate = async getOptionsArray => {
```
howard-e, line 242: This is a pretty neat way to control what was needed

howard-e (commented):
I'm very impressed with all that has gone into this!

server/models/services/ModelService.js
```diff
@@ -1,3 +1,4 @@
+const { isFunction } = require('lodash');
```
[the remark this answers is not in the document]
alflennik: Ah, dang, it's annoying how yarn workspaces allow you to import stuff that's not in the correct package.json. I added it.

server/models/TestPlanTarget.js
```diff
@@ -11,20 +11,20 @@ module.exports = function(sequelize, DataTypes) {
                 autoIncrement: true
             },
             title: { type: DataTypes.TEXT },
-            at: { type: DataTypes.INTEGER },
-            browser: { type: DataTypes.INTEGER },
+            atId: { type: DataTypes.INTEGER },
+            browserId: { type: DataTypes.INTEGER },
```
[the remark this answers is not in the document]
alflennik: Fixed!

server/models/services/ModelService.js
alflennik: Thanks 😁

howard-e (approved):
Looks good!

server/graphql-schema.js
```diff
@@ -197,46 +222,94 @@ const graphqlSchema = gql`
...
+        testPlanRunId: ID
+        testResultIndex: Int
+        passThroughResultIndex: Int
+    }
+
+    """
+    All the records that are associated with the id(s) of a given locationOfData
+    """
+    type PopulatedData {
+        locationOfData: LocationOfData!
```
jesdaigle, line 264: @alflennik can you explain LocationOfData? What does that mean?

alflennik: Sure - the PR description has a basic description, but let me describe in more in depth.

It's actually quite common through the app to need to point to a particular piece of data. We often need to say something like, "I want this TestPlanRun on this TestPlanReport on this TestPlanVersion on this TestPlan".

An example of a pressing use case for this PR is the testPlanTarget, atVersion and browserVersion which are associated with a TestPlanReport. When we create a new TestPlanReport for the Test Queue, I wanted to provide a list of data which was newly created. That list might look like this:

```js
[
  { testPlanReportId: 12 },
  { testPlanReportId: 12, testPlanTargetId: 3 },
  { testPlanReportId: 12, testPlanTargetId: 3, atVersion: '2021.2' }
]
```

This means that we created 3 new records, a TestPlanReport, TestPlanTarget and AtVersion, and using this list we could communicate that information to the user.

And GraphQL can automatically take one of those small little objects like `{ testPlanReportId: 12, testPlanTargetId: 3, atVersion: '2021.2' }` and just by traversing the associations it can tell you all the data which is implied by those IDs: the TestPlan, the TestPlanVersion, the TestPlanReport, the TestPlanTarget, and the AtVersion.

Looking a bit into the future, I can see this being useful for stuff like the Test Reports pages, where we need to support navigating a giant tree of reports, especially when we need to load a bunch of data just from a given URL. And the idea originally came to me when I realized I needed it for expressing conflicts between TestPlanRuns.

Hopefully that helps!

jesdaigle (approved):

PR #304, opened 2021-06-17
Add Testers.txt File

alflennik:
Note: the base is currently GraphQL Client but when that is merged I will change it to develop.

Adds a testers.txt file with a list of users who should be allowed into the app as testers. Admins do not need to be included.


server/services/GithubService.js
```diff
@@ -76,7 +70,7 @@ module.exports = {
             response.data.data.viewer.username
         );
     },
-    async getGithubTeams({ githubAccessToken, githubUsername }) {
+    async isMemberOfAdminTeam({ githubAccessToken, githubUsername }) {
```
howard-e, line 73: Much cleaner way 👍

howard-e (approved):
Works well

jesdaigle (approved):

PR #308, opened 2021-06-29
Test Queue Frontend Implementation

howard-e:
This attempts to replicate the functionality that was present before the backend migrations as it relates to the Test Queue page in the frontend application.

A new feature added to the Test Queue page here however, is allowing a user to add a Test Plan to the Test Queue via a modal that is now present at the top of the page. The commit version/message that was previously there has been removed.

**Recommended Setup Script** 
```bash
    yarn db-init:dev;
else
fi;

yarn sequelize db:migrate;
yarn sequelize db:seed:all;
yarn workspace server db-import-tests:dev -c 4ca7842ea7777b668546e74c9b5ed5b09696d927;
```


howard-e:
Currently investigating the long running builds on the CI. Unable to reproduce locally at the moment.

jscholes:
@howard-e When do you think this will be testable for accessibility, and will that take place on the staging site?

s3ththompson:
@jscholes we are working to finish landing this PR and the dependent PRs in the next day or two, along with some internal accessibility testing. we plan to do a review with Matt and then we should have a link to share with you (and the community group) following that. perhaps we can discuss more about testing the new changes at the App meeting on Thursday. (At this stage, the preview link will be to our new [sandbox environment](https://github.com/w3c/aria-at-app/pull/292), not staging.)

client/components/TestQueue/index.jsx
```diff
@@ -1,329 +1,255 @@
...
+const TestQueue = ({ auth }) => {
+    const { loading, data, refetch } = useQuery(TEST_QUEUE_PAGE_QUERY);
+
+    const [testers, setTesters] = useState([]);
+    const [testPlanReports, setTestPlanReports] = useState([]);
+    const [structuredTestPlanTargets, setStructuredTestPlanTargets] = useState(
+        {}
+    );
+    const [deleteResultsDetails, setDeleteResultsDetails] = useState({});
+    const [isShowingDeleteResultsModal, enableDeleteResultsModal] = useState(
```
alflennik, line 28: Perhaps the setter function should start with `set`? I guess that's the way it's normally done at least!

client/components/TestQueue/index.jsx
```diff
@@ -1,329 +1,255 @@
...
-        activeRunConfiguration,
-        testsFetched,
-        usersById,
-        userId,
-        admin,
-        ats: state.ats
-    };
+    const { auth } = state;
+    return { auth };
 };
```
alflennik, line 253: I was imagining that auth was now fully in the GraphQL world, but it seems there's something still stuck in Redux. Is there a way to remove this? I think I mentioned somewhere else that I was worried about how confusing it might be to have Redux and GraphQL in the same component, since removing Redux will probably happen once GraphQL is enabled everywhere.

client/components/TestQueueRun/index.jsx
```diff
@@ -8,217 +9,275 @@ import {
...
-    }
+    const evaluateTestRunTitle = () => {
+        const { testPlanTarget, testPlanVersion } = testPlanReport;
+        const { title: testPlanTargetName } = testPlanTarget;
+        const { title: apgExampleName, directory } = testPlanVersion;
 
-    renderOpenAsDropdown() {
-        const { runId, testers, usersById } = this.props;
+        return `${apgExampleName || directory} for ${testPlanTargetName}`;
+    };
```
alflennik, line 179: This seems like business logic so maybe we can add it to GraphQL - actually I thought maybe the database had this field already? Anyway I think this probably makes more sense on the server, or else we might end up repeating this function in multiple places.

client/index.js
```diff
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import { BrowserRouter } from 'react-router-dom';
 import { ApolloClient, InMemoryCache, ApolloProvider } from '@apollo/client';
 import { Provider } from 'react-redux';
-import store from './store';
+import store from './redux/store';
```
alflennik, line 6: I like the idea of isolating all the Redux. That's a good move that simplifies the visible files a huge amount and adds some much needed context. Yay.

client/tests/__mocks__/GraphQLMocks.js
```diff
@@ -0,0 +1,464 @@
...
+                            __typename: 'TestPlanVersion'
+                        },
+                        draftTestPlanRuns: [],
+                        __typename: 'TestPlanReport'
+                    }
+                ]
+            }
+        }
+    }
+];
```
alflennik, line 464: I think this is a valid approach - although part of me wonders if there's a way to use the dbCleaner functionality and the test database. That would allow us to test code that uses mutations and it would allow us to test the backend and frontend simultaneously. Maybe that's too ambitious but I'm curious what you think.

client/utils/formatter.js
```diff
@@ -0,0 +1,11 @@
+export const capitalizeFirstLetterOfWords = (
+    sentence,
+    splitChar = ' ',
+    joinChar = ' '
+) => {
```
alflennik, line 5: This one is totally optional, but what do you think about using a "options" object - that way, when you use it, it would be like:

```js
capitalizeFirstLetterOfWords(sentence, { joinChar: '_' })
```

Which allows readers to understand which options are passed, and makes it harder to mess up the order of arguments.

client/utils/formatter.js
```diff
@@ -0,0 +1,11 @@
+export const capitalizeFirstLetterOfWords = (
```
alflennik, line 1: Is `capitalizeSentence` a valid name here? I didn't look too closely at what it does so I'm not 100% sure.

server/graphql-schema.js
```diff
@@ -98,7 +98,7 @@ const graphqlSchema = gql`
 
     type TestPlanVersion {
         id: ID!
-        title: String!
+        title: String
```
alflennik, line 101: Interesting! 

Thinking from a business perspective, do you think it should remain required (and will use the directory name if an explicit name is not provided)?

server/resolvers/TestPlanVersion/directoryResolver.js
```diff
@@ -0,0 +1,5 @@
+const directoryResolver = testPlanVersion => {
+    return testPlanVersion.metadata.directory || '';
+};
+
+module.exports = directoryResolver;
```
alflennik, line 5: We can do this later, but in the future it would be better to store this as a column in the database - I think we agreed on that before. But it can stay like this for now since adding migrations would be a pain with a frontend-focused PR.

alflennik (commented):
Oh wow, this PR was so rich and full of well-considered decisions it took me two days to review. I think it's super impressive what you've accomplished here, and the only actual bugs I found were pretty microscopic, and are already fixed before I even started this review!

I couldn't resist weighing in on all the small decisions I see here, and I hope that's not too annoying. It's just fun to constantly poke at things 😆

Let me know what you think on the comments, I'm definitely willing to skip things you don't think are show stoppers.

Huge leap here!

client/components/AddTestPlanToQueue/AddTestPlanToQueueContainer.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Button } from 'react-bootstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faPlus } from '@fortawesome/free-solid-svg-icons';
+import PropTypes from 'prop-types';
+
+const AddTestPlanToQueueContainer = ({ handleOpenDialog = () => {} }) => {
```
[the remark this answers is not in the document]
howard-e: Performance wise, it's not much of a problem and it would be easier for persons to see the types and defaults in a single place. However, the React team has plans to deprecate the use of defaultProps for **functional** components here: https://github.com/facebook/react/pull/16210 with the ES6 standard being preferred which is why I forego its usage. You can read more on the motivation behind that here as well: https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md

client/components/AddTestPlanToQueue/AddTestPlanToQueueModal.jsx
```diff
@@ -0,0 +1,273 @@
+import React, { useState, useEffect } from 'react';
+import { Button, Modal, Form } from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import { useQuery, useMutation } from '@apollo/client';
+import {
+    ADD_TEST_QUEUE_MUTATION,
+    POPULATE_ADD_TEST_PLAN_TO_QUEUE_MODAL_QUERY
+} from '../TestQueue/queries';
```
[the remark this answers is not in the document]
howard-e: > When they're all extracted out it become harder to track! And the queries file might get uncomfortably long and hard to organize. For that reason I think it might be better to keep the queries in the component file.

Well a counter-point here is that having all the GQL statements in the component may potentially make it unnecessarily long and difficult to navigate.

The overall idea here was that the parent folder for that very parent and/or its children would have an associated query file of all the queries related to them. This may not be obvious due to the folder structure but `/components/AddTestPlanToQueue` would be a direct child of `/components/TestQueue` which is why I referenced the GQL statements from the `/component/TestQueue/queries` file in this instance.

> On the other hand, putting them in one file would make things like GraphQL fragments easier to pass around between queries. Maybe there are other good reasons I haven't thought of. What do you think?

I'm certainly a fan of having a dedicated file as a 'separation of concerns' for related queries but just for my clarification, did you mean a single file across the entire application?

client/components/AddTestPlanToQueue/AddTestPlanToQueueModal.jsx
```diff
@@ -0,0 +1,273 @@
...
+        POPULATE_ADD_TEST_PLAN_TO_QUEUE_MODAL_QUERY
+    );
+    const [addTestPlanReport] = useMutation(ADD_TEST_QUEUE_MUTATION);
+
+    useEffect(() => {
+        if (data) {
+            const {
+                ats = [],
+                browsers = [],
+                testPlanVersions: testPlans = []
```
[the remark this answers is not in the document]
howard-e: Think I was actually just avoiding a variable clash here and couldn't come up with something better at the time 🙂. Apologies on that confusion.

>  I think the right way to fix it would be to update the query to something like this:

I do know the query you speak of though. When we initially took a look at it, it doesn't return what was exactly wanted here and I do know you had further plans there so didn't want to have a clash in that method. However, there is a `filteredTestPlans` variable in this file which I believe may come close to what we'd want in that GQL resolver so having that functionality expanded and completed on the resolver would be great.

server/graphql-schema.js
```diff
@@ -107,6 +107,7 @@ const graphqlSchema = gql`
         # isLatest: Boolean!
         updatedAt: Timestamp!
         exampleUrl: String!
+        directory: String!
```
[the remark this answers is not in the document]
howard-e: Well would there ever be any need for the application to know the actual ID of the `TestPlanVersion` record? How to definitely know a previously reported on version has been been outdated comes to mind.

server/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "DOTENV_CONFIG_PATH=../config/dev.env nodemon --ignore './scripts/' -r dotenv/config server.js --watch",
     "dev-debug": "DEBUG=express:* DOTENV_CONFIG_PATH=../config/dev.env nodemon --inspect -r dotenv/config server.js --watch",
-    "jest": "SEQUELIZE_TEST_HELPERS_JEST_RUNNER=true dotenv -e ../config/test.env -- jest",
+    "jest": "SEQUELIZE_TEST_HELPERS_JEST_RUNNER=true dotenv -e ../config/test.env -- jest --detectOpenHandles --forceExit",
```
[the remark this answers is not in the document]
howard-e: Nope, this is intentional for the moment due to a performance hang inside `server/tests/integration/authentication.test.js` that happens since the move to using the GitHub Action CI in place of Travis.
We can probably create a separate issue for this!

server/resolvers/TestPlanReportOperations/assignTesterResolver.js
```diff
@@ -9,7 +9,13 @@ const assignTesterResolver = async (
...
     { user }
 ) => {
-    if (!user.roles.includes('ADMIN')) {
+    const roles = user.roles.map(role => role.name);
+
+    // prettier-ignore
+    if (
+        !roles.includes('ADMIN') &&
+        (roles.includes('TESTER') && testerUserId === user.id)
+    ) {
```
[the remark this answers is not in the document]
howard-e: Lol oh, good one. Will definitely try running through some use cases with that

client/components/TestQueueRun/index.jsx
```diff
@@ -8,217 +9,275 @@ import {
...
+    testPlanReport = {},
+    triggerDeleteResultsModal = () => {},
+    triggerTestPlanReportUpdate = () => {}
+}) => {
+    const startTestingButtonRef = useRef();
+    const [alertMessage, setAlertMessage] = useState('');
+
+    const [assignTester] = useMutation(ASSIGN_TESTER_MUTATION);
+    const [removeTester] = useMutation(REMOVE_TESTER_MUTATION);
+    const [removeTesterResults] = useMutation(REMOVE_TESTER_RESULTS_MUTATION);
```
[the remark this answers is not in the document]
howard-e: Well this was to support existing functionality so if we're removing that previous functionality, that's fine. Or are you saying to have both functionalities supported by a single mutation?

>  I think removing a test plan run should also remove its results

Well following some discussions, as long as a `TestPlanRun` that's in a FINAL / REVIEW state cannot be removed OR has its test results preserved somehow, I think that's fine

client/components/AddTestPlanToQueue/AddTestPlanToQueueContainer.jsx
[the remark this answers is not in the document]
howard-e: Sounds good to me!

client/components/AddTestPlanToQueue/AddTestPlanToQueueModal.jsx
```diff
@@ -0,0 +1,273 @@
...
+    const [selectedBrowser, setSelectedBrowser] = useState('');
+    const [atVersion, setAtVersion] = useState('');
+    const [browserVersion, setBrowserVersion] = useState('');
+    const [selectedTestPlan, setSelectedTestPlan] = useState('');
+    const [selectedTestPlanVersion, setSelectedTestPlanVersion] = useState('');
+
+    // eslint-disable-next-line no-unused-vars
+    const { loading, error, data } = useQuery(
+        POPULATE_ADD_TEST_PLAN_TO_QUEUE_MODAL_QUERY
+    );
```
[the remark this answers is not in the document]
howard-e: I'll try that out!

client/components/AddTestPlanToQueue/AddTestPlanToQueueModal.jsx
```diff
@@ -0,0 +1,273 @@
...
+    const [filteredTestPlans, setFilteredTestPlans] = useState([]);
+    const [testPlanVersions, setTestPlanVersions] = useState([]);
+    const [selectedAt, setSelectedAt] = useState('');
+    const [selectedBrowser, setSelectedBrowser] = useState('');
+    const [atVersion, setAtVersion] = useState('');
+    const [browserVersion, setBrowserVersion] = useState('');
+    const [selectedTestPlan, setSelectedTestPlan] = useState('');
+    const [selectedTestPlanVersion, setSelectedTestPlanVersion] = useState('');
+
+    // eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
howard-e: Well I'd actually like to make sure we're properly accounting for `loading` and `error` states to make things a bit more obvious to the user. My thoughts here stem from when I was introduced to the application. I found myself falling into invalid states where what was going on wasn't explained to me unless I dug into the console. Additionally, there were unrecoverable error states where the text on the page would say 'Loading...' so there was some false reporting there.

So doing this and doing it properly against the variables provided for the network calls feels like a bare minimum to me so having a discussion around that would be great

client/components/App/App.jsx
```diff
@@ -1,39 +1,45 @@
...
-    const { client, loading, data } = useQuery(APP_QUERY);
+const App = ({ auth, dispatch }) => {
+    const { client, error, loading, data } = useQuery(ME_QUERY);
     const signinUrl = useSigninUrl();
     const location = useLocation();
 
+    const { isSignedIn, isSignOutCalled, isTester, isAdmin, username } = auth;
+
     const signOut = async () => {
+        dispatch(signOutAction());
```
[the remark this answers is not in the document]
howard-e: Actually not using the `user/users` reducers/actions. Created a new state variable, `auth`. I want to slowly remove `user / users` as well as the others as the refactor continues. Managing the auth session through redux is commonplace for me but if this is something easily done with Apollo here, we could remove this altogether and I'll follow your lead there.

client/components/InvalidRequest/index.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { Container } from 'react-bootstrap';
+import { Helmet } from 'react-helmet';
+
+const InvalidRequest = () => {
+    return (
+        <Container as="main">
+            <Helmet>
+                <title>{`Invalid Request | ARIA-AT`}</title>
```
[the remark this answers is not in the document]
howard-e: Copy & Paste detected 😅

client/components/TestQueue/index.jsx
```diff
@@ -1,329 +1,255 @@
...
-        if (!admin) {
-            configuredAtNameIds = currentUser.configured_ats.map(
-                a => a.at_name_id
-            );
-        } else {
-            configuredAtNameIds = ats.map(at => at.id);
-        }
+    const renderAtBrowserList = (title = '', testPlanReports = []) => {
+        // means structuredTestPlanTargets would have been generated
+        if (testPlanReports.length) {
```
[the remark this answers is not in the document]
howard-e: Good catch!

client/components/TestQueue/index.jsx
howard-e: I typically prepend with `enable` and `toggle` for my boolean hooks, but that's just a me thing 🙂 so I'll go with the common pattern here

client/components/TestQueue/index.jsx
howard-e: Think my previous comment on this covers this:

> Managing the auth session through redux is commonplace for me but if this is something easily done with Apollo here, we could remove this altogether and I'll follow your lead there.

Let me know if it doesn't

client/components/TestQueueRun/index.jsx
howard-e: It has for the constructed TestPlanTarget title - `<at> <atVersion> with <browser> <browserVersion>`, which is `testPlanTargetName` is. There isn't a combination for it with the Test Plan's title, or in the absence of the Test Plan's title, the actual test plan (directory) coming from the server though. We could definitely do that that if we don't want to create a utility method within the frontend for it

client/tests/__mocks__/GraphQLMocks.js
howard-e: Well I admittedly didn't look too long into it but best practice here indicates the use of the `MockedProvider` which could certainly become difficult to maintain the more this grows which is what this is for, or some automatic mocking HOC around the `ApolloProvider`.

> although part of me wonders if there's a way to use the dbCleaner functionality and the test database

Making use of the dbCleaner could certainly help with creating some more dynamic and robust tests. It'd definitely be ambitious as you said and will take some time ha, but it would be beneficial. We could look into it.

client/utils/formatter.js
howard-e: That approach is fine with me

client/utils/formatter.js
howard-e: it would be converting `hello_world` to `Hello World`. The function name of `capitalizeSentence` to me would mean converting to `Hello world`.

Maybe something like `capitalizeEachWord` is easier to process?

server/graphql-schema.js
howard-e: Well I think there would still be need for the application to know the `title` different from the `directory` here so I don't view it as a 'backup' in this instance.
However, I do think my changing of the schema here isn't valid. I believe I did this when nested queries were failing when pulling `TestPlanVersion`s with `null` titles. This should certainly be managed from the resolver.

server/resolvers/TestPlanVersion/directoryResolver.js
howard-e: Had the same thought when doing so! Agreed!

client/components/AddTestPlanToQueue/AddTestPlanToQueueContainer.jsx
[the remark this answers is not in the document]
alflennik: Ah, thank you for that info, I had 0 idea. Good thing you noticed so we didn't get bit in the behind later.

client/components/AddTestPlanToQueue/AddTestPlanToQueueModal.jsx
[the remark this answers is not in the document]
alflennik: > The overall idea here was that the parent folder for that very parent and/or its children would have an associated query file of all the queries related to them.

Ah, interesting - you're right, I missed the implication there.

Okay, that definitely mitigates the point about the query file getting too long, then. There is a point where it won't need to be extended any more.

I thought of another pro to this approach, which is that it might potentially be easier to hook up to suspense for data fetching if that ever makes it into React:

The problem with suspense for data fetching is that it doesn't solve the "waterfall" problem, where a parent component makes a request, then loads a child component which makes a request, then loads another child component which makes a request. The requests happen sequentially instead of in parallel. So to reap the benefits of suspense, you have to group the queries for child components in to a parent component ... which is quite similar to what we have here.

All that to say, I think I get why you went with it - I'm on board!

server/graphql-schema.js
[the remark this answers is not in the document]
alflennik: I'm not sure I understand that! There is an ID for TestPlanVersions, right?

server/package.json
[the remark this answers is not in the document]
alflennik: Oh no, that test is causing issues? I'll fix it when I get a chance.

client/tests/__mocks__/GraphQLMocks.js
alflennik: Ok, perhaps it's an option ... but no need to hold this PR up over it 😜

client/utils/formatter.js
alflennik: Aaaaah, ok, up to you! I think both are good.

alflennik:
@howard-e Thanks for all the context! I felt quite comfortable after hearing more about your motivations so that means there really isn't much to tweak at all.

The one thing still open in my mind is question of how to think about the test plan version - it definitely might make sense to discuss in a call.

server/graphql-schema.js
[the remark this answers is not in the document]
howard-e: Apologies, I just noticed the distinction you were making between 'Test Plan' and 'Test Plan Version' here.

Well with that being the case, I suppose it's just for us to finalize the following query:

```graphql
testPlans {
  latestTestPlanVersion {
    id
    title
  }
}
```
Also with the inclusion of that `id` field as the `directory` like you've mentioned. 

What I've experienced with it to be clear is that it outputs the information for just a single test plan ('checkbox') all the time. With that complete, I suppose it would clear this up in the intended way

client/components/AddTestPlanToQueue/AddTestPlanToQueueModal.jsx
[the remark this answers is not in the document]
howard-e: > The problem with suspense for data fetching is that it doesn't solve the "waterfall" problem, where a parent component makes a request, then loads a child component which makes a request, then loads another child component which makes a request. The requests happen sequentially instead of in parallel. So to reap the benefits of suspense, you have to group the queries for child components in to a parent component ... which is quite similar to what we have here.

Yep, I agree that it could help out there. Good observation

server/package.json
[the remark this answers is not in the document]
alflennik: I fixed this! And we merged it into the PR 👍

server/resolvers/TestPlanReportOperations/updateStatusResolver.js
```diff
@@ -12,7 +12,11 @@ const updateStatusResolver = async (
     { status: status },
     { user }
 ) => {
-    if (!user.roles.includes('ADMIN')) {
+    let roles = [...user.roles];
```
howard-e, line 15: @alflennik seems there's 2 structures of the user roles coming through to this resolver.

`[ 'ADMIN' ... ]` and `[ { name: 'ADMIN' }, ... ]`

Currently accounting for both and this isn't a blocker to the PR but investigation may be required. I've went ahead and copied this logic to other resolvers dependent on the user role as well.

alflennik: Thanks for finding this - sounds like you were able to fix it. Thank you!

alflennik (approved):

jesdaigle (approved):
I just have to say that this code, and the teamwork, is beautiful. Nice work @howard-e and @alflennik

PR #309, opened 2021-07-02
Allow import-tests to account for both the old and new build structure when importing from aria-at

howard-e:
Given the changes introduced by https://github.com/w3c/aria-at/pull/451, the `import-tests.js` has to account for both the old and new structure when importing tests from https://github.com/w3c/aria-at. Otherwise, rows would be inserted into the `TestPlanVersion` table without the `TestPlanVersion.tests` column being populated.

**Notes**
1. Assuming your database is fresh, run the following script.
2. Query the `TestPlanVersion` table and observe that rows related to both 'versions' of the https://github.com/w3c/aria-at that have been imported have their `tests` columns populated.
```bash
    yarn db-init:dev;
else
fi;

yarn sequelize db:migrate;
yarn sequelize db:seed:all;
yarn db-import-tests:dev; # imports most recent commit from aria-at
yarn workspace server db-import-tests:dev -c 4ca7842ea7777b668546e74c9b5ed5b09696d927; # most recent commit hash on aria-at before #451 was merged
```


alflennik:
@howard-e Sorry I somehow missed this PR!

I checked out the branch, and merged develop into it, and then I triggered it by calling `http://localhost:3000/api/test/import` with a body of `{ "git_hash": "137716071d58d6d404d3dca45f8d6c12accb8388" }`, which is the most recent commit on ARIA-AT.

I got the following error:

```
Reference file, /Users/alflennik/Projects/aria-at-app/server/scripts/import-tests/tmp/tests/combobox-autocomplete-both/data/references.csv, does not exist!
Error found: Error: ENOENT: no such file or directory, open '/Users/alflennik/Projects/aria-at-app/server/scripts/import-tests/tmp/tests/combobox-autocomplete-both/data/references.csv'
    at Object.openSync (node:fs:505:3)
    at Object.readFileSync (node:fs:401:35)
    at Object.getMostRecentTests (/Users/alflennik/Projects/aria-at-app/server/scripts/import-tests/index.js:131:41)
```

howard-e:
Thanks for pointing this out @alflennik 

`combobox-autocomplete-both` was an old test plan that has now been removed from https://github.com/w3c/aria-at/tree/master/tests following the merge of https://github.com/w3c/aria-at/pull/417. It seems that the cleanup task when test plans are re-generated with `npm run build` did not address removing leftover files of this test plan, located at:
* https://github.com/w3c/aria-at/tree/master/build/tests/combobox-autocomplete-both/reference &
* https://github.com/w3c/aria-at/blob/master/build/review/combobox-autocomplete-both.html

because of the presence of these `./build` files without the `./tests/combobox-autocomplete-both/` counterpart, this error was thrown.

I will adjust the PR to make an explicit check for the validity of the test plans based on `./tests/`.

Additionally, the cleanup task on _aria-at_ when running `npm run build` will also need to be revised to account for this
cc: @s3ththompson

alflennik (approved):
I tested very old commits and was able to import them no problem, and I tested the newest commit and was able to import all 16 test plans. I also confirmed I can run the import script from the command line, a cron job and through the HTTP API, and it worked in all three cases 🤩

howard-e:
Thanks for the review @alflennik! Closing in favor of an updated `server/scripts/import-tests/index.js` being introduced in #315.

PR #313, opened 2021-07-23
TestPlanRun GraphQL Support

howard-e:
Adds GraphQL support for retrieving and updating `TestPlanRun` related information.

**Sample Queries & Mutations:**
_querying `tests` through `testPlanVersions`_
```gql
query {
  testPlanVersions {
    id
    title
    gitSha
    directory
    testPlanReports {
      id
      createdAt
      status
    }
    tests {
      title
      index
      testFilePath
    }
  }
}
```

_querying `testResults` through `testPlanReport`_
```gql
query {
  testPlanReport(id: 1) {
    id
    status
    testPlanTarget {
      title
    }
    testPlanVersion {
      title
      tests {
        title
        index
        testFilePath
      }
    }
    draftTestPlanRuns {
      id
      tester {
        id
        username
      }
      testResults {
        title
        index
        testFilePath
        assertions {
          manualAssertion
          priority
          command
        }
      }
    }
  }
}
```

_querying `testResults` through `testPlanRun`_
```gql
query {
  testPlanRun(id: 1) {
    testResultCount
    tester {
      id
      username
    }
    testResults {
      title
      index
      testFilePath
      isComplete
      isSkipped
      assertions(priority: REQUIRED) {
        command
        priority
        manualAssertion
      }
      assertionsCount(priority: REQUIRED)
      assertionsPassed(priority:REQUIRED)
      unexpectedBehaviorCount
      result {
        test
      }
      serializedForm {
        name
      }
      issues
    }
  }
}
```

_mutating `testPlanRun.testResults` by updating `testPlanRun.testResult[].serializedForm`_
```gql
mutation {
  testPlanRun(id: 1) {
    updateTestResult(
      input: {
        index: 1
        test: {
          htmlFile: "tests/checkbox/test-01-navigate-to-unchecked-checkbox-reading.html"
          testFullName: "Navigate to an unchecked checkbox in reading mode"
          executionOrder: 1
        }
        serializedForm: [
          {
            name: "result-0-0"
            value: "on"
            checked: true
            disabled: false
            indeterminate: false
          }
          {
            name: "result-0-1"
            value: "on"
            checked: false
            disabled: false
            indeterminate: false
          }
        ]
      }
    ) {
      testPlanRun {
        testResultCount
        tester {
          id
          username
        }
        testResults {
          index
          serializedForm {
            name
          }
        }
      }
    }
  }
}
```

_mutating `testPlanRun.testResults` by removing `testPlanRun.testResult[].serializedForm` & `testPlanRun.testResult[].result`_
```gql
mutation {
  testPlanRun(id: 1) {
    clearTestResult(
      input: {
        index: 1
      }
    ) {
      testPlanRun {
        testResultCount
        tester {
          id
          username
        }
        testResults {
          index
          isComplete
          result {
            test
          }
          serializedForm {
            name
          }
        }
      }
    }
  }
}
```


alflennik:
I'm testing the GraphQL endpoints! Exciting to get some data back in new places.

I got an error when I ran GraphQL query for the results of a test plan run which I hadn't started yet, I think because it returned undefined for a field instead of an empty array.

howard-e:
Got it! Thanks for the early check there

server/resolvers/TestPlanRunOperations/updateResultResolver.js
```diff
@@ -0,0 +1,40 @@
...
+        testResults[testResultToUpdateIndex].test.isComplete = false;
+        testResults[testResultToUpdateIndex].test.completedAt = null;
+    } else {
+        testResults[testResultToUpdateIndex].result = {
+            ...testResults[testResultToUpdateIndex].result,
+            ...result
+        };
+        testResults[testResultToUpdateIndex].test.isComplete = true;
+        testResults[testResultToUpdateIndex].test.completedAt = new Date();
+    }
```
alflennik, line 31: Using a TestResultInput type might help result in a lot of this code not being needed - let me know when you're ready to work on it and maybe we can pair on it!

server/graphql-schema.js
```diff
@@ -113,54 +116,71 @@ const graphqlSchema = gql`
...
         # TODO: account for at-specific assertions
         # TODO: account for optional assertions
         # TODO: account for automation
         manualAssertion: String!
     }
 
+    type Assertion implements BaseAssertion {
+        manualAssertion: String!
+        command: String!
+        priority: String!
```
alflennik, line 152: Instead of priority I had it in my head that we might be able to just have optional and required assertions. If we're using priority I would definitely want to remove the term optional / required from GraphQL since that would mean we have two terms for the same thing.

server/graphql-schema.js
```diff
@@ -113,54 +116,71 @@ const graphqlSchema = gql`
...
+    interface BaseTest {
         title: String!
         index: Int!
+        referenceFilePath: String!
+    }
+
+    type Test implements BaseTest {
+        title: String!
+        index: Int!
+        referenceFilePath: String!
```
alflennik, line 128: Is this the `exampleUrl`? If so, that's in the schema as being part of the TestPlanVersion type.

server/resolvers/TestResult/unexpectedBehaviorCountResolver.js
```diff
@@ -0,0 +1,11 @@
+const unexpectedBehaviorCount = testResult => {
+    const { result } = testResult;
+    const { details } = result;
+    const { summary } = details;
+
+    if (!details || !summary) return 0;
+
+    return summary.unexpectedCount;
```
alflennik, line 8: The database values are quite different from the GraphQL values and I'm worried this could get confusing! Is this something you think we can clean up as we go or do think we can clean it up before we merge?

alflennik (commented):
After being slow to review for a couple times I wanted to take this opportunity to be early for once! Looking good so far.

server/graphql-schema.js
```diff
@@ -113,54 +116,71 @@ const graphqlSchema = gql`
         testPlanReports(status: TestPlanReportStatus): [TestPlanReport]!
     }
 
-    interface Test {
+    interface BaseTest {
```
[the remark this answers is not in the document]
howard-e: Well the interface can't act as a type unless explicitly defined as a `type`. But if you did mean to make that change, I do like the idea of having a "Base" type for a test. I figure additional types may come into play later to implement it such as an `AutomatedTest` in addition to `ManualTest` (what the `Test` type is now that I think about it) which is why I've left `BaseTest` interface but we can discuss further. I assume properties such as `optionalAssertionCount` and the like may not need to exist in `AutomatedTest`

server/graphql-schema.js
```diff
@@ -113,54 +116,71 @@ const graphqlSchema = gql`
...
         instructions: [Instruction]!
         assertions: [Assertion]!
         passThroughs: [PassThrough]!
         assertionCount: Int!
         optionalAssertionCount: Int!
 
-        startedAt: Timestamp!
-        completedAt: Timestamp!
+        startedAt: Timestamp
+        completedAt: Timestamp
```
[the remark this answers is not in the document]
howard-e: > Is it possible to have test results that aren't complete?

Seems like you can have test results that aren't complete or rather, they are 'partially complete'. Once it's skipped, it is considered 'partially complete'. And it being skipped just means those results weren't submitted

> So startedAt still seems like it should be required!

Yep, I agree with that!

server/graphql-schema.js
```diff
@@ -9,6 +9,9 @@ const graphqlSchema = gql`
     """
     scalar Timestamp
 
+    # Required for passing free form Object data
+    scalar Object
```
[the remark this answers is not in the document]
howard-e: I added the support for this as we're still unsure what exactly we'll be getting back and using from the harness and how much it may differ from the previous implementation. For that reason, I'm expecting the information that will updated on `TestPlanRun.testResults` will continue to be free-form data until that's locked in. I suppose at that point, we can create the proper types here?

server/graphql-schema.js
```diff
@@ -309,6 +330,12 @@ const graphqlSchema = gql`
         updateStatus(status: TestPlanReportStatus!): PopulatedData!
     }
 
+    type TestPlanRunOperations {
+        # Allow null so tester may 'start over'
+        updateResult(result: Object, index: Int!): PopulatedData!
```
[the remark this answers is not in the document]
howard-e: Okay, I see. Got it!

server/resolvers/TestPlanRunOperations/updateResultResolver.js
howard-e: Sure, we could do that

server/graphql-schema.js
```diff
@@ -309,6 +330,12 @@ const graphqlSchema = gql`
...
+        updateSerializedForm(form: [Object], index: Int!): PopulatedData!
```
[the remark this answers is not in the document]
howard-e: Was going along with the shape of the `testResults` array object being:
```javascript
{
    test: { ..., executionOrder: 1, ... },
    result: { ... },
    serializedForm: { ... }
}
```

> I wonder if there's a way to operate on just the mutation above - do we need both? Is this, like a bulk version of the mutation above?

These are separate operations on the `result` and `serializedForm` objects. I initially felt having them separate would provide greater control of the data when I started.

> do we need both?

I do agree having a single operation for both is also okay looking back

server/graphql-schema.js
howard-e: Well, it is referred to as an assertion priority:
> Assertion priority: Specifies whether an assertion is “must have” or “optional”.

[From _Working Mode_](https://github.com/w3c/aria-at/wiki/Working-Mode)

> just have optional and required assertions

Maintaining that here to me seems to be something like 2 separate booleans that would say `isRequired` or `isOptional`? But the value of `priority` here could realistically just be `required` or `optional` or any other priority 'state' that may be introduced.

> If we're using priority I would definitely want to remove the term optional / required from GraphQL since that would mean we have two terms for the same thing.

I'm not exactly seeing the need for anything to get replaced or removed here

server/graphql-schema.js
howard-e: That is actual the reference file for the Test Page to be opened, eg. `tests/checkbox/test-01-navigate-to-unchecked-checkbox-reading.html`, not the APG example that TestPlanVersion would have

server/resolvers/TestResult/unexpectedBehaviorCountResolver.js
howard-e: Actually, these are the same variables currently stored for our examples in the `TestPlanRun.testResults` table columns. They were originally ported over from the old implementation. This object structure and variables come directly from the result of the iframe's result when submitted.
What are your suggestions here?

server/graphql-schema.js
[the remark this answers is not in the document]
alflennik: Oh wow, I'm totally wrong about interface types 😱

I'm not sure about the AutomatedTest and ManualTest distinction yet, but we can talk about that when the task requires it.

server/graphql-schema.js
alflennik: > I'm not exactly seeing the need for anything to get replaced or removed here

I was confused about why you were confused so I looked at the schema and saw that while I had included the term `optional`, I never included the term `required`.

That makes what I said ... confusing.

So the previous schema didn't quite model assertion priority well enough to avoid being confusing, and I want to fix that. One way would be to rename the "assertions" and "assertionCount" fields to "requiredAssertions" and "requiredAssertionCount" ... but I think you're actually onto something with the priority field, which is honestly a lot simpler. What about this:

```gql
    type TestResult implements BaseTest {
        title: String!
        index: Int!
        # ...
        assertions(priority: AssertionPriority): [Assertion]!
        assertionCount(priority: AssertionPriority): Int!
        passThroughs: [PassThrough]!
        # optionalAssertionCount: Int! # remove
        # ...
        passThroughResults: [PassThroughResult]!
        assertionsPassed(priority: AssertionPriority): Int!
        # optionalAssertionsPassed: Int! # remove
        unexpectedBehaviorCount: Int!
        # ...
    }
```

Plus an enum for AssertionPriority:

```gql
    enum AssertionPriority {
        REQUIRED
        OPTIONAL
    }
```

This would allow us to keep the priority field, reduce the number of fields, and would allow splitting up the optional and required assertions as needed in the UI. What do you think about that? And do you understand why I think there are two conflicting approaches in place right now?

server/resolvers/TestResult/unexpectedBehaviorCountResolver.js
alflennik: I think we should design an elegant data structure that fits nicely into GraphQL, which I think is what we're going to do with the TestResultInput type, and then I think we have two options about how to handle the old model:

First, we can implement [an anticorruption layer](https://dev.to/asarnaout/the-anti-corruption-layer-pattern-pcd), basically a remapping function takes an instance of the old model and returns the same data but organized in the same way as the GraphQL model. We do that immediately, probably on the frontend, or as soon as possible, to avoid using the old model in any code. The idea there is that having inaccurate or out-of-date models will make the terminology less meaningful, or "corrupt" it.

Second, and this is probably what we have to do eventually, but is probably too difficult to do immediately, is actually rewrite the system so that it produces data in the correct format without any remapping needed.

server/graphql-schema.js
howard-e: Yep, I understand the disconnect here. And this approach could work

server/graphql-schema.js
howard-e: Yep

server/resolvers/TestResult/unexpectedBehaviorCountResolver.js
howard-e: > First, we can implement an anticorruption layer, basically a remapping function takes an instance of the old model and returns the same data but organized in the same way as the GraphQL model. We do that immediately, probably on the frontend, or as soon as possible, to avoid using the old model in any code. The idea there is that having inaccurate or out-of-date models will make the terminology less meaningful, or "corrupt" it.

Do you mean doing something like this? https://github.com/w3c/aria-at-app/blob/614f9234eb218c8aec26688b7c79bfb02768fbde/server/util/validateTestResult.js#L58

> Second, and this is probably what we have to do eventually, but is probably too difficult to do immediately, is actually rewrite the system so that it produces data in the correct format without any remapping needed.

Agreed that it could come to that but too big of an ask at the current moment

server/graphql-schema.js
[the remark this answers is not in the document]
howard-e: Introduced temporary types. Need to document their usage.

server/resolvers/TestResult/unexpectedBehaviorCountResolver.js
howard-e: > I think we should design an elegant data structure that fits nicely into GraphQL, which I think is what we're going to do with the TestResultInput type

@alflennik and I experimented with this and it proved to be too much effort currently

server/graphql-schema.js
```diff
@@ -113,54 +116,71 @@ const graphqlSchema = gql`
         testPlanReports(status: TestPlanReportStatus): [TestPlanReport]!
     }
 
-    interface Test {
+    interface BaseTest {
         title: String!
         index: Int!
+        referenceFilePath: String!
```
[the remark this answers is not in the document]
howard-e: @alflennik and I spoke on this and agreed that just a revision of the variables might be sufficient

server/resolvers/PopulatedData/index.js
```diff
@@ -66,6 +66,18 @@ const PopulatedData = async ({ parentContext: { locationOfData } }) => {
...
+        testPlanRun &&
+        testPlanRun.testResults &&
+        testPlanRun.testResults.length
+    ) {
+        const testResults = testPlanRun.testResults.map(each => ({
+            ...each.test,
+            ...each
+        }));
+        testPlanRun.testResults = [...testResults];
+    }
```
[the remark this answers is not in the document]
howard-e: Good catch! This may be subject to change and I'll document as such

alflennik (approved):
This is awesome, onto the frontend! 😈

jesdaigle (approved):

PR #315, opened 2021-08-16
Update test run page & replace renderer's iframe

howard-e:
This aims to update the Test Run page's functionality to bring it up to date with our current backend implementation.

Before merge checklist:

- [x] Merge https://github.com/w3c/aria-at/pull/466
- [x] Merge #309 

**Recommended Setup Script** 
```bash
    yarn db-init:dev;
else
fi;

yarn sequelize db:migrate;
yarn sequelize db:seed:all;
yarn workspace server db-import-tests:dev;
```

The major feature addition here is that the iframe which would previously hold the test content is now built and ran locally and depends on https://github.com/w3c/aria-at/pull/466.

**Known Issues**
- [x] Focus is lost after closing Test Window - to be addressed


client/components/TestRenderer/index.jsx
```diff
@@ -0,0 +1,875 @@
...
+    userCloseWindow,
+    userOpenWindow
+} from '../../resources/aria-at-test-run.mjs';
+import {
+    TestRunExport,
+    TestRunInputOutput
+} from '../../resources/aria-at-test-io-format.mjs';
+import { TestWindow } from '../../resources/aria-at-test-window.mjs';
+
+const Container = styled.div`
```
howard-e, line 15: The benefit of using styled components may not come in the short term. Would rather to revert to using a custom stylesheet for this component

client/components/RaiseIssueModal/index.jsx
```diff
@@ -1,197 +1,154 @@
...
             <IssueCards key="render-issues-cards" issues={issues} />,
             <Fragment key="render-issues-buttons">
-                <Button variant="secondary" onClick={onHide}>
-                    My issue exists in this list
+                <Button variant="secondary" onClick={onHideButtonClick}>
+                    My Issue exists in this list
                 </Button>
                 <Button variant="primary" onClick={onCreateNewIssueClick}>
-                    My issue is not in this list
+                    My Issue is not in this list
```
alflennik, line 129: This capitalization looks funny!

client/components/TestRenderer/index.jsx
alflennik: I like StyledComponents 🤓 I'd be down to gradually move over to it if you like.

sinabahram:
We should pick a style for buttons and then stick with it e.g. title case or just initial caps?

I'm confused about "My Issue is not in this list". I think the only two logical possibilities are "My Issue Is Not In This List" or "My issue is not in this list". I don't understand the caps on both "My" and "Issue".

Title case for buttons seems very logical IMHO, so that would be my vote.

alflennik (commented):
This is aweeesome, can't wait to merge it.

I'm curious about the aria-at-harness.mjs, aria-at-test-io-format.mjs and the other files in this resources folder - I think I sort of recognize them from the ARIA-AT repo? Are they copied here? Is that permanent? Or are they meant to stay?

No real notes besides that question, I should be able to approve pretty soon after you feel like the blockers are clear.

howard-e:
> @sinabahram 
> We should pick a style for buttons and then stick with it e.g. title case or just initial caps?

Agreed. Title Case v Sentence case is a decision to be made.

> @alflennik 
> I'm curious about the aria-at-harness.mjs, aria-at-test-io-format.mjs and the other files in this resources folder - I think I sort of recognize them from the ARIA-AT repo? Are they copied here? Is that permanent? Or are they meant to stay?

This isn't be permanent. This was mainly for ease of debugging purposes.

client/components/RaiseIssueModal/index.jsx
howard-e: Good catch!

client/components/TestRenderer/index.jsx
howard-e: Following an internal discussion, we'll be moving forward with styled components as it may promote ease of maintainability

howard-e:
cc @alflennik @jesdaigle, this is now for your review.

alflennik (approved):
Having tested this thoroughly on the sandbox, staging, etc., and using it as the base for my current branch, I can safely say this is ready for merge, and it's fantastic stuff!

jesdaigle (approved):
this is so pretty. 🤩

PR #319, opened 2021-08-30
Reporting Backend Queries

alflennik:
This PR takes the GraphQL schema from something half functional and full of guesses about how tests / assertions etc. would eventually be represented, and makes every type and field fully functional, as well as adding full interactive documentation.

The GraphQL playground showing docs for TestPlanVersion's metadata field.

- Each field is documented
- Nearly every field is tested
- The tests include a check to make sure fields we add in the future get tested as well 

With the reporting backend in place, this allows us to create encapsulated React components which can independently query the data they need, allowing for the prospect of richer reporting.

This is the backend only, and, if we decide to merge it, the Test Queue pages on the client will cease to function.

Future checklist to fix the Test Queue pages:

- Migrate the database to completely remove the old schema.
- Implement the several GraphQL mutations which have not yet been implemented.
- On the client, use the new API Z has implemented / is implementing for the Test Renderer.
- Update existing pages to use the reporting-aware GraphQL API.


howard-e (approved):
Looks good to me! I've been messing around with it a bit in the Playground and haven't seen any clear issues. Looking forward to having this integrated.

server/package.json
```diff
@@ -35,11 +35,13 @@
     "express": "^5.0.0-alpha.8",
     "express-session": "^1.17.2",
     "graphql": "^15.5.0",
+    "js-base64": "^3.6.1",
```
howard-e, line 38: Could the encoding and decoding of a b64 string you're using this library for be done using Buffer instead?

alflennik: Yes, definitely. The con is that it adds a dependency. There are two pros I think: the biggest plus for me is the fact that the characters it uses are the letters and numbers only (it doesn't insert `/` or `=` characters like Node does), which works better in IDs, and the second biggest plus is that the API is more readable.

PR #323, opened 2021-09-08
Sort TestPlanVersion records by updatedAt by default

howard-e:
Updates `testPlanVersionsResolver` to additionally sort the results of `TestPlanVersion` by the `updatedAt` column, by default.


alflennik (approved):
Thanks for addressing that so fast! 🙌

PR #324, opened 2021-09-10
Fix for conflictsResolver

howard-e:
Given the new TestRenderer's output, a reference update had to be made to the conflictsResolver.


alflennik (approved):
Confirmed fixed!

PR #325, opened 2021-09-13
Remove references for old implementation

howard-e:
Removes several references and code from the older implementation. These pieces will either be completely removed or refactored to be aligned with the new GraphQL structure.


client/utils/evalAuth.js
```diff
@@ -0,0 +1,15 @@
+export const evalAuth = (data = {}) => {
+    const roles = data.roles || [];
+
+    return {
+        // calculated booleans
+        isAdmin: roles.includes('ADMIN') || false,
```
alflennik, line 6: I don't think we need the `|| false` part!

Should it be `roles && roles.includes('ADMIN')` or is roles always going to be defined?

server/controllers/AuthController.js
```diff
@@ -87,7 +87,7 @@ const oauthRedirectFromGithubController = async (req, res) => {
     return loginSucceeded();
 };
 
-const signoutController = (req, res) => {
+const signoutController = async (req, res) => {
```
alflennik, line 90: What's this change for!

alflennik (approved):
Ah, I am super excited for these changes, this should make perusing the codebase much easier and make working with the code more welcoming to anyone who wants to dive in.

client/utils/evalAuth.js
```diff
@@ -0,0 +1,15 @@
+export const evalAuth = (data = {}) => {
```
alflennik, line 1: Super optional: maybe including "eval" in the name will make people think about the `eval` function in JS. Maybe some other word like "check" would work better?

client/utils/evalAuth.js
howard-e: This could just be `roles.includes('ADMIN')` actually since roles is always defined on L2.

server/controllers/AuthController.js
howard-e: An unnecessary change now :). I promisified the `destroy` in its own method to use elsewhere in the `AuthController` as well but there was no real benefit to it. Forgot to remove the async keyword here.

client/utils/evalAuth.js
howard-e: Well it's also not really a "check" I'd say. That indicates a boolean to me. It is forming (evaluating?) a simplified auth object to be returned.

> maybe including "eval" in the name will make people think about the eval function in JS

I do understand that though. `evaluateAuth` maybe? Along with adding the missing JSDocs.

alflennik: Sure! That would be good.

PR #336, opened 2021-10-18
Fix A11y Issues

alflennik:
Addresses accessibility issues with the app raised in the Community Group.


richnoah:
@alflennik I installed JAWS and was able to verify that upon loading /run/110 and pressing 1 JAWS vocalized "Testing task: 1 Navigate forwards to a grid in reading mode"

Issue #334, opened 2021-10-01
Disclaimer Copy/Icon


richnoah:
In order to make it clear that the reports displayed are only from candidate tests, a disclaimer needs to be included on the reports page.  It is also desired to include iconography and messaging for each data set.

richnoah:
Per Matt:

place disclaimer after first h2 and third h2 along with each h2 (browser/AT combo) on /reports.

place disclaimer after first h2 and for each h2 for each test on /reports/#

Issue #333, opened 2021-10-01
Public view only of test-queue


richnoah:
In an effort to share the work that has been done on ARIA-AT App and allow potential members to get more of an insight into the tests it is desired to make the test-queue viewable without being logged in.

Users that are viewing should not have to assign themselves to view a test plan.

Issue #332, opened 2021-10-01
Public view only of test-run with messaging


richnoah:
In an effort to share the work that has been done on ARIA-AT  App and allow potential members to get more of an insight into the tests it is desired to make the test-run viewable without being logged in.

Users that are viewing should not be able to save any data and need to be informed that their data will not be saved.

PR #337, opened 2021-10-19
Public pages & disclaimer info

howard-e:
* Allows public access to the Test Queue and Test Run pages.
* Now displays a disclaimer on relevant Test Reports related pages to inform a viewer that the reports displayed are only from candidate tests.

Addresses #332, #333 & #334.

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -553,13 +553,13 @@ const TestRun = () => {
                         />
                     </li>
                     <li>
-                    <OptionButton
-                        text="Start Over"
-                        icon={<FontAwesomeIcon icon={faRedo} />}
-                        onClick={handleStartOverButtonClick}
-                        disabled={!isSignedIn}
-                    />
-</li>
+                        <OptionButton
+                            text="Start Over"
+                            icon={<FontAwesomeIcon icon={faRedo} />}
+                            onClick={handleStartOverButtonClick}
+                            disabled={!isSignedIn}
+                        />
+                    </li>
                     <li>
                         <OptionButton
                             text={!isSignedIn ? 'Close' : 'Save and Close'}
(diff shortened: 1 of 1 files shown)


client/components/Reports/SummarizeTestPlanVersion.jsx
```diff
@@ -77,6 +79,24 @@ const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
...
+                                    tests are in review by assistive technology
+                                    developers and lack consensus regarding:
+                                    <ol>
+                                        <li>
+                                            applicability and validity of the
+                                            tests, and
+                                        </li>
+                                        <li>accuracy of test results.</li>
+                                    </ol>
+                                </>
```
alflennik, line 97: Are these props needed given they are the defaults?

howard-e: No, it wouldn't be needed here as well. Missed that one.

client/components/Reports/SummarizeTestPlanVersion.jsx
```diff
@@ -37,6 +38,7 @@ const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
                 </Breadcrumb.Item>
             </Breadcrumb>
             <h2>Introduction</h2>
+            <DisclaimerInfo />
```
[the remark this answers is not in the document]
howard-e: Yes, this was what was requested by the CG. See relevant comment: https://github.com/w3c/aria-at-app/issues/334#issuecomment-943670554

alflennik (commented):
Great job with the disclaimers and the especially with the public test queue. I know we didn't have a huge amount of guidance, but I think you made all the right judgment calls in bringing it to life. This actually will produce a dramatic difference in communicating what is happening inside the app, all of a sudden a huge amount of content is suddenly viewable by the public. That's very exciting!

PR #352, opened 2021-11-16
Command Sequences

alflennik:
~This cannot be merged until Z and I fix an issue manifesting in the Test Renderer.~ The issue is fixed!

Changes the GraphQL schema to support command sequences, migrates the existing data to match the change, reenables the import script to import tests with the updated structure, and updates the existing code and tests to handle an array where there was previously an object. This also fixes the build errors which were caused by the disabled import script.


client/resources/aria-at-test-run.mjs
```diff
@@ -139,7 +139,7 @@ export function instructionDocument(resultState, hooks) {
             },
             `.`,
           ],
-          `Activate the "Open test page" button below, which opens the example to test in a new window${setupScriptDescription || '.'}`,
+          `Activate the "Open test page" button below, which opens the example to test in a new window${setupScriptDescription}`,
```
howard-e, line 142: May need to verify if the issue where the string `"undefined"` was being injected into instructions text for tests without setup scripts, is now gone.

server/migrations/20211116172218-commandSequences.js
```diff
@@ -0,0 +1,21 @@
+const { omit } = require('lodash');
+const { TestPlanVersion } = require('../models');
+
+module.exports = {
+    up: async (/* queryInterface, Sequelize */) => {
```
howard-e, line 5: Could remove the parameters here

server/scripts/import-tests/index.js
```diff
@@ -233,6 +228,9 @@ const getTests = ({ builtDirectoryPath, testPlanVersionId, ats, gitSha }) => {
     const renderedUrlsByNumber = {};
     const allCollectedByNumber = {};
     fse.readdirSync(builtDirectoryPath).forEach(filePath => {
+        if (filePath.includes('breadcrumb')) {
```
howard-e, line 231: Assuming this block is just temporary?

alflennik: Lol whoops

alflennik: Fixed

server/migrations/20211116172218-commandSequences.js
alflennik: Okay, good idea. Fixed.

client/resources/aria-at-test-run.mjs
alflennik: I still see it! I've deployed this branch to the sandbox, and if you sign into the sandbox and then go to this page you should see it: https://aria-at-app-sandbox.bocoup.com/run/120?user=5

alflennik: After clearing my cache I was able to verify that the issue is fixed! Since I've deployed this branch to the Sandbox, you can see it here (must be logged-in to the sandbox for the link to work) https://aria-at-app-sandbox.bocoup.com/run/120?user=5

alflennik: Actually, this is still occurring.

alflennik: This is now genuinely fixed (by https://github.com/w3c/aria-at/pull/576)

alflennik:
@howard-e I've removed the WIP label! Take a second look? The main change is that the migration is quite a bit more involved. On my first migration attempt I missed a case that was causing some corrupt data.

I've deployed the latest version of the branch to the [sandbox](https://aria-at-app-sandbox.bocoup.com/) for testing.

jscholes:
`Activate the "Open test page" button below, which opens the example to test in a new window${setupScriptDescription}`,

I don't understand this wording.  If the setup script description is taken from the tests as currently written, the sentence will be badly worded and not make much sense.  E.g.:

> Activate the "Open test page" button below, which opens the example to test in a new windowsets focus on a link before the button

What is the setup script description text in this case?

alflennik:
Hello James! If a `setupScriptDescription` exists for the test, the instruction text line is formatted as _Activate the “Open test page” button below, which opens the example to test in a new window **and runs a script that** sets focus on a link before the button._

The source code which formats the string can be seen at https://github.com/w3c/aria-at/blob/66babf3af40ae9b4c7580c892a5503fe1b009f9c/tests/resources/aria-at-test-run.mjs#L96-L98.

howard-e (approved):
Looks good!

I've left some very minor feedback.

server/resolvers/Test/renderableContentResolver.js
```diff
@@ -3,7 +3,8 @@ const { UserInputError } = require('apollo-server');
 const renderableContent = (test, { atId }) => {
     if (!atId && !test.inferredAtId) {
         throw new UserInputError(
-            'atId was not provided and could not be inferred'
+            'although the atId is optional when it can be inferred from ' +
```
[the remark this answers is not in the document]
alflennik: I get what you're saying, to literally ask the user to provide the atId, I can put that part back in.

PR #353, opened 2021-11-17
Increase test page size

alflennik:
Copies over the change from https://github.com/w3c/aria-at/pull/577. See that PR for more information!


mzgoddard (approved):
Looks good to me.

Issue #350, opened 2021-11-15
Improvements to test plan conflict modal


s3ththompson:
From [11/11 CG meeting](https://www.w3.org/2021/11/11-aria-at-minutes.html):

> - Conflict resolution modal should show only conflicts for current test
> - Conflict resolution should clearly label test and assertion
> - Raise an issue button should be per-test
> - Conflicts should include GitHub usernames to track down origin of inconsistencies
> - Inconsistent language ("N/A" vs "No unexpected behaviors")
> - Add top-level conflict resolution summary from Test Queue
> - Add link to original test to view assertion in context

richnoah:
Conflict resolution modal should show only conflicts for current test
Inconsistent language ("N/A" vs "No unexpected behaviors")
Modal Dialog Label "Reviewing Conflicts for [Test Name]"
Include github id as user
Update Your result: N/A to "No"
Update Your to github id in the case of Your result

s3ththompson:
Per CG meeting 11/18:
- If an assertion has no output, `FAILED: No Output (for output "no output")` is confusing since it makes it sound like "no output" *was* the output, and should just read `FAILED: No Output`

PR #355, opened 2021-11-18
Improve Raise an Issue

alflennik:
As of Nov 18, 2021, this branch has been deployed to the [Sandbox](https://aria-at-app-sandbox.bocoup.com/). While testing, be sure to look in both the test queue and the reports page, and look at tests in the test queue with and without conflicts.

This PR doesn't include improvements to the way conflicts are formatted - that will be coming in another PR soon, and I'll make sure those improvements make their way into the issue template.

### Example title: 
Feedback: "Open a Modal Dialog" (Modal Dialog Example, Test 3)

I did capitalize the word "Test" and add a comma between the test plan name and the Test 3 part. That resolves the feeling I had that the word "test", even though it was lowercase, felt like part of the test plan name. I think the new title feels really clean!

#### Test Setup

- Test File at Exact Commit: [test-03-open-modal-dialog-interaction-voiceover_macos.collected.html](https://aria-at.w3.org/aria-at/66babf3af40ae9b4c7580c892a5503fe1b009f9c/build/tests/modal-dialog/test-03-open-modal-dialog-interaction-voiceover_macos.collected.html)
- AT: VoiceOver for macOS (version 11.5.2)
- Browser: Safari (version 14.1)

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -33,28 +33,30 @@ import './TestRun.css';
 const createGitHubIssueWithTitleAndBody = ({
     test,
     testPlanReport,
-    conflictsFormatted,
-    isReportViewer = false
+    conflictsFormatted
 }) => {
-    const issueRaiser = isReportViewer ? 'Report Viewer' : 'Tester';
-    const title = `${issueRaiser}'s Issue Report for "${test.title}"`;
-
-    const { testPlanTarget } = testPlanReport;
+    const { testPlanVersion, testPlanTarget } = testPlanReport;
     const { at, browser } = testPlanTarget;
+
+    const title =
+        `Feedback: "${test.title}" (${testPlanVersion.title}, ` +
+        `Test ${test.rowNumber})`;
+
     const shortenedUrl = test.renderedUrl.match(/[^/]+$/)[0];
 
     let body =
-        `### Test File at Exact Commit\n\n` +
-        `[${shortenedUrl}](https://aria-at.w3.org${test.renderedUrl})\n\n` +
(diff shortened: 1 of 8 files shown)


client/components/TestRun/index.jsx
```diff
@@ -33,28 +33,30 @@ import './TestRun.css';
...
+        `<!-- write your description here -->\n\n` +
+        `#### Test Setup\n\n` +
+        `- Test File at Exact Commit: ` +
+        `[${shortenedUrl}](https://aria-at.w3.org${test.renderedUrl})\n` +
+        `- AT: ` +
+        `${at.name} (version ${testPlanTarget.atVersion})\n` +
+        `- Browser: ` +
+        `${browser.name} (version ${testPlanTarget.browserVersion})\n`;
 
     if (conflictsFormatted) {
```
howard-e, line 58: It appears that once a test exists in the suite with conflicts, it is also included in the Issue's output regardless of whether or not the current test which an issue is being raised for has any conflicts.

howard-e (changes requested):

client/components/TestRun/index.jsx
howard-e: Based on the 11/18 CG meeting, it seems that this is already known based on https://github.com/w3c/aria-at-app/issues/350. Will change my review to approval as that issue isn't related to this PR.

howard-e (approved):

client/components/TestRun/index.jsx
alflennik: Thanks for mentioning that! I'll definitely be fixing this in the conflicts improvement work.

PR #356, opened 2021-11-22
Improve Conflict Presentation 

alflennik:
Using the issue https://github.com/w3c/aria-at-app/issues/350 as guidance, this improves the presentation of conflicts and should make resolving conflicts much easier.

As of Nov 22, this branch has been deployed to the [Sandbox environment](https://aria-at-app-sandbox.bocoup.com/).

See below for an example of the new conflict summary. Some changes I made are:

- Made the conflict text much more structured in HTML, for example using an ordered list instead of a series of numbered headlines. Again, the full markup is below if you want to check it out.
- Made the text more semantic and human-readable.
- Made sure the same data is presented in the same order in all cases.
- Fixed a vulnerability in the previous version to XSS.

We discussed removing the output text in the case of no output. I think this is a good idea but it's better to address it holistically instead of only within this feature. For example, the Reports page and Test Run page will faithfully show the tester's recorded output, so I don't want it to be mysteriously missing here.

For testing, the following areas of the app are affected:

- The "Raise an Issue" button found in the detailed results part of the Reports pages.
- The "Raise an Issue" button on the test run page.
- The "Raise an Issue" button in the conflicts modal
- The content presented inside the conflicts modal.

## Review Conflicts for "Open a collapsed select-only combobox in reading mode"

1.  ### Unexpected Behaviors for "Alt+Down" Command
    
    *   Tester esmeralda-baggins recorded output "automatically seeded sample output" and noted no unexpected behavior.
    *   Tester tom-proudfeet recorded output "automatically seeded sample output" and noted "Output is excessively verbose, e.g., includes redundant and/or irrelevant speech" and "Seeded other unexpected behavior".
2.  ### Assertion Results for "Alt+Down" Command and "State of the combobox (expanded) is conveyed" Assertion
    
    *   Tester esmeralda-baggins recorded output "automatically seeded sample output" and marked assertion as failing with no output.
    *   Tester tom-proudfeet recorded output "automatically seeded sample output" and marked assertion as failing with incorrect output.


howard-e (approved):
LGTM

PR #366, opened 2021-11-26
Describe and Waypoint Main Menu via Signpost Heading.

jkva:

> The navigation bar items ("Test Reports", " Test Queue", etc.) are not marked up inside an unordered list to convey their grouped nature to screen reader users. They are also not preceded by a heading, either visible or off-screen/screen-reader-only.

The "The navigation bar items ("Test Reports", " Test Queue", etc.) are not marked up inside an unordered list to convey their grouped nature to screen reader users" has previously been addressed.

---

Effective changes:

- A visually hidden `h2` element containing "Main Menu" was added to precede the main menu `nav`; and
- the main menu `nav` is described by this `h2`.

### DIFF
--- a/client/components/App/App.jsx
+++ b/client/components/App/App.jsx
@@ -36,10 +36,13 @@ const App = () => {
     return (
         <ScrollFixer>
             <Container fluid>
+                <h2 id="main-menu-heading" className="sr-only">
+                    Main Menu
+                </h2>
                 <Navbar
                     bg="light"
                     expand="lg"
-                    aria-label="Main Menu"
+                    aria-labelledby="main-menu-heading"
                     expanded={isNavbarExpanded}
                     onToggle={() => setIsNavbarExpanded(previous => !previous)}
                 >


jscholes (changes requested):
@jkva Looks good.  A few things:

1. Can we change "Main Menu" to "Main Navigation"?
2. Exchange `aria-labelledby` on the `nav` for `aria-label="Main"`, to avoid duplicate "navigation" announcements.
3. Move the `<h2>` inside the `nav`.
4. If not already done, add `aria-labelledby` to the actual list wrapping the nav links, with `aria-labelledby` referencing the new heading.

isaacdurazo:
@jkva I noticed a couple of things here about the UI:

1. The Home link now reads "HOME" instead of "ARIA-AT"
2. When any link in the main nav is clicked, the "Skip to main content" button gets displayed. Is this intentional?

jkva:
> @jkva I noticed a couple of things here about the UI:
> 
> 1. The Home link now reads "HOME" instead of "ARIA-AT"
> 2. When any link in the main nav is clicked, the "Skip to main content" button gets displayed. Is this intentional?

Hmm, neither of these things are proper. I've locally rebased this PR on #367, which overlaps with this (#366) PR in the sense that it also addresses navigation and styling.

Once #367 gets merged I'll rebase #366 on `main` and force push to this branch, that should put things in their proper place. 

cc @alflennik @jscholes

PR #370, opened 2021-12-06
Ensure Each TestQueueRow Uses a Unique ID-ref.

jkva:
This change ensures each id generated for a Test Queue Row combines its test plan ID and tester name. This addresses a problem where two test rows assigned to the same tester generate (elements with) duplicate IDs.

### DIFF
--- a/client/components/TestQueueRow/index.jsx
+++ b/client/components/TestQueueRow/index.jsx
@@ -337,6 +337,15 @@ const TestQueueRow = ({
     const { status, results } = evaluateStatusAndResults();
     const nextReportStatus = evaluateNewReportStatus();
 
+    const getRowId = tester =>
+        [
+            'plan',
+            testPlanReport.id,
+            'assignee',
+            tester.username,
+            'completed'
+        ].join('-');
+
     return (
         <tr className="test-queue-run-row">
             <th>{renderAssignedUserToTestPlan()}</th>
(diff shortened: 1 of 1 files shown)


jkva:
@jscholes previously, you would get multiple elements with id `assignee-[username]-completed`, so in the case of multiple plans with the same assigned user, this would result in duplicate ids on the page.

jscholes:
@jkva I looked at the rendered mark-up in the actual app as it is currently deployed, and this makes more sense now.  The use of the word "row" is what was tripping me up.  You're talking about a `div`, representing a tester's completion status, inside a list item, in an `ul`.  But that content itself is inside a table.  So when you said "row", I thought you meant a row of the table.

However, now that I understand it, I want to confirm that the proposed change fully resolves the issue.  The Test Queue will contain multiple instances of the same plan, across different browser/AT combinations.  If the same tester is assigned to the same plan across multiple combos, will the IDs still be unique?

jkva:
That's a good use case for me to double-check, thank you.

jkva:
@jscholes I've pushed a commit including the Test Plan Run ID within the generated ID, further guaranteeing uniqueness.

jscholes (approved):

howard-e (approved):
LGTM other than the optional change request.

client/components/TestQueueRow/index.jsx
```diff
@@ -337,6 +337,17 @@ const TestQueueRow = ({
...
 
+    const getRowId = tester =>
+        [
+            'plan',
+            testPlanReport.id,
+            'run',
+            currentUserTestPlanRun.id,
+            'assignee',
+            tester.username,
+            'completed'
```
howard-e, line 348: Optional nitpick to address: I understand the id's original structure was being preserved with `-completed` being appended. Taking another look again at its use, perhaps something like `-completed-tests` is clearer for the id's purpose, but removing `-completed` altogether also works well and is helpful to reduce some of id's verbosity.

evmiguel (approved):

PR #372, opened 2021-12-06
Use routable locations when navigating through testplan tasks.

jkva:


client/components/TestRun/index.jsx
```diff
@@ -814,6 +835,7 @@ const TestRun = () => {
                 <TestNavigator
                     show={showTestNavigator}
                     tests={tests}
+                    testPlanRunId={parseInt(testPlanRunId)}
```
jkva, line 838: I'm not a fan of this explicit int typecasting; why is this not addressed by the routing regular expression?

jscholes (approved):

PR #376, opened 2021-12-08
Remove Duplicate Level One Heading

jkva:

> There are two level 1 headings for each testing task, making it potentially unclear where the main content of interest begins. There should only ever be one level 1 heading on a page.

--- 

**Question**: since the `h1` and `h2` on this page are quite similar:

- `h1` Testing task: 1. Open a Modal Dialog
- `h2` Testing task: Open a Modal Dialog

Could the `h2` not be done without? That way this PR can be simplified by just removing the duplicate `h1` altogether and thus keeping the rest of the heading structure intact. This would then also ensure no changes need to be made to #375.

---

Effective changes:

- Move second h1 (Testing task title) to h2; and
- any h2 to h3, h3 to h4.

### DIFF
--- a/client/components/TestRenderer/index.jsx
+++ b/client/components/TestRenderer/index.jsx
@@ -29,11 +29,13 @@ const ErrorSection = styled.section`
 
 const InstructionsSection = styled.section``;
 
-const HeadingText = styled.h1``;
+const HeadingText = styled.h2`
+    margin-top: 0;
+`;
 
-const SubHeadingText = styled.h2``;
+const SubHeadingText = styled.h3``;
 
-const InnerSectionHeadingText = styled.h3``;
+const InnerSectionHeadingText = styled.h4``;
 
 const Text = styled.p`
     > textarea {


jscholes (changes requested):
@jkva I agree with your assertions here, but I think restructuring the headings completely is a larger task.  E.g. I'd personally have the test plan name, AT/browser info and number of completed tests earlier on the page, maybe before the test navigator or in that section?  Not sure.  But either way it would also require some design changes.

For now, does this also address the formatting of the remaining `h1`, whereby the text is missing whitespace after the colon for screen reader users?

jkva:
@jscholes I've added a non breaking space between the "Testing task:" span and the rest of the h1 content. There's otherwise not really anything to do there apart from setting the `aria-label` on the `h1` which seems overkill.

As for the heading restructuring, the PR in its current state already manages that. It just results in two headings (h1 and h2) being quite similar in text content.

isaacdurazo:
@jkva I noticed that adding the non-breaking-space creates an alignment issue in the UI, which pushes the "Testing task" name slightly to the right.

Before:
Screen Shot 2022-02-03 at 2 56 59 PM

After:
Screen Shot 2022-02-03 at 2 57 03 PM

Would it be possible and semantically correct to have the non-breaking-space right after the colon but within the `<span>` element?

jkva:
@isaacdurazo good point, thank you. Addressed.

isaacdurazo:
Thanks for addressing, @jkva

PR #380, opened 2021-12-09
Implement Test Navigator Toggle button as Disclosure Widget

jkva:

> Directly after the "Test Navigator" heading, there is a button which reads: "Toggle button to close test navigator". This label is overly verbose, and points to the fact that a Disclosure pattern is not being used (e.g. as demonstrated by the APG's example of disclosures for a navigation menu). This control would be far more usable as a button with aria-expanded, collapsed down into the previous heading. That is, the heading should wrap a button inside it, labelled "Test Navigator", which programmatically indicates its expanded/collapsed state via aria-expanded. As an aside, that will also give the button a visible label because the heading is on-screen; right now the button only has an icon.

@jscholes some duplication exists in #362. The `h2` around the `button` is now used to describe the `ol` inside the `nav`, but is not itself a child of the `nav` since it is its toggle control. This does not follow the usual pattern of placing the heading in the nav ahead of the nav list, but another specific heading seems overkill. Thoughts?

--- 

Effective changes:

- Wrap a `h2` around toggle `button`;
- set a static `aria-label` onto the `button`;
- manage `aria-expanded` attribute on the `button`;
- associate `button` with `nav` using `aria-controls` idref; and
- describe navigation list using `h2`.

### DIFF
--- a/client/components/TestRun/TestNavigator.jsx
+++ b/client/components/TestRun/TestNavigator.jsx
@@ -41,64 +41,61 @@ const TestNavigator = ({
                     </button>
                 </h2>
             </div>
-            {show && (
-                <nav id="test-navigator-nav">
-                    <ol
-                        aria-labelledby="test-navigator-heading"
-                        className="test-navigator-list"
-                    >
-                        {tests.map(test => {
-                            let resultClassName = 'not-started';
-                            let resultStatus = 'Not Started';
+            <nav id="test-navigator-nav" hidden={!show}>
+                <ol
+                    aria-labelledby="test-navigator-heading"
+                    className="test-navigator-list"
+                >
+                    {tests.map(test => {
+                        let resultClassName = 'not-started';
+                        let resultStatus = 'Not Started';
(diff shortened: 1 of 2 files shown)


jscholes (changes requested):
@jkva I agree with pretty much all of this; it's a nice compromise between discoverability and polluting the page (e.g. with a nav region that only contains a collapsed button by default).

Only one thing: `aria-label="Test Navigator"` should really be `aria-label="Test Navigation"`, so that it fits with the "Test navigation" that users will hear when they expand it and access the labelled `nav`.

jkva:
@jscholes thanks for catching that. I found some redundant class and CSS usage - removed that to simplify while I was at it.

jscholes (approved):

alflennik:
Once the conflicts are resolved, we feel ready to merge this!

jkva:
@alflennik The PR has been rebased onto `main` and the conflict has been resolved.

PR #397, opened 2022-01-12
Adding date and time to test queue and new test plan dropdown

evmiguel:
In this PR, I did not include weeks or months as a unit of time. The only units I included were days and hours, mostly because when I started to add weeks and months, the calculations were not precise enough. In this implementation, if the number of days passes 30 days, the date as a string will be used instead; this approach is just like GitHub. [See how React commits look like](https://github.com/facebook/react/commits/main)

This PR adds the days and git SHAs to the Test Plan rows. It also adds the days and git SHAs to the version picker in the Add Test Plan picker.

Screen Shot 2022-01-12 at 12 37 00 PM

Screen Shot 2022-01-12 at 12 37 42 PM


client/utils/dateInterval.js
```diff
@@ -0,0 +1,30 @@
...
+
+    if (timeIntervalInDays >= 30) {
+        const dateString = `${startDate.toLocaleString('default', {
+            month: 'short'
+        })} ${startDate.getDate()}, ${startDate.getFullYear()}`;
+        return dateString;
+    }
+
+    return `${intervalNumber} ${intervalUnit} ago`;
+};
```
alflennik, line 30: Nicely done!

client/utils/dateInterval.test.js
```diff
@@ -0,0 +1,38 @@
...
+        expect(dateIntervalString).toBe('3 days ago');
+    });
+
+    it('returns month, day, and year of start date', () => {
+        const startDate = new Date(2022, 0, 1);
+        const endDate = new Date(2022, 1, 2);
+        const dateIntervalString = calculateDateInterval(startDate, endDate);
+        expect(dateIntervalString).toBe('Jan 1, 2022');
+    });
+});
```
alflennik, line 38: Great job with the tests too 👍

client/utils/dateInterval.js
s3ththompson: agreed with @alflennik that this is really slick! However, Matt prefers that we use absolute dates instead of relative times, always (I mentioned that GitHub uses this approach, but he felt consistency and no relative times was clearer). Can we change the output to always be the formatted date string?

Also, can we add the time in order to disambiguate between two commits on the same day?

What do you think of something like this?
 
```js
`${startDate.toLocaleString('default', { month: 'short' })} ${startDate.getDate()}, ${startDate.getFullYear()} ${startDate.toLocaleTimeString();}`
```

client/components/TestQueueRow/index.jsx
```diff
@@ -114,25 +115,44 @@ const TestQueueRow = ({
     };
 
     const renderAssignedUserToTestPlan = () => {
+        const gitShaDateDifference = (
```
s3ththompson, line 118: Matt felt strongly that the Test Queue rows should just show the date, not the Git SHA... Can we remove the Git SHA here, (but keep it where it appears in the NewTestPlanReportModal picker)?

s3ththompson (changes requested):
@evmiguel this PR is awesome!

I received some feedback from Matt in a 1:1 last night that changes the requirements slightly. Sorry for the churn!

client/components/TestQueueRow/index.jsx
richnoah: If it is just the date displayed then what would be a delineation if there are two tests published the same day?

s3ththompson: Yep, had the same thought--see sibling comment about adding time to date...

alflennik:
Test queue table with the word revision under the test plan

Just to throw more variables in to the mix, I think including the word "revised" will help both sighted and unsighted users understand the contextual meaning of the date.

We could put `(commit d1abe0)` in parentheses following the text, if we decide it is needed.

Visually, I think we should bump up the font size of the test plan link to `1.1rem` and lower the font size of the revision text to `0.9rem`, and make it italic, as pictured.

richnoah:
@alflennik would it still be revised if it was an initial test?  Maybe the use of Published would fit all situations.

alflennik:
Test queue test plan showing the word published and a more exact date

@richnoah here's what that looks like. I also threw in a more specific date assuming that's what we go with.

client/components/NewTestPlanReport/NewTestPlanReportModal.jsx
```diff
@@ -176,7 +176,11 @@ const NewTestPlanReportModal = ({
                                 key={`${item.gitSha}-${item.id}`}
                                 value={item.id}
                             >
-                                {item.gitSha}
+                                {calculateDateInterval(
+                                    new Date(item.updatedAt),
+                                    new Date()
+                                )}{' '}
+                                ({item.gitSha.substring(0, 5)})
```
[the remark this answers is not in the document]
alflennik: Assuming we keep the shas at all. Seems like we've decided to drop them.

server/resources/commands.json
```diff
@@ -419,6 +475,10 @@
         "id": "TAB_AND_SHIFT_TAB",
         "text": "Tab / Shift+Tab"
     },
+    {
+        "id": "TWO",
+        "text": "2"
+    },
     {
         "id": "T_THEN_DOWN",
         "text": "T followed by Down Arrow"
```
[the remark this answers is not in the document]
evmiguel: I took this out. Sneaky file!

client/components/TestQueueRow/index.jsx
isaacdurazo: + 1 on adding time in addition to absolute date format. SHAs don't really mean anything whereas time solves the delineation issue brought up by Rich and does mean something for the user.

isaacdurazo:
I like the use of the label "Published". should we consider other labels like "Last modified"?

client/components/NewTestPlanReport/NewTestPlanReportModal.jsx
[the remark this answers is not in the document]
evmiguel: @s3ththompson are we keeping the SHAs?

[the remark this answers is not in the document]
s3ththompson: Sorry for not being clearer... Matt's feedback was to remove SHAs from the TestQueueRows, but keep them in the NewTestPlanReportModal:

- TestQueueRows:
    - `Nov 18, 2021 at 1:31pm`
- NewTestPlanReportModal:
    - `Nov 18, 2021 at 1:31pm (sha1234)` or
    - `Nov 18, 2021 at 1:31pm "Create tests for APG design pattern example: Main Landmark" (sha1234)`
   
Matt expressed that if we could add commit messages to the modal that would be even better, but I think we should do whatever is easier here since this is a temporary fix.

evmiguel:
@s3ththompson @alflennik @richnoah @isaacdurazo - Here's what I did:

- [x] Absolute date in `TestQueueRow`
- [x] Absolute date, git commit message, and git SHA in `NewTestPlanReportModal`
- [x] Styling suggested by @alflennik
- [ ] Use of "Published" or "Revised"

s3ththompson (approved):

client/utils/gitUtils.js
```diff
@@ -0,0 +1,9 @@
+export const gitUpdatedDateToString = dateString => {
+    const date = new Date(dateString);
+    return `${date.toLocaleString('default', {
+        month: 'short'
+    })} ${date.getDate()}, ${date.getFullYear()} at ${date.toLocaleTimeString(
+        'default',
+        { timeZone: 'UTC' }
+    )}`;
```
alflennik, line 8: Is it possible to make the AM / PM lowercase? If there's no good way to do it, i.e. you have to use regex replace or something, then I guess I'd be happy with the uppercase version, but I do think the lowercase version seems more friendly.

alflennik (approved):
I did make one note about the capitalization, but I think it's optional. This looks great, and good job rolling with the punches.

alflennik:
Confirmed the PM is now pm, nice.

PR #398, opened 2022-01-13
Revert "Add Test Plan to Test Queue modal dialog accessibility modifications"

alflennik:
Reverts w3c/aria-at-app#361

We discovered that this PR had some unintended side effects. We were unable to add new test plans to the queue - when clicking the add button there was a 400 error. So we will revert the merge!


evmiguel (approved):
Changes look reverted.

Issue #360, opened 2021-11-24
Meaningless Test Plan Version information within Add New Test Plan Dialog


jkva:
When opening the "Add New Test Plan to the Test Queue" dialog, the "Test Plan Version" select element's options are hashes.

![Screenshot of the "Add New Test Plan to the Test Queue" dialog, showing the "Test Plan Version" select element reading "66c6b64294b0b5ac3b6"](https://user-images.githubusercontent.com/217264/143233455-d545486f-19d5-4592-93d8-15c22c6609d3.png)

Issue #359, opened 2021-11-24
Unclear unavailable affordance with in Add New Test Plan Dialog


jkva:
When opening the "Add New Test Plan to the Test Queue" dialog, the "Add" button is initially disabled.

![Screenshot of the "Add New Test Plan to the Test Queue" dialog, showing the "Add" button grayed out](https://user-images.githubusercontent.com/217264/143233455-d545486f-19d5-4592-93d8-15c22c6609d3.png)

The reason for this is that each input field within this dialog has mandatory input requirements; hence the "Add" button becomes available once all input requirements have been satisfied.

From a UX and cognitive accessibility perspective, it is preferred that a data-entry interface notifies the user of its requirements. Hence the "Add" button should be available at all times, and the interface should potentially display one or more requirements that were not met when this button is used. Besides this, required text-input fields (AT Version, Browser Version) should be indicated as requiring mandatory input.

Besides this, the "disabled" styling of such affordances makes them difficult to see.

## Proposed modifications:
Since this is not a classic form with submit element, the `required` attribute on the text input field isn't useful. Hence:

- Indicate each text input field as required via `aria-required`;
- add a "required" signifier to each required input's label, ignored by AT via `aria-hidden`;
- enable the "Add" button by default; and
- when the "Add" button is clicked and not all input requirements are unsatisfied, mark these input fields as having insufficient input as such, and move focus to a rendered list of input failures, each item in the list as a link which places focus on the associated input field.

PR #399, opened 2022-01-13
New test plan modal accessibility fixes

alflennik:
Note: this PR was originally authored by @jkva here: https://github.com/w3c/aria-at-app/pull/361. After merging we got a 400 error when adding a new test plan to the queue, so we decided to revert the merge and continue reviewing this PR here.

The original description is below.

--

This PR addresses the following items from PAC's "ARIA-AT App Screen Reader Accessibility Observations" document where applicable to the "Add New Test Plan" modal dialog:

> There is no level 1 title heading at the start of the modal content, nor an accessible name on the modal dialog container.

A level 1 title heading now exists in the modal dialog header, and it describes the dialog document via an `aria-labelledby` IDREF.

> When the modal is opened, keyboard focus is incorrectly placed on the entire dialog container rather than an appropriate element inside it.

Keyboard focus is now placed on the first interactive input field within the dialog body.

> When the dialog is closed, keyboard focus is temporarily lost before being restored to the button that was used to open it. This causes superfluous and verbose screen reader feedback.

Keyboard focus is now explicitly placed on the dialog's trigger once the dialog is closed.

**Question**: the [React Bootstrap Modal Dialog API documentation](https://react-bootstrap.github.io/components/modal/#modal-props) mentions a `restoreFocus` option that should cover this functionality and is enabled by default; is it obvious why this is currently not working and a workaround was needed?

> There are no labels for the form fields.

Each form field has been given an accessible label which is associated via `for=` IDREF.

---

Besides this:

- Each category of input (AT, Browser, Test Plan) has been grouped within a `fieldset`, each `fieldset`'s `legend` containing a `h2` describing the `fieldset` context; and
- each `select` element automatically selects its first option given that none of these values are mandatory.

---

Issues created out of work on this PR:

* #359 
* #360

---
 
A screenshot showing the modified modal dialog

### DIFF
--- a/client/components/NewTestPlanReport/NewTestPlanReportContainer.jsx
+++ b/client/components/NewTestPlanReport/NewTestPlanReportContainer.jsx
@@ -4,10 +4,14 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faPlus } from '@fortawesome/free-solid-svg-icons';
 import PropTypes from 'prop-types';
 
-const NewTestPlanReportContainer = ({ handleOpenDialog = () => {} }) => {
+const NewTestPlanReportContainer = ({
+    handleOpenDialog = () => {},
+    openDialogTriggerId
+}) => {
     return (
         <div className="add-test-plan-queue-container">
             <Button
+                id={openDialogTriggerId}
                 className="add-test-plan-queue-button"
                 variant="primary"
                 onClick={handleOpenDialog}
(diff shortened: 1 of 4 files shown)


jkva:
@alflennik thank you for catching this. On review I spotted a payload bug my changes had introduced, commit 30aad53 is meant to address this.

evmiguel:
Hi @jkva, can you resolve the conflicts? Will take a second look after.

jkva:
@evmiguel I've rebased on `origin/main` and force-pushed the branch. While this PR now includes the git timestamp code changes, I've found an edge case in its test coverage, a PR exists at #400.

evmiguel (approved):

PR #404, opened 2022-03-15
Test plan updater

evmiguel:
This is the PR for the Test Plan Updater Modal. To see it locally,

- Start development server at localhost:3000: `yarn dev`
- Log into the app as an admin
- Navigate to the Test Queue
- Pick a report and press  the `Update Test Plan Version` to see the modal
- Try creating a new test plan report

----
Open questions:

- Which versions should be shown in the dropdown?
- What should happen when the admin does not choose to delete the old report?

cc: @richnoah @s3ththompson


client/components/TestPlanUpdater/TestPlanUpdater.jsx
```diff
@@ -0,0 +1,436 @@
...
+        setIsOldReportDeleted(true);
+    };
+
+    return (
+        <Container id="main" as="main" tabIndex="-1">
+            <Helmet>
+                <title>Test Plan Updater | ARIA-AT</title>
+            </Helmet>
+            <h1>Test Plan Updater</h1>
+            <h2>Explanation</h2>
```
alflennik, line 306: Shall we remove this UI which is no longer used?

client/routes/index.js
```diff
@@ -56,6 +57,14 @@ export default [
...
         component: Reports
     },
+    {
+        path: '/test-plan-updater',
+        render: () => (
+            <ConfirmAuth requiredPermission="ADMIN">
+                <Route component={TestPlanUpdater} />
+            </ConfirmAuth>
+        )
+    },
```
alflennik, line 67: This might not be needed anymore, right?

client/index.js
```diff
@@ -8,7 +8,37 @@ import App from './components/App';
...
+                        }
+                    },
+                    testPlanRun: {
+                        merge(existing, incoming) {
+                            return incoming;
+                        }
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
evmiguel: Works!

client/routes/index.js
evmiguel: Right. I've removed it.

richnoah:
@s3ththompson @evmiguel I don't see the need to downgrade as such a high priority.  As explained there are bigger implications of reverting a test plan version.  I would be fine with just the most recent version being shown.  If I have selected the most current version and then go back into the updater is there a check if the selected version is the most current?

@evmiguel I would say it is bad form to have a pop-up above a modal.  Possibly using a toast message that an admin would have to confirm

s3ththompson:
> If I have selected the most current version and then go back into the updater is there a check if the selected version is the most current?

If I understand correctly @richnoah, I think this is a great approach. So instead of a dropdown, the test plan updater only shows the most recent test version. If the test plan is already the recent version, there would be no "Update Test Plan" button in the test queue. Would that be feasible @evmiguel?

> I would say it is bad form to have a pop-up above a modal.

@richnoah I don't know if we have that kind of toast component in our component library. Perhaps it's enough to have a warning and to make the button red?

evmiguel:
@s3ththompson @richnoah @isaacdurazo:

I've removed the select dropdown to make sure that we can only update on the most recent version. Reports that are already on the recent version do not have the update option.

I added alerts to the modal when the old report is being deleted. It feels a bit awkward to me, but I'd like to get y'alls opinions.

client/index.js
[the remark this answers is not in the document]
alflennik: This is minor, but do you want to put the `{ merge: false }` and `{merge: true }` parts on one line? That'll shorten this code by 8 lines 😈

client/components/TestQueueRow/index.jsx
```diff
@@ -120,6 +127,24 @@ const TestQueueRow = ({
...
+        const updateTestPlanVersionButton =
+            isAdmin && latestTestPlanVersion.length === 0 ? (
+                <Button
+                    className="updater-button"
+                    onClick={() => setShowTestPlanUpdaterModal(true)}
+                    size="sm"
+                    variant="outline-secondary"
+                >
+                    {' '}
+                    Update Test Plan Version
```
alflennik, line 143: Since it's the "Test Plan Updater" now I feel like this button can be called "Update Test Plan" which would also make it more compact. I like the new styling. Good job picking its location, so far everyone's been happy with it.

Also could you remove the `{' '}`?

client/components/TestQueueRow/TestQueueRow.css
```diff
@@ -103,6 +103,10 @@ button.more-actions:active {
     font-size: 1.1rem;
 }
 
+.updater-button {
+    max-width: 30%;
+}
```
alflennik, line 108: At a larger screen size this is solving the problem of making the button a reasonable size instead of stretching across the entire cell - but on smaller screens the button shrinks too fast!

```
.btn.updater-button {
    width: initial;
}
```

This will accomplish the same task and work on smaller screens too.

I had to add the `.btn` part to override another style.

client/components/LoadingSpinner/LoadingSpinner.jsx
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from 'prop-types';
+
+const LoadingSpinner = ({ percentage = null, className = null }) => {
+    return (
+        <div
+            className={
+                className
+                    ? `spinner-container ${className}`
+                    : 'spinner-container'
+            }
```
alflennik, line 12: A nice way to get this on one line would be 

```
className={`spinner-container ${className}`}
```

And change the default className to `className = ''`  on line 5.

client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
```diff
@@ -0,0 +1,541 @@
...
+                                        <>
+                                            All test results can be copied from
+                                            the old report to the new report.
+                                        </>
+                                    );
+                                } else {
+                                    deletionNote = (
+                                        <>
+                                            Note that {testsToDelete.length}{' '}
+                                            tests differ between the old and new
```
alflennik, line 442: Could we just make it "test" in the case that there's just one test which can't be copied?

client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
```diff
@@ -0,0 +1,541 @@
...
+                type="checkbox"
+                onChange={() => {
+                    setDeleteChecked(!deleteChecked);
+                }}
+                checked={deleteChecked}
+            />
+            <Form.Check.Label className="delete-label">
+                <FontAwesomeIcon icon={faTrashAlt} className="trash-icon" />{' '}
+                Delete old Test Plan
+            </Form.Check.Label>
```
alflennik, line 365: Clicking the text didn't check the checkbox. I guess you might need to either put the checkbox inside the label or add an `id` prop to the checkbox and a `for` prop for the label.

alflennik (commented):
I'm really proud of what we've accomplished with this feature. At this point we've left our previous sketchy version so far behind, it's amazing. I think what we have here is really user-friendly, and it's really well implemented too. Great job!

client/components/TestQueueRow/index.jsx
```diff
@@ -120,6 +127,24 @@ const TestQueueRow = ({
...
+                </Button>
+            ) : (
+                ''
```
howard-e, line 146: nitpick and completely optional: I'm not sure if the same happens with hooks as it did with `React.Component` but doing this could create a `React.Text` node I believe.
It definitely has no impact on performance like this of course but the standard for preventing component rendering also seems to be defaulting to `null`.

Could also just avoid the ternary altogether and do something like:
```jsx
const updateTestPlanVersionButton =
            isAdmin && latestTestPlanVersion.length === 0 && (
                <Component ... />
            );
```

client/components/TestQueueRow/index.jsx
```diff
@@ -120,6 +127,24 @@ const TestQueueRow = ({
...
+        const latestTestPlanVersion = latestTestPlanVersions.filter(
+            version => version.latestTestPlanVersion.id === testPlanVersion.id
+        );
+        const updateTestPlanVersionButton =
+            isAdmin && latestTestPlanVersion.length === 0 ? (
+                <Button
+                    className="updater-button"
+                    onClick={() => setShowTestPlanUpdaterModal(true)}
+                    size="sm"
+                    variant="outline-secondary"
```
howard-e, line 140: My initial feeling about this visually was that I didn't expect it could be interacted with because of the variant. Would be curious to know if anyone else had the same impression. But if I'm alone in my thoughts here, then no problem!

howard-e (commented):
Impressive stuff @evmiguel and happy to see this getting in there! 🙂

client/components/TestPlanUpdater/TestPlanUpdaterModal.css
```diff
@@ -0,0 +1,80 @@
...
+.submit-button {
+    width: 100%;
+    justify-self: end;
+}
+
+.submit-buttons-row {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
+    width: 100%;
+}
```
alflennik, line 52: I noticed the button text was wrapping, and I think it should be possible to keep the button text on one line. I think the css can be simplified slightly! And the HTML can be simplified too - you can remove the divs with the classes which are no longer needed.

```css
.side-button-container {
    display: grid;
    grid-template-columns: auto auto;
    grid-gap: 0.5em;
}
/* Can be removed */
/*
.side-button-container.no-backup {
    grid-column: 2/3;
}

.submit-button {
    width: 100%;
    justify-self: end;
}

.submit-buttons-row {
    display: grid;
    grid-template-columns: 1fr 1fr;
    width: 100%;
}
*/
```

client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
```diff
@@ -0,0 +1,555 @@
...
+        if (backupChecked) {
+            setAlertCompletion({
+                message: 'Completed without errors.',
+                success: true,
+                visible: true
+            });
+            setCloseButton(true);
+        } else {
+            setAlertCompletion({
+                message: 'Old test plan will be deleted.',
```
alflennik, line 306: Could we do "The old test plan will be deleted."?

client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
```diff
@@ -0,0 +1,555 @@
...
+                            {alertCompletion.message}
+                        </Alert>
+                    </Modal.Body>
+                </>
+            )}
+            {!showModalData && (
+                <Modal.Footer className="test-plan-updater-footer">
+                    {(!closeButton && (
+                        <Button variant="secondary" onClick={deleteOldTestPlan}>
+                            Next
```
alflennik, line 532: Could this be red and say "Delete"? That's assuming this suggestion doesn't conflict with any specific direction you've already received about how to implement this.

client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
```diff
@@ -0,0 +1,555 @@
...
+                                        <Form.Check.Input
+                                            type="checkbox"
+                                            onChange={() => {
+                                                setBackupChecked(
+                                                    !backupChecked
+                                                );
+                                            }}
+                                            checked={backupChecked}
+                                        />
+                                        Backup Test Plan before Update
```
alflennik, line 471: What would you think about changing this to "Retain old test plan"?

Or if the copy is pretty-much decided already, could you change this to "Back up test plan before update"? I do prefer the word "retain" if that's possible.

alflennik (approved):
Looking solid! I couldn't really find any notes for the parts of the code I've already reviewed, I'm really happy with it. I did have a couple suggested changes inside the brand new additions, which hopefully won't be too involved. Almost there!

client/components/TestPlanUpdater/TestPlanUpdaterModal.css
evmiguel: Screen Shot 2022-03-22 at 3 10 33 PM

The backup checkbox gets rendered conditionally, so I need the styles. to keep the checkbox on the left and the two buttons on the right.

alflennik: Ah, I get it, there's some styles and HTML to keep the checkbox above the buttons. Ok, so the main thing I wanted to address was the fact the button didn't have enough space and was wrapping. I would say, if you could figure out a way to fix that, that's the main thing motivating my suggestion. I'm happy to look at it together if that helps.

client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
evmiguel: I have no preference. @s3ththompson @isaacdurazo what do you think is the preferred wording?

client/components/TestPlanUpdater/TestPlanUpdaterModal.css
evmiguel: @alflennik I've addressed the text wrapping. It's on sandbox. Mind taking a look?

alflennik (approved):

client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
alflennik: I do think the existing test will work, thanks for updating it.

client/components/TestPlanUpdater/TestPlanUpdaterModal.css
alflennik: Looks great.

PR #405, opened 2022-03-16
Add support for build folder being removed on aria-at

howard-e:
aria-at-app TestRenderer depends on artifacts which were removed following the merge of https://github.com/w3c/aria-at/pull/621. In combination with https://github.com/w3c/aria-at/pull/646, this PR requires updates to return to the expected behavior by:
1. Update the `server/scripts/import-tests` script to now build the required files locally to populate the database.
2. Does a date check when rendering the example test window to maintain backward compatibility with the old structure of the aria-at repo, while also accounting for the new structure.

cc @s3ththompson  @richnoah


client/components/TestRenderer/index.jsx
```diff
@@ -245,14 +246,26 @@ const TestRenderer = ({
...
         testRunIO.setConfigInputFromQueryParamsAndSupport(configQueryParams);
 
+        // https://github.com/w3c/aria-at/commit/9d73d6bb274b3fe75b9a8825e020c0546a33a162
+        // This is the date of the last commit before the build folder removal.
+        // Meant to support backward compatability until the existing tests can
+        // be updated to the current structure
+        const buildRemovalDate = new Date('2022-03-10 18:08:36.000000 +00:00');
+        const testUpdatedAtDate = new Date(testUpdatedAt);
+
+        const finalTestPageUrl =
```
howard-e, line 256: Based on discussion with @alflennik, we may also be able to completely remove the dependency on the implicit commit attached to this date and reference the file directly from the repo's root. ie. `<proxy_domain>/tests/<testplan>/path/file.html` instead of `<proxy_domain>/<commit>/tests/<testplan>/path/file.html` but may require further discussion.

client/package-lock.json
```diff
@@ -1,131 +0,0 @@
-{
```
howard-e, line 1: The existence of this caused issues for me during a fresh `yarn install` and it isn't seemingly required when running `yarn install` from the project root due to the workspace configuration.

But let me know if I should add it back.

server/package.json
```diff
@@ -30,10 +30,12 @@
     "apollo-server-express": "^3.4.0",
     "axios": "^0.21.1",
     "connect-pg-simple": "^6.2.1",
+    "cross-spawn": "^7.0.3",
     "dotenv": "^8.2.0",
     "dotenv-cli": "^4.0.0",
     "express": "^5.0.0-alpha.8",
     "express-session": "^1.17.2",
+    "fs-extra": "^10.0.1",
```
howard-e, line 38: Noticed this dependency was missing despite being imported in `server/scripts/import-tests/index.js` so I've now added it.

I suppose it was being used through another package's dependencies?

evmiguel (approved):
Works for me!

Issue #406, opened 2022-03-18
AT and Browser Version management Mockups


isaacdurazo:
# AT and Browser Version management Text-Based and Visual Mockups

## Introduction
This issue presents a series of mockups for how a tester can provide AT and Browser version details when starting a new Test Run and for Admins to add new AT versions to the ARIA AT App. The changes presented here take into consideration several different scenarios for each type of user. There might be edge cases that are not considered yet. Please make a comment on this issue if you identify one that's not being taken into account here.

## Assistive Technology and Browser Details Modal

When a Tester or an Admin opens a Test Plan from the Test Queue, they will be prompted with a modal before landing on the Test Run page. This modal has different variations based on the different scenarios a Tester or an Admin can run into. I will first explain the general UI characteristics of this modal and will expand on the specifics of the different scenarios for each user persona afterward.

### General UI Characteristics and details

- **Heading:** This reads: Assistive Technology and Browser Details

- **Description:** This reads: Before continuing, please provide details about the Assistive Technology and Browser versions you will be testing with.

- **Assistive Technology Details:** This area has two columns with a `label` and a `select` element each. One column is dedicated to the AT Name, which can't be modified and the other column is dedicated to the AT Version, which can be modified if needed by changing the dropdown selection.

- **Browser Details:** This area has two columns with a `label` and a `select` element each. One column is dedicated to the Browser Name, which can't be modified and the other column is dedicated to the Browser Version, which can be modified if needed by changing the dropdown selection.

- **Actions:** This area contains the buttons for the Modal. The buttons are "Cancel" and a "Save and Continue"

- **Alerts:** Depending on the scenario, different kinds of alerts can be triggered in different areas within the modal.

#### Scenario 1 -  A Tester Opens a Test Plan for the first time
When a Tester opens a Test Plan for the first time, they must choose an AT and a Browser Version before being able to continue. The variations in this scenario are:

- **Assistive Technology Details:** The AT Name will be selected by default because this information comes from the Test Queue; this selection can't be modified. The tester will have to manually select a version for the AT to be able to continue.

- **Browser Details:** The Browser Name will be selected by default because this information comes from the Test Queue; this selection can't be modified. The Browser version is automatically detected and selected in the dropdown.

- **Alerts:** An informational Alert in light blue color is displayed above the Browser details that reads: We have automatically detected your version of [Browser Name]

[...]

PR #411, opened 2022-04-06
DRAFT: GraphQL updates based on issue #406

alflennik:
This PR proposes GraphQL schema changes for issue https://github.com/w3c/aria-at-app/issues/406. Some changes might not be obvious from reading the diff:

- Giving TestResults their own table, instead of continuing to store them as JSON data in the TestPlanRun table. This will permit setting up Postgres associations between each TestResult and the AT and Browser versions.
- Giving Tests a table as well, which will keep the Test model symmetrical with the TestResult model.
- The FindOrCreateTestPlanReport mutation will no longer automatically create atVersions and browserVersions, since mutations are being added to explicitly manage these.

### DIFF
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -114,31 +114,6 @@ const graphqlSchema = gql`
         atVersions: [String]!
     }
 
-    # TODO: remove or rework this type in order to support recording exact
-    # versions
-    """
-    Specifies an AT and browser combination to test, divided into buckets by
-    major version.
-    """
-    type TestPlanTarget {
-        """
-        Postgres-provided numeric ID.
-        """
-        id: ID!
-        title: String!
-        at: At!
-        atVersion: String!
-        browser: Browser!
-        browserVersion: String!
-    }
-
-    input TestPlanTargetInput {
-        atId: ID!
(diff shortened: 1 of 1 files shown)


server/graphql-schema.js
```diff
@@ -736,7 +711,8 @@ const graphqlSchema = gql`
     """
     input TestPlanReportInput {
         testPlanVersionId: ID!
-        testPlanTarget: TestPlanTargetInput!
+        atId: ID!
```
howard-e, line 714: Thanks for looking at this so quickly! I know this is still marked as a draft but was wondering if this shouldn't include `atVersion` and `browserVersion` as well?

alflennik: Thank YOU for looking at this so quickly 😁 So, splitting up the At / Browser and AtVersion / BrowserVersion is intentional - I put the atVersion and browserVersion fields on the TestResult type. This means that reports are locked to an AT and Browser, but they're no longer locked to an exact version. And now each TestResult can have slightly different versions. Let me know if you imagined something different maybe?

alflennik:
Closed because Erika took this one over the finish line!

PR #414, opened 2022-04-12
Create/Edit AT Version

evmiguel:
This PR implements [ENV DATA: 1. Backend work required to edit AT Versions](https://app.asana.com/0/1201709804960055/1202072717120762). 

The following changes have been made:

- Added an availability column to the AtVersion table to support the availability date
- Updated AtService to handle availability field
- Updated GraphQL schema to support AT version type and mutations for adding/editing AT versions
- Implemented resolvers for adding/editing AT versions

### DIFF
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -114,6 +114,29 @@ const graphqlSchema = gql`
         atVersions: [String]!
     }
 
+    """
+    A version of an Assistive Technology
+    """
+    type AtVersion {
+        """
+        The Postgres given ID for an AT
+        """
+        atId: ID!
+        """
+        The string version of an AT
+        """
+        atVersion: String!
+        """
+        The approximate date of availability for the AT Version
+        """
+        availability: String
+    }
+
+    input AtVersionInput {
+        atVersion: String
+        availability: String
+    }
+
     # TODO: remove or rework this type in order to support recording exact
     # versions
     """
@@ -856,6 +879,26 @@ const graphqlSchema = gql`
     }
 
     # Mutation-specific types below
+    """
+    Mutations scoped to a given Assistive Technology
+    """
+    type AtVersionOperations {
+        """
+        Create a new AT Version if one doesn't exist
+        """
+        findOrCreateAtVersion(
+            atVersion: String!
+            availability: String
+        ): AtVersion!
+        """
+        Edits the AT Version
+        """
+        editAtVersion(
+            atVersion: String!
+            availability: String
+            updateParams: AtVersionInput!
+        ): AtVersion!
+    }
 
     """
     Mutations scoped to a previously-created TestPlanReport.
@@ -942,6 +985,11 @@ const graphqlSchema = gql`
     }
 
     type Mutation {
+        """
+        Get available mutations for a Assistive Technology Versions,
+        given an Assistive Technology
+        """
+        atVersion(id: ID!): AtVersionOperations!
         """
         Adds a report with the given TestPlanVersion and TestPlanTarget and a
         state of "DRAFT", resulting in the report appearing in the Test Queue.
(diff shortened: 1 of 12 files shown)


server/tests/integration/graphql.test.js
```diff
@@ -133,10 +133,13 @@ describe('graphql', () => {
         const excludedTypeNames = [
```
evmiguel, line 133: I excluded the new types from the GraphQL test because I'd like to understand how the test works.

cc: @alflennik

server/models/services/AtService.js
```diff
@@ -194,14 +194,14 @@ const removeAt = async (id, deleteOptions = { truncate: false }) => {
...
-    { atId, atVersion },
+    { atId, atVersion, availability = null },
     atVersionAttributes = AT_VERSION_ATTRIBUTES,
     atAttributes = AT_ATTRIBUTES,
     options = {}
 ) => {
     return ModelService.getByQuery(
         AtVersion,
-        { atId, atVersion },
+        { atId, atVersion, ...(availability && { availability }) },
```
howard-e, line 204: Not sure if this property is ever likely to be used when querying but shouldn't affect anything. You should also be able to pass it as a query even if `null` so may not need the null check here.

howard-e (approved):
@evmiguel looks good!

server/graphql-schema.js
```diff
@@ -856,6 +883,26 @@ const graphqlSchema = gql`
...
+        ): AtVersion!
+        """
+        Edits the AT Version
+        """
+        editAtVersion(
+            atVersion: String!
+            availability: String
+            updateParams: AtVersionInput!
+        ): AtVersion!
+    }
```
alflennik, line 905: One thing we should figure out is how to resolve some of the contradictions between this PR and the schema proposal I made here: https://github.com/w3c/aria-at-app/pull/411. One example difference is that I had the creating and editing AT versions under the "AtOperations" type but you have those functions under an "AtVersionOperations" type.

server/graphql-schema.js
```diff
@@ -810,6 +833,10 @@ const graphqlSchema = gql`
         """
         ats: [At]!
         """
+        Get AT Version information, given an AT ID and version string
+        """
+        atVersion(atId: ID!, atVersion: String!): AtVersion!
```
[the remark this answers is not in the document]
evmiguel: See comments below in bold:

> I was looking at this for a little while, trying to put my finger on what seemed off about it, and then I realized that there are two concepts here that have the name "atVersion". First, atVersion is a string, which contains the version string for the AT. Second, AtVersion is a type with fields like availability.
> 
> What I've been thinking is something like this:
> 
> ```
> type AtVersion {
>   id: ID!
>   name: String!
>   releasedAt: Timestamp!
> }
> ```
> 
> Some thoughts on this:
> 
> * Adding an ID makes some sense since we are going to allow editing the AtVersions. Otherwise the API gets really awkward where you need to provide the old AtVersion name in order to provide a new AtVersion name.

**I agree with you on using IDs. I'm curious about how each AtVersion would be associated with an AT. Perhaps, just leaving that column in the table?**

> * We would want to use an ID anywhere we are currently using the atVersion string. This does mean probably migrating the existing data (!) and I realize this might get a little hairy but ... I do think ... it's doable.

**This seems like something we have to do.**

> * With an ID, it would make editing the AtVersion name way more intuitive. The type is currently a bit special and tricky and it would be great to make it look like the other types.

**+1**

> * There is already a custom "scalar" type specifically for timestamps so I put it in there.
> * The reason I suggested "releasedAt" is that all the timestamps currently in the schema end with the word "at" which is sorta neat I guess.

**Agreed. Let's replace the availability field with this.**

> * The other big thing I proposed there is making the releasedAt a required field. I feel like it should be required? Or not? What would be a reason to leave it blank?

**The only reason I left it optional was for the existing entries without that data.**

PR #417, opened 2022-04-27
AT Version Table ID update

evmiguel:
This PR changes the AtVersion table. It contains migrations to add columns and convert the composite primary key into a serial id primary key. The PR also addresses the affected files on the backend.

### DIFF
--- a/server/migrations/20220427192638-changeConstraintsAtVersion.js
+++ b/server/migrations/20220427192638-changeConstraintsAtVersion.js
@@ -1,8 +1,8 @@
 'use strict';
 
 module.exports = {
-    up: async queryInterface => {
-        queryInterface.sequelize.transaction(async transaction => {
+    up: queryInterface => {
+        return queryInterface.sequelize.transaction(async transaction => {
             await queryInterface.removeConstraint(
                 'AtVersion',
                 'AtVersion_pkey',
@@ -11,8 +11,8 @@ module.exports = {
         });
     },
 
-    down: async queryInterface => {
-        queryInterface.sequelize.transaction(async transaction => {
+    down: queryInterface => {
+        return queryInterface.sequelize.transaction(async transaction => {
             await queryInterface.addConstraint(
                 'AtVersion',
                 {
--- a/server/migrations/20220427192638-changeConstraintsAtVersion.js
+++ b/server/migrations/20220427192638-changeConstraintsAtVersion.js
@@ -1,0 +1,27 @@
+'use strict';
+
+module.exports = {
+    up: async queryInterface => {
+        queryInterface.sequelize.transaction(async transaction => {
+            await queryInterface.removeConstraint(
+                'AtVersion',
+                'AtVersion_pkey',
+                { transaction }
+            );
+        });
+    },
+
+    down: async queryInterface => {
+        queryInterface.sequelize.transaction(async transaction => {
+            await queryInterface.addConstraint(
+                'AtVersion',
+                {
+                    type: 'primary key',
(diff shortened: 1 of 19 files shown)


server/migrations/20220427204420-changeAtVersionColumn.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
```
evmiguel, line 1: @howard-e @alflennik This is the last migration file. For some reason changing the column name from `atVersion` to `name` isn't working. Trying to figure this one out, but could use some help

howard-e: What is the error that's thrown?

You could try removing the `atVersion` column then adding the `name` column in the migration if all else fails

howard-e: Okay seems that's solved! 👍🏽

server/migrations/20220427204420-changeAtVersionColumn.js
```diff
@@ -0,0 +1,21 @@
...
+    },
+
+    down: async queryInterface => {
+        queryInterface.sequelize.transaction(async transaction => {
+            await queryInterface.sequelize.query(
+                'ALTER TABLE public."AtVersion" RENAME COLUMN "name" TO "atVersion";',
+                { transaction }
+            );
+        });
+    }
```
alflennik, line 20: It should be possible to combine all four migrations into one, that'll just make the migration more compact and easier to take in all at once.

alflennik: Does the migration work? Does the ID field auto-populate? If so, that's one of the biggest hurdles overcome, which would be sweet.

alflennik: I think we should also make a symmetrical set of changes to the Browser table, i.e. giving it an ID field and possibly a releasedAt field too. @s3ththompson are we going to track the dates where browser versions were released as well?

evmiguel: > Does the migration work? Does the ID field auto-populate? If so, that's one of the biggest hurdles overcome, which would be sweet.

Yes, the IDs are auto-populated.

evmiguel: > It should be possible to combine all four migrations into one, that'll just make the migration more compact and easier to take in all at once.

I was running into some issues with the ordering of changes, which is why I broke them out. I know multiple files isn't ideal, but they are easier to read, if anything.

server/migrations/20220427192638-changeConstraintsAtVersion.js
```diff
@@ -0,0 +1,27 @@
...
+                'AtVersion',
+                'AtVersion_pkey',
+                { transaction }
+            );
+        });
+    },
+
+    down: queryInterface => {
+        return queryInterface.sequelize.transaction(async transaction => {
+            await queryInterface.addConstraint(
```
howard-e, line 16: Seems this should be changed from the [v6](https://sequelize.org/api/v6/class/src/dialects/abstract/query-interface.js~queryinterface#instance-method-addConstraint) version to the [v5](https://sequelize.org/v5/class/lib/query-interface.js~queryinterface#instance-method-addConstraint) as well

evmiguel: Good catch!

howard-e (approved):

alflennik (approved):

PR #418, opened 2022-05-02
Create, Edit, Delete AT Versions by ID

evmiguel:
This PR handles [ENV DATA: 1. Backend work required to edit AT Versions](https://app.asana.com/0/1201709804960055/1202072717120762) and [ENV DATA: 2. Backend work required to remove AT Versions](1202072717120764).

### DIFF
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -879,10 +879,10 @@ const graphqlSchema = gql`
     Mutations scoped to an Assistive Technology version
     """
     type AtOperations {
-        createAtVersion(name: String!, releasedAt: Timestamp!): NoResponse
+        createAtVersion(name: String!, releasedAt: Timestamp!): AtVersion!
     }
     type AtVersionOperations {
-        editAtVersion(updatedName: String!): NoResponse
+        editAtVersion(updatedName: String!): AtVersion!
         deleteAtVersion: NoResponse
     }
     """
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -882,7 +882,10 @@ const graphqlSchema = gql`
         createAtVersion(name: String!, releasedAt: Timestamp!): AtVersion!
     }
     type AtVersionOperations {
-        editAtVersion(updatedName: String!): AtVersion!
+        editAtVersion(
+            updatedName: String!
+            updatedReleasedAt: Timestamp!
+        ): AtVersion!
         deleteAtVersion: NoResponse
     }
     """
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -129,7 +129,7 @@ const graphqlSchema = gql`
         """
         Date for approximate availability of the version
         """
-        releasedAt: Timestamp!
+        releasedAt: Timestamp
     }
 
     # TODO: remove or rework this type in order to support recording exact
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -129,7 +129,7 @@ const graphqlSchema = gql`
         """
         Date for approximate availability of the version
         """
-        releasedAt: Timestamp
+        releasedAt: Timestamp!
     }
 
     # TODO: remove or rework this type in order to support recording exact
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -876,6 +876,16 @@ const graphqlSchema = gql`
     # Mutation-specific types below
 
     """
+    Mutations scoped to an Assistive Technology version
+    """
+    type AtOperations {
+        createAtVersion(name: String!, releasedAt: Timestamp!): NoResponse
+    }
+    type AtVersionOperations {
+        editAtVersion(updatedName: String!): NoResponse
+        deleteAtVersion: NoResponse
+    }
+    """
     Mutations scoped to a previously-created TestPlanReport.
     """
     type TestPlanReportOperations {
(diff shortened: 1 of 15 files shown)


server/models/services/AtService.js
```diff
@@ -184,6 +184,28 @@ const removeAt = async (id, deleteOptions = { truncate: false }) => {
...
+ */
+const getAtVersionById = async (
+    id,
+    atAttributes = AT_ATTRIBUTES,
+    options = {}
+) => {
+    return ModelService.getById(
+        AtVersion,
+        id,
+        atAttributes,
```
alflennik, line 203: I think line 203 should be atVersionAttributes not atAttributes.

server/models/services/AtService.js
```diff
@@ -255,17 +276,17 @@ const getAtVersions = async (
  * @returns {Promise<*>}
  */
 const createAtVersion = async (
-    { atId, atVersion },
+    { atId, name, releasedAt = null },
```
alflennik, line 279: Maybe releasedAt doesn't need a default value? Because I think it's a required field.

server/models/services/AtService.js
```diff
@@ -282,17 +303,48 @@ const createAtVersion = async (
  * @returns {Promise<*>}
  */
 const updateAtVersionByQuery = async (
-    { atId, atVersion },
+    { atId, name, releasedAt = null },
```
alflennik, line 306: Same note about the default value as above.

server/resolvers/AtOperations/createAtVersionResolver.js
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    let version = await getAtVersionByQuery({ atId, name, releasedAt });
+
+    if (!version) {
+        version = await createAtVersion({ atId, name, releasedAt });
+    }
+
+    return version;
+};
```
alflennik, line 23: This is clean, nicely done.

server/graphql-schema.js
```diff
@@ -942,6 +970,8 @@ const graphqlSchema = gql`
     }
 
     type Mutation {
+        at(id: ID!): AtOperations!
+        atVersion(id: ID!): AtVersionOperations!
```
alflennik, line 974: I like the way you did this! Nice.

server/migrations/20220427145338-alterAtVersionTable.js
```diff
@@ -0,0 +1,38 @@
...
+                },
+                { transaction }
+            );
+            await queryInterface.addColumn(
+                'AtVersion',
+                'releasedAt',
+                {
+                    type: Sequelize.DataTypes.DATE
+                },
+                { transaction }
```
alflennik, line 23: I feel like this might be a good migration to add a value for releasedAt, making sure that column is fully populated. When it's populated it will be possible to make the field required. As for what value to use, I don't think it matters much. It can be today's date. In a future migration in the TestPlanTarget PR I'm going to delete all the currently existing versions anyway.

server/tests/integration/graphql.test.js
```diff
@@ -137,6 +137,7 @@ describe('graphql', () => {
         const excludedTypeNameAndField = [
             // Items formatted like this:
             // ['TestResult', 'startedAt'],
+            ['AtVersion', 'releasedAt']
```
alflennik, line 140: This line won't be needed when there's a populated value for this column.

alflennik (commented):
Wow, this PR is pretty epic, I hope you took a moment to pat yourself on the back 😜 Also I think it's just a couple tweaks from being able to go in too. The main thread of my comments was around making the releasedAt field required. I'm excited to use this as a base for the migration I'm working on.

server/models/services/AtService.js
evmiguel: Good catch!

evmiguel: Good catch!

server/graphql-schema.js
```diff
@@ -875,6 +875,16 @@ const graphqlSchema = gql`
...
     # Mutation-specific types below
 
+    """
+    Mutations scoped to an Assistive Technology version
+    """
+    type AtOperations {
+        createAtVersion(name: String!, releasedAt: Timestamp!): AtVersion!
+    }
+    type AtVersionOperations {
+        editAtVersion(updatedName: String!): AtVersion!
```
howard-e, line 885: Think `updatedReleasedAt` is a missing property here?

server/resolvers/AtVersionOperations/editAtVersionResolver.js
```diff
@@ -0,0 +1,15 @@
+const { AuthenticationError } = require('apollo-server');
+const { updateAtVersionById } = require('../../models/services/AtService');
+const editAtVersionResolver = async (
+    { parentContext: { id } },
+    { updatedName },
```
howard-e, line 5: Same question as before. Are we missing the update field for `releasedAt`?

howard-e (commented):
@evmiguel this all looks good to me and works well!

Only comment I have is around the missing property to update  the `releasedAt` entry for the atVersion as well. Currently, it seems that should be supported based on 406. Not sure if I may have missed any follow-ups?

server/graphql-schema.js
howard-e: Thanks, looks good!

alflennik (approved):

PR #419, opened 2022-05-03
Add New Manage Test Queue Component

howard-e:
This implements the following components and scenarios based on #406:
* Test Queue - Admin View
  * Manage Assistive Technology Versions
  * Add Test Plans to the Test Queue
* Managing Assistive Technology Versions Modals
  * Scenario 1 - Adding a New AT Version Modal
  * Scenario 2 - Edit an Existing AT Version Modal
  * Scenario 3 - Removing an Existing AT Version - Danger Modal
  * Scenario 4 - Adding an Existing AT Version
  * Scenario 5 - Removing an AT Version already in use - Warning Modal

_Note: WIP while server-based work to account for this feature continues_


client/components/common/UpdateVersionModal/index.jsx
```diff
@@ -34,6 +42,21 @@ const UpdateVersionModal = ({
         setUpdatedDateAvailabilityText(value);
     };
 
+    const handleDateAvailabilityTextKeyPress = e => {
+        let input = e.target;
+        if ((e.charCode < 48 && e.charCode !== 45) || e.charCode > 57) {
```
alflennik, line 47: Maybe assign these codes to variables so it's easier to tell what specific keys we're looking for?

client/tests/__mocks__/GraphQLMocks.js
```diff
@@ -89,6 +95,75 @@ export const TEST_QUEUE_PAGE_POPULATED_MOCK_ADMIN = [
...
+                            {
+                                id: '2',
+                                name: '2020.1',
+                                releasedAt: null
+                            },
+                            {
+                                id: '1',
+                                name: '2019.3',
+                                releasedAt: null
+                            }
```
alflennik, line 138: releasedAt is a required field so probably best to throw values in here.

client/utils/formatter.js
```diff
@@ -19,3 +21,14 @@ export const capitalizeEachWord = (
 
     return words.join(joinChar);
 };
+
+export const convertDateToString = (date, format = 'DD-MM-YYYY') => {
+    if (!date) return '';
+    return moment(date)
+        .format(format)
+        .toString();
```
alflennik, line 29: Maybe `toString()` is redundant? I think `format` returns a string (?)

alflennik (approved):
Looks good just a couple little comments

client/components/common/UpdateVersionModal/index.jsx
howard-e: Good point. Was hoping the following comment was enough but reading it back, this could be explained much better with variable names and expanding on that comment.

client/utils/formatter.js
howard-e: Seems to be the case. Good catch!

alflennik (commented):
I'll need to continue reviewing this, I feel like I only got through 10% of all the capabilities we've added and I know a few more are incoming too. Here are a couple of thoughts I did have which hopefully are helpful.

isaacdurazo (commented):
This is a test. Please ignore, Howard 😅

isaacdurazo (changes requested):
Another test.

isaacdurazo (approved):

client/components/common/UpdateVersionModal/index.jsx
```diff
@@ -64,8 +106,11 @@ const UpdateVersionModal = ({
                         </Form.Label>
                         <Form.Control
                             type="text"
-                            value={dateAvailabilityText}
+                            placeholder="DD-MM-YYYY"
+                            value={updatedDateAvailabilityText}
                             onChange={handleDateAvailabilityTextChange}
+                            onKeyPress={handleDateAvailabilityTextKeyPress}
+                            maxLength={10}
                         />
```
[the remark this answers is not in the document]
howard-e: @alflennik good points! I had initially overlooked such validation at the time with the plan to return to it and never did, so thanks for pointing that out.
I've now pushed commits which should include proper feedback to these situations. Please check again when you can

evmiguel (approved):
@howard-e (great job!) and I walked through the code together. I tested the different scenarios as well. Changes from #423 have also been added to this PR.

PR #420, opened 2022-05-04
Browser Version Operations

evmiguel:
This PR handles [ENV DATA: 20. Update Browser table and resolvers](https://app.asana.com/0/1201709804960055/1202219817103347). The following changes have been made:

- Migration for the BrowserVersion table to support id primary keys and column change from `browserVersion` to `name`
- Update GraphQL schema to support BrowserVersion type and CRUD mutations for Browser and Browser Version
- Update backend Models and tests to handle database changes and support the resolver functions

To test in dev:
- Run the latest migration for dev: `yarn sequelize db:migrate;`
- Go to localhost:3000/api/graphql to try out the mutations

For running jest tests, rebuild the test database by dropping the `aria_at_report_test` database and afterwards, running through [these docs](https://github.com/w3c/aria-at-app/blob/main/docs/database.md#test-database)

### DIFF
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -999,15 +999,15 @@ const graphqlSchema = gql`
         """
         at(id: ID!): AtOperations!
         """
-        Get the available mutations for the given AT Version.
+        Get the available mutations for the given AT version.
         """
         atVersion(id: ID!): AtVersionOperations!
         """
-        Get the available mutations for the given Browser.
+        Get the available mutations for the given browser.
         """
         browser(id: ID!): BrowserOperations!
         """
-        Get the available mutations for the given Browser Version.
+        Get the available mutations for the given browser version.
         """
         browserVersion(id: ID!): BrowserVersionOperations!
         """
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -900,6 +900,19 @@ const graphqlSchema = gql`
         deleteAtVersion: NoResponse
     }
     """
+    Mutations scoped to an Browser version
+    """
+    type BrowserOperations {
+        createBrowserVersion(name: String!): BrowserVersion!
+    }
+    type AtVersionOperations {
+        editAtVersion(
+            updatedName: String!
+            updatedReleasedAt: Timestamp!
+        ): AtVersion!
+        deleteAtVersion: NoResponse
+    }
+    """
     Mutations scoped to a previously-created TestPlanReport.
     """
     type TestPlanReportOperations {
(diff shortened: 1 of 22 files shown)


server/graphql-schema.js
```diff
@@ -973,9 +994,23 @@ const graphqlSchema = gql`
...
+        """
+        Get the available mutations for the given AT.
+        """
         at(id: ID!): AtOperations!
+        """
+        Get the available mutations for the given AT Version.
+        """
         atVersion(id: ID!): AtVersionOperations!
         """
+        Get the available mutations for the given Browser.
```
alflennik, line 1006: Browser isn't capitalized 😝 sorry I have to play my former-English-teacher card.

server/graphql-schema.js
```diff
@@ -973,9 +994,23 @@ const graphqlSchema = gql`
...
+        """
+        browser(id: ID!): BrowserOperations!
+        """
+        Get the available mutations for the given Browser Version.
```
alflennik, line 1010: Same as above.

server/resolvers/BrowserVersionOperations/editBrowserVersionResolver.js
```diff
@@ -0,0 +1,25 @@
...
+    { updatedName },
+    { user }
+) => {
+    if (
+        !(
+            (
+                user?.roles.find(role => role.name === 'ADMIN') ||
+                user?.roles.find(role => role.name === 'TESTER')
+            ) // Tester is allowed to change Browser Version when it cannot be automatically detected
+        )
```
alflennik, line 17: This could be something like `if (user === null) {`! I don't think a comment is needed since the code is pretty clear.

server/resolvers/browsersResolver.js
```diff
@@ -1,7 +1,7 @@
 const { getBrowsers } = require('../models/services/BrowserService');
 
 const browsersResolver = () => {
-    return getBrowsers(undefined, undefined, undefined, undefined, {
+    return getBrowsers(undefined, undefined, undefined, undefined, undefined, {
         order: [['name', 'asc']]
     });
```
alflennik, line 6: The previous version seems correct to me! Are you sure the change is needed?

alflennik (approved):
You've dispatched this feature with extreme prejudice, nice. Just a couple little notes.

server/resolvers/browsersResolver.js
evmiguel: Good catch! After looking at the `getBrowsers` API, it is not needed.

howard-e (approved):

PR #422, opened 2022-05-09
Fix Safari TestRenderer not correctly re-rendering

howard-e:
An edge case was found for the TestRenderer component when submitting the form information on Safari. It appears that there is an inconsistent timing with how React's `useLayoutEffect` hook works on Safari.

How this should work is that after populating the testing tasks' fields on the Test Run page and submitting the form, the request would be sent off to the server to be submitted to the database and through a re-render of the TestRenderer component, the summary table of the submission is presented. This works as expected on Chrome and Firefox.
When submitted on Safari, the form data is cleared and errors are displayed for all the fields (although the data was successfully submitted to the database)

This PR addresses this by doing necessary hook method cleanups and with an additional temporary fix to address another issue around using the edit form functionality introduced by the cleanup of the hook methods.

Known Issues:

- [x] Brief flicker when using `Edit Results` functionality

### DIFF
--- a/client/components/TestRenderer/index.jsx
+++ b/client/components/TestRenderer/index.jsx
@@ -1,4 +1,10 @@
-import React, { Fragment, useEffect, useLayoutEffect, useState } from 'react';
+import React, {
+    Fragment,
+    useEffect,
+    useLayoutEffect,
+    useState,
+    useRef
+} from 'react';
 import styled from '@emotion/styled';
 import PropTypes from 'prop-types';
 import nextId from 'react-id-generator';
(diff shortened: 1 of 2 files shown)


evmiguel (approved):
I tried this branch in Safari, and it worked! Great job on the cleanup and the commenting. I learned a bunch about the TestRenderer and `useRef` from this PR.

alflennik (approved):
This can't have been easy! Huge congrats to fixing this very challenging issue that bested me on more than one occasion. It might be nice in the future to look holistically at how we're loading all the data, but for now it seems good to me.

PR #427, opened 2022-05-31
Allow tester to create Browser Version

howard-e:
Allows for a tester to create a browser version through the `AtAndBrowserDetailsModal`.


evmiguel (approved):
Looks good and works well!

PR #428, opened 2022-06-02
Fix for displaying Run History with unfinished test results from testers

howard-e:


client/components/Reports/SummarizeTestPlanReport.jsx
```diff
@@ -95,7 +95,8 @@ const getTestersRunHistory = (
         if (
             testPlanReportId === testPlanReport.id &&
             testPlanReport.status === 'FINALIZED' &&
-            testResult
+            testResult &&
+            testResult.completedAt
```
alflennik, line 99: It could be `testResult?.completedAt` which is more compact 🙂 but that's optional because adding a second line might make the intent more clear.

alflennik (approved):

client/components/Reports/SummarizeTestPlanReport.jsx
howard-e: I'm okay with making this change, thanks

Issue #429, opened 2022-06-02
Combo boxes do not announce how many items or relationship


richnoah:
During testing conducted for the AT and Browser Capture feature, the following issue was discovered.  Comboboxes for browser and screen reader versions do not announce how many items there are or the relationship of the focused item compared to the overall number of items that can be selected.

PR #431, opened 2022-06-07
Accessibility fixes for combobox and disclosures

howard-e:
Addresses #429.

This PR moves away from using bootstrap's `<Form.Control as="select" />` which used the native HTML `select` widget for a user to select possible values from a list. In it's place, a [combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) has been implemented which is based on the [Combobox APG Select Only Example](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html).

Additionally, the robustness of the disclosure components used with the `ManageTestQueue` component and `Run History` on relevant reporting pages has been improved.

### DIFF
--- a/client/components/ManageTestQueue/index.jsx
+++ b/client/components/ManageTestQueue/index.jsx
@@ -13,6 +13,7 @@ import PropTypes from 'prop-types';
 import BasicModal from '../common/BasicModal';
 import UpdateVersionModal from '../common/UpdateVersionModal';
 import BasicThemedModal from '../common/BasicThemedModal';
+import SelectCombobox from '../common/SelectCombobox';
 import {
     ADD_AT_VERSION_MUTATION,
     EDIT_AT_VERSION_MUTATION,
@@ -139,6 +140,10 @@ const DisclosureContainer = styled.div`
         grid-auto-flow: column;
         grid-template-columns: 1fr 1fr 1fr 1fr;
         grid-gap: 1rem;
+    }
+
+    .form-group {
+        min-width: 100%;
     }
 
     .disclosure-form-label {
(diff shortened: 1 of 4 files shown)


jscholes:
@howard-e This is significantly concerning.  The APG's select-only combobox pattern does not work with macOS Safari, so unless steps have been taken to address that, the PR will exclude macOS testers and admins.  It's also the case that the native `<select>` has good accessibility for single-select cases, and it would have potentially been less disruptive to fix whatever the problems were with the mark-up than implementing a new component from scratch.

howard-e:
Thanks for raising that concern @jscholes. I did address the issues I found when with testing on Safari, but after another brief test, I've found another edge case. Will draft until additional use cases have been exhausted.

jscholes:
@howard-e How have you fixed the Safari issues?

howard-e:
> @howard-e How have you fixed the Safari issues?

A big one (that I've just noticed was accidentally stashed) was forcing focus to the last selected listbox option when the "openKeys" were used to avoid the issue of the listbox not being focused with keys which weren't 'ArrowDown'. Then keeping that focus looped within the listbox when the VO combinations to navigate was used, until closed. But even then, there still appears to be some caveats  with the navigation and feedback (aria-labels were used due to lack of feedback in some cases).

Thanks again for the feedback @jscholes. The implementation still appears to be quite brittle and I think it would make more sense for me to iterate on it and contribute to APG should I get it into a better state for review.

> @jscholes said: 
It's also the case that the native `<select>` has good accessibility for single-select cases, and it would have potentially been less disruptive to fix whatever the problems were with the mark-up

Agreed, will be closing this PR and addressing the feedback concerns of #429 with the native `<select>`.

The disclosure accessibility fixes included in this PR have already been pulled into #432.

PR #432, opened 2022-06-07
Improve robustness of disclosure component

howard-e:
Address rare case where VO could loop within disclosure.


alflennik (approved):
Thanks for fixing this! I checked the UI in VoiceOver and the issue is fixed, and I also gave it a couple successful passes in NVDA.

evmiguel (approved):
Looks good! I did not encounter any loops with open disclosures.

Issue #436, opened 2022-06-14
Version Collection Mockups revised


isaacdurazo:
## General updates
- All scenarios should include a Close icon (X) in the upper right corner, which will display an exit dialog (see design below). After confirming they want to exit they should be taken back to the Test Queue.
- If no changes have been made in the AT or Browser versions we should display only a "Continue" button
- If there has been a change, we should display a "Save and Continue" button and a "Cancel" button
- When a change has been made and the user saves and continues, we should display a Success toast message on the Test Run page (see design below).

### Opening a Test Run for the very first time.
- The Browser and AT names are predefined because this information comes from the Test Queue.
- The Browser Version is automatically detected but can be modified.
- The AT Version needs to be provided by the Tester.
- If the user clicks Save and Continue without providing an AT Version, we should display a validation error.

![ARIA-AT - Tester View - AT and Browser Details Modal - First view](https://user-images.githubusercontent.com/1379244/173665188-f958ffe6-9279-4380-9d33-adb35064319d.png)

### Coming back to continue testing  and my AT/Browser version hasn’t changed
- The only available action is “Continue”
- If the user modifies any of the selections in the form, the “Continue” button changes to “Save and Continue” and a “Cancel” button appears in the modal. (see next mockup)

![ARIA-AT - Tester View - AT and Browser Details Modal - Returning with no changes](https://user-images.githubusercontent.com/1379244/173665296-6e404dbf-e2be-45c7-bdfc-bf01588124cd.png)

- This is how the modal actions should change if the user makes a change
- The Cancel button lets the user proceed to the Test Run page, but without saving the changes

![ARIA-AT - Tester View - AT and Browser Details Modal - Returning with no changes and then making a change](https://user-images.githubusercontent.com/1379244/173665290-39606d69-060d-489c-a51d-688e33072fac.png)

### Coming back to continue testing with a different browser.
- In this scenario, we can’t automatically pre-fill the form with the new Browser details because this information is predefined in the Test Queue.
- A Continue button is the only button available.
- If for whatever reason the user modifies anything in the form, the Continue button should change to Save and Continue and a Cancel button should be displayed

![ARIA-AT - Tester View - AT and Browser Details Modal - Different Browser](https://user-images.githubusercontent.com/1379244/173665442-cfbe88ac-a486-42f3-9a35-2d5a1436fdf5.png)

### Opening a Test plan for the first time with the wrong Browser
[...]

PR #439, opened 2022-06-23
Version Collection revision

howard-e:
Implements version collection revisions submitted in #436.


evmiguel (approved):
Nice work @howard-e! I ran through each scenario as well. Looks good.

PR #441, opened 2022-06-30
Query speed improvments

howard-e:
**Resolvers and Server Changes**
* Created `runnableTestsLength` and `testResultsLength` resolvers to eliminate the need for expensive operations to be done in the GraphQL layer to create`runnableTests` and `testResults` to then get the same length.
  * Custom column queries were added onto existing dynamic queries using `sequelize.literal` to support this.
  * On the Test Queue page, when running locally, this reduced the query speed from `~6s` to `~2.85s`.

**Test Queue Page Changes**
* Staggered the loading on the Test Queue page such that conflicts are loaded after an initial because conflicts evaluation is another source of huge time being spent.
  * Added `TestQueuePageNoConflicts` to load data onto the Test Queue page except conflicts data.
  * Added `TestQueuePageConflicts` query to load the conflicts data and merge it with the initial data.
  * On the Test Queue page, when running locally, the initial query loads at `~700ms`, and the conflicts data query loading around `~2.2s`. Seeing around x1.8 those speeds in deployed environments.
  * This allows the user to view and interact with the page much quicker (NOTE: For admins, conflicts still need to load before being able to update a report's status; conflicts need to load before being able to view a report's status or number of conflicts; all other actions can be performed as expected during the conflicts loading period). 
* Other Test Queue Loading Changes
  * On initial load, status change actions are hidden from admins and the status text is replaced with a loading indicator until conflicts are loaded.
  * On all actions for a row except removing, the entire page's query no longer has to be refetched.
  * Added `TestQueuePageConflicts` to use the `TestPlanReportById` GraphQL function to make updating a single row much faster (such as assigning or unassigning), after a mutation.
  * When adding or removing a row from the Test Queue, the entire page's query will still need to be refetched (initial load + conflicts load).


s3ththompson:
@howard-e this looks great!

if i'm reading correctly, it seems like the takeaway is that pushing calculations into the Postgres query (rather than getting all the data and calculating in userspace) is much faster?

if so, that seems like good news for the future... if we hit a certain bottleneck, we can always try to move it into a Postgres query? although i suppose there's always some maintenance tradeoff there (since the raw SQL queries are a bit less flexible and harder to read)...

howard-e:
@s3ththompson 

> if i'm reading correctly, it seems like the takeaway is that pushing calculations into the Postgres query (rather than getting all the data and calculating in userspace) is much faster?

That's correct and significantly so as well. With just adding the `COUNT` queries for `runnableTestsLength` and `testResultsLength`, I've seen improvements from ~12s to ~900ms when testing locally. Would like to do the same with the `conflicts` calculation as well. It also avoids having the unnecessary need of pulling all the data to perform unavoidable operations on it just to get the `[data].length` in the end.

> if so, that seems like good news for the future... if we hit a certain bottleneck, we can always try to move it into a Postgres query? 

Yep, definitely a feasible option in future cases

> although i suppose there's always some maintenance tradeoff there (since the raw SQL queries are a bit less flexible and harder to read)...

We can keep the raw queries to a minimum and also try to make the wrapper around them as easy as possible to work with, in the hopes that any maintenance task(s) happens there. But going deeper to the query itself may just have to be something we have to be okay with should the need arise. Much of the other raw queries have remained untouched since being added so maybe that's a good sign!

client/components/TestQueue/index.jsx
```diff
@@ -9,14 +9,24 @@ import ManageTestQueue from '../ManageTestQueue';
...
 import { evaluateAuth } from '../../utils/evaluateAuth';
 import './TestQueue.css';
 
 const TestQueue = () => {
-    const { loading, data, refetch } = useQuery(TEST_QUEUE_PAGE_QUERY);
+    const { loading, data, refetch } = useQuery(
+        TEST_QUEUE_PAGE_NO_CONFLICTS_QUERY
+    );
+    const { data: conflictsData } = useQuery(TEST_QUEUE_PAGE_CONFLICTS_QUERY, {
+        skip: !data
```
evmiguel, line 24: Slick.

evmiguel (approved):
Well done, @howard-e! The reorganization in the backend yields to a very clean frontend change. Are there plans to revisit the skipped tests?

howard-e:
> The reorganization in the backend yields to a very clean frontend change.

@evmiguel thanks for taking a look! 

> Are there plans to revisit the skipped tests?

And definitely need to revisit the tests. A few of these skipped tests exist based on past unrelated implementations and should be removed altogether but haven't taken the chance to fully look into them. Updating them is definitely in the plans.

server/graphql-schema.js
```diff
@@ -685,6 +685,10 @@ const graphqlSchema = gql`
         found on the TestPlanVersion type.
         """
         testResults: [TestResult]!
+        """
+        The number of completed tests for this TestPlanRun.
+        """
+        testResultsLength: Int!
```
alflennik, line 691: Maybe explain why this is needed in the docs, i.e. the tests take a long time to be returned

server/resolvers/TestPlanReport/runnableTestsLengthResolver.js
```diff
@@ -0,0 +1,9 @@
+const runnableTestsLengthResolver = async testPlanReport => {
+    // Custom column added with sequelize.literal provides an array as such
+    // runnableTestsCount: [ x, y, z] where x=jawsCount, y=nvdaCount, z=voCount
+    return testPlanReport.testPlanVersion.dataValues.runnableTestsCount[
+        testPlanReport.atId - 1
+    ];
```
alflennik, line 6: Will this work with `.dataValues` removed? Like: `testPlanReport.testPlanVersion.runnableTestsCount[testPlanReport.atId - 1]`.

alflennik (approved):
I saw Erika's comment and I wanted to look too! And it looks nice!

server/graphql-schema.js
howard-e: Sure, can do

server/resolvers/TestPlanReport/runnableTestsLengthResolver.js
howard-e: It doesn't seem to work whenever it's through an association like `testPlanVersion` is here. Appreciate anything extra you think I could try there.

PR #444, opened 2022-07-12
Improve performance by changing Postgres JSONB column type

s3ththompson:
This PR converts database columns that were of type `JSONB[]` to `JSONB`. On my local machine, with a dump of the production database, this improves page load speed from ~10s to 1.5s.

---

(These charts were produced with `yarn dev-debug` and the Chrome Node.js DevTools CPU Profiler set to "Chart" view.)

## Before
Screen Shot 2022-07-12 at 2 06 15 AM

## After
Screen Shot 2022-07-12 at 2 06 24 AM

## Problem
Screen Shot 2022-07-12 at 2 06 58 AM

I noticed that 8.5s of total load time was spent inside a single function, `parse()` from the `postgres-array` module.

The problem is that Postgres returns a string for every column. Our JSONB[] columns, looked like this `{"{\"id\": \"1\"}","{\"id\": \"2\"}"}`. Then, the node `pg` client had to parse the string into a JavaScript array with `parse()` from `postgres-array` (e.g. `"{1,2,3}"` to `[1,2,3]`). Unfortunately, the only way to do this was to scan the entire string character by character. This was pathologically slow for our JSONB[] columns which contained all the tests for a test plan in a single row.

The fix is to convert `TestPlanVersion.tests` and `TestPlanRun.testResults` from JSONB[] columns to `JSONB`. Since arrays are valid JSONB, our `tests` column can just look like `"[{id: 1}, {id: 2}]"` and the `pg` client can just use the native `JSON.parse` directly to turn the string into a JavaScript array. Fortunately, this means that the fix also doesn't touch any additional logic in the app, beyond the model definitions.

### DIFF
--- a/server/migrations/20220712044433-changeJsonbArraysToJsonb.js
+++ b/server/migrations/20220712044433-changeJsonbArraysToJsonb.js
@@ -1,0 +1,86 @@
+'use strict';
+
+module.exports = {
+    up: (queryInterface, Sequelize) => {
+        return queryInterface.sequelize.transaction(async (transaction) => {
+            // https://stackoverflow.com/a/45231776
+            await queryInterface.sequelize.query(
+                `ALTER TABLE "TestPlanVersion" ALTER COLUMN "tests" DROP DEFAULT;`,
+                { transaction }
+            );
+            await queryInterface.changeColumn(
+                'TestPlanVersion',
+                'tests',
+                {
+                    type: 'JSONB USING to_jsonb(tests)',
+                },
+                { transaction }
+            );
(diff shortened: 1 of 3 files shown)


howard-e (approved):
Having tested this locally and on sandbox, can verify everything is looking great!
Massive change towards the performance improvements, thanks @s3ththompson!

s3ththompson:
@alflennik @howard-e thank you! it's great to be able to combine this with all of the other perf work and howard's latest PR—a multi-pronged approach!

PR #446, opened 2022-07-12
Fix report urls

alflennik:
Applies an existing fix from the test queue page to the reports page as well. A more permanent solution which fixes the urls within the database is still needed.


howard-e (approved):
Looks good! Looking forward to what comes after!

PR #448, opened 2022-07-14
Update loading indicators and error feedback

howard-e:
* Added loading indicator to the `Test Reports` and `Settings` pages.
* Updated loading indicator to the `Test Queue` page.
* Added loading indicators and error feedback for several actions on the `Test Queue` page.
* Added error feedback for the `Test Reports` and `Test Queue` page queries


alflennik (commented):
I tested the reports page, the test queue, assigning yourself to a test plan, adding AT versions, removing AT versions, adding test plans and removing test plans. The modals are working.

I'm going to leave an optimistic bit of feedback, which is that I was wondering if the modals could have a threshold time of say 200-300 milliseconds, and if the request loads before then, it won't show the loading modal.

However I realize we can fairly easily remove the modals if the performance gets to a state where they're not needed. So let me know if you prefer to go that route.

I noticed the UI is implemented differently for each loading state, and I wonder if it would be possible -- if we are planning on keeping these permanently -- to have a single loading abstraction that we can feed a promise or something. That would have the benefit of having a single UI experience for all loading which could be more comfortable for the users. But again, perhaps these loading states are not meant to be permanent.

I think we could merge, I mean, I did test it and it's all working as designed which is awesome. Perhaps my feedback perhaps is getting more at a future iteration.

howard-e:
Thanks for the feedback @alflennik!

> However I realize we can fairly easily remove the modals if the performance gets to a state where they're not needed. So let me know if you prefer to go that route.

I think leaving in the loading indicators even if we get there is okay - in instances of times of low connectivity problems with someone's device for example.

> I noticed the UI is implemented differently for each loading state, and I wonder if it would be possible -- if we are planning on keeping these permanently -- to have a single loading abstraction that we can feed a promise or something. That would have the benefit of having a single UI experience for all loading which could be more comfortable for the users. But again, perhaps these loading states are not meant to be permanent.

Created a custom loading wrapper component + custom hook utility which should make it easier to inject into any component. Let me know your thoughts again when you can.

howard-e:
@alflennik also included your changes from #449 to test the change in CI and seems to build properly so went ahead and approved!

evmiguel (approved):
@howard-e code looks good! I like the use of `useTriggerLoad` and loading animation. I've tested out the different loading scenarios; they look good. Nice use of the loading modals as well. Great job!

howard-e:
Thanks for the feedback @evmiguel!

PR #450, opened 2022-07-26
Unexpected behavior note

alflennik:
Makes the unexpected behavior note available for all unexpected behaviors instead of just "Other."

Goes along with a PR in the ARIA-AT repo which is here: https://github.com/w3c/aria-at/pull/802 The two PRs can be merged in any order and at different times if needed. But following review changes to one might require changes to the other.

### DIFF
--- a/client/resources/aria-at-harness.mjs
+++ b/client/resources/aria-at-harness.mjs
@@ -590,15 +590,18 @@ function renderVirtualResultsTable(results) {
    * @param {object} list
    * @param {Description} list.description
    * @param {Description[]} list.items
+   * @param {String} [list.note]
    */
-  function commandDetailsList({ description, items, note }) {
+  function commandDetailsList({
+    description,
+    items,
+    note: { value: noteValue = '', description: noteDescription } = {},
+  }) {
     return div(
       description,
       ul(
-        ...items.map(
-          description => li(rich(description)),
-          note.length ? li(rich('Explanation:'), em(note)) : fragment()
-        )
+        ...items.map(description => li(rich(description))),
+        noteValue.length ? li(rich(noteDescription), ' ', em(noteValue)) : fragment()
       )
     );
   }
--- a/client/resources/aria-at-harness.mjs
+++ b/client/resources/aria-at-harness.mjs
@@ -1,7 +1,16 @@
-import {element, fragment, property, attribute, className, style, focus, render} from "./vrender.mjs";
-import {userCloseWindow, userOpenWindow, WhitespaceStyleMap} from "./aria-at-test-run.mjs";
-import {TestRunExport, TestRunInputOutput} from "./aria-at-test-io-format.mjs";
-import {TestWindow} from "./aria-at-test-window.mjs";
+import {
+  element,
+  fragment,
+  property,
+  attribute,
+  className,
+  style,
+  focus,
+  render,
+} from './vrender.mjs';
+import { userCloseWindow, userOpenWindow, WhitespaceStyleMap } from './aria-at-test-run.mjs';
+import { TestRunExport, TestRunInputOutput } from './aria-at-test-io-format.mjs';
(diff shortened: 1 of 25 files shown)


server/resolvers/TestResultOperations/saveTestResultCommon.js
```diff
@@ -95,23 +94,31 @@ const assertTestResultIsValid = newTestResult => {
...
-            (!!unexpectedBehavior.otherUnexpectedBehaviorText &&
-                unexpectedBehavior.id !== 'OTHER') ||
-            (!unexpectedBehavior.otherUnexpectedBehaviorText &&
-                unexpectedBehavior.id === 'OTHER')
-        ) {
+    const checkUnexpectedBehavior = (
+        unexpectedBehavior,
+        unexpectedBehaviorNote
+    ) => {
+        if (!unexpectedBehaviorNote && unexpectedBehavior.id === 'OTHER') {
```
evmiguel, line 101: Non-blocking: Are we still expecting a behavior note to be filled when OTHER? I thought the note would be decoupled from all undesirable behaviors.

client/resources/aria-at-test-io-format.mjs
```diff
@@ -460,8 +473,8 @@ class UnexpectedInput {
   static fromBuiltin() {
     return new UnexpectedInput({
       behaviors: [
-        ...UNEXPECTED_BEHAVIORS.map(description => ({description, requireExplanation: false})),
-        {description: "Other", requireExplanation: true},
+        ...UNEXPECTED_BEHAVIORS.map(description => ({ description, requireExplanation: false })),
+        { description: 'Other', requireExplanation: true },
```
evmiguel, line 477: Same question here. Is Other supposed to be coupled to the explanation?

client/resources/aria-at-harness.mjs
```diff
@@ -371,47 +393,69 @@ function renderVirtualInstructionDocument(doc) {
...
-                  )
+            label(
+              forInput(`${failOption.description}-${commandIndex}`),
+              rich(failOption.description)
+            ),
+            br()
+          )
+        ),
+        fragment(
+          // TODO: Figure out why this isn't appearing
```
evmiguel, line 443: Looks like this is appearing!

evmiguel (changes requested):
Looking good! There's a bug in the form validation. 

https://user-images.githubusercontent.com/2789098/184236046-840ae019-fada-4e5f-be99-bd0b0fc83652.mov

UPDATE: Actually, the bug persists when switching between No and Yes.

client/resources/aria-at-test-run.mjs
```diff
@@ -207,39 +209,41 @@ export function instructionDocument(resultState, hooks) {
           {
             required: true,
             highlightRequired: resultStateCommand.atOutput.highlightRequired,
-            description: "(required)",
+            description: '(required)',
```
evmiguel, line 212: I'm not seeing this. How do I get to this state?

client/resources/aria-at-test-run.mjs
```diff
@@ -263,54 +268,73 @@ export function instructionDocument(resultState, hooks) {
...
+              {
+                required: resultUnexpectedBehavior.behaviors.some(
+                  ({ checked, requireExplanation }) => requireExplanation && checked
+                ),
+                highlightRequired:
+                  resultState.currentUserAction === 'validateResults' &&
+                  resultUnexpectedBehavior.behaviors.some(
+                    ({ checked, requireExplanation }) => requireExplanation && checked
+                  ),
+                description: resultUnexpectedBehavior.behaviors.some(
```
evmiguel, line 319: I'm not seeing this. How do I get to this state?

alflennik:
Superseded by https://github.com/w3c/aria-at-app/pull/596

PR #451, opened 2022-07-28
Add vendor role

evmiguel:
This PR includes:

- Backend support for vendor role
- New vendors.txt for participating vendors
- Fake role support for vendor role

To test:

- Seed the database: `yarn sequelize db:seed:all`
- Sign out of the app 
- Go to localhost:3000/?fakeRole=vendor
- Hit 'Sign in'
- Observe that the Setting Page is no longer available and Test Queue page is read only

### DIFF
--- a/client/components/TestQueueRow/index.jsx
+++ b/client/components/TestQueueRow/index.jsx
@@ -48,7 +48,7 @@ const TestQueueRow = ({
         false
     );
 
-    const { id, isAdmin, isVendor, username } = user;
+    const { id, isAdmin, isTester, isVendor, username } = user;
     const {
         id: testPlanReportId,
         testPlanVersion,
@@ -60,7 +60,7 @@ const TestQueueRow = ({
 
     const checkIsTesterAssigned = username => {
         return draftTestPlanRuns.some(
-            testPlanRun => testPlanRun.tester.username === username && !isVendor
+            testPlanRun => testPlanRun.tester.username === username && isTester
         );
     };
 
--- a/client/components/TestQueueRow/index.jsx
+++ b/client/components/TestQueueRow/index.jsx
@@ -48,7 +48,7 @@ const TestQueueRow = ({
         false
     );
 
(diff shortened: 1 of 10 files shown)


client/components/TestQueueRow/index.jsx
```diff
@@ -60,7 +60,7 @@ const TestQueueRow = ({
 
     const checkIsTesterAssigned = username => {
         return draftTestPlanRuns.some(
-            testPlanRun => testPlanRun.tester.username === username
+            testPlanRun => testPlanRun.tester.username === username && !isVendor
```
evmiguel, line 63: This is a case I ran into being from assuming tester and vendor roles.

howard-e: Just getting a thought out here and absolutely not currently recommending anything to action on.

Could `&& isTester` work as a better conditional check here and elsewhere within this component? It somewhat reads better in my head when considering the name of the function here and the condition being evaluated.

Additionality, I'm just thinking that currently we assume a vendor can never be a tester, but this could change if there is ever a reasonable request for such. This is in the same way all admins are testers (at least at the database level through `AuthController.js`, currently -- and this could probably be made explicit in `evaluateAuth.js` as well, as an additional safeguard).

This could also future-proof this component and others for tester-based functionality, SHOULD additional roles ever be added which may or may not need to have tester-based functionality as well. It could potentially prevent having to write out additional logic to support and debugging easier.

howard-e (approved):
@evmiguel this looks good to me and a definitely a good start to Vendor role support in the app!

client/components/TestQueueRow/index.jsx
evmiguel: I added the `isTester` for better readability. For now, since `isVendor` is being used in a few places, I'll leave the usage alone. Based on conversations in Slack, it seems that there won't be additional roles, and I'd like not to over-engineer anything at this point. I do agree with the future proofing in the case of additional roles - just seems like we're not there yet.

PR #452, opened 2022-08-01
Update JAWS Test Instructions

howard-e:
Updated JAWS test instructions regarding exiting Forms Mode based on https://github.com/w3c/aria-at/issues/731#issuecomment-1158078906 and https://github.com/w3c/aria-at/issues/731#issuecomment-1167912369.

When being displayed on the Test Run pages, the instructions will now read as:
> 2. ...
> 3. Verify the Virtual Cursor is active by pressing Alt+Delete. If it is not, exit Forms Mode to activate the Virtual Cursor by pressing Escape.
> 4. ...

instead of:
> 2. ...
> 3. Verify the Virtual Cursor is active by pressing Alt+Delete. If it is not, turn on the Virtual Cursor by pressing Insert+Z.
> 4. ...

This is also related to https://github.com/w3c/aria-at/pull/794.

### DIFF
--- a/server/migrations/20220801210224-updateJawsTestInstructions.js
+++ b/server/migrations/20220801210224-updateJawsTestInstructions.js
@@ -1,0 +1,83 @@
+'use strict';
+
+module.exports = {
+    up: queryInterface => {
+        return queryInterface.sequelize.transaction(async transaction => {
+            await queryInterface.sequelize.query(
+                `update "TestPlanVersion"
+                        set tests = update_data.updated_tests
+                        from (select id,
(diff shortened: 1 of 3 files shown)


server/migrations/20220801210224-updateJawsTestInstructions.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
```
evmiguel, line 1: Wow! Your use of the json functions is very clean and easy to read!

evmiguel (approved):
Looks good @howard-e!

PR #455, opened 2022-08-11
Candidate modals

evmiguel:
## Summary
- Upgrade storybook from version 5 to 6
- Implement the modals found in [these wireframes](https://xd.adobe.com/view/47249b7e-a7f3-4ac7-807c-6dd96e37342b-1277/screen/95eea079-2dc3-47b8-b4e2-c0c07f6a5f21/specs/)

## To run
- Update storybook: `yarn`
- Open storybook server: `yarn storybook`
- Observe the modals when server opens


client/components/common/CandidateModals/ProvideFeedbackModal/ProvideFeedbackModal.css
```diff
@@ -0,0 +1,63 @@
...
+    padding: 0px 30px 0px 40px;
+}
+
+.feedback .modal-footer {
+    padding-right: 30px;
+}
+
+.feedback-bold {
+    font-weight: 600;
+    font-style: 'Roboto';
```
howard-e, line 39: Think you meant `font-family` here as 'Roboto' isn't a valid property for `font-style`.
But alternatively, I don't think we've been overriding the default font-family applied for other components, is that intentional here?

client/components/common/CandidateModals/ProvideFeedbackModal/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+    };
+
+    return (
+        <BasicModal
+            show={show}
+            actionLabel={'Submit Review'}
+            centered={true}
+            content={
+                <div className="feedback-content">
+                    <p>
```
howard-e, line 53: Should this still be shown when 0 issues have been raised? What would it link to?

client/components/common/CandidateModals/ProvideFeedbackModal/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+                        You have raised{' '}
```
howard-e, line 54: nit: the mockup also includes "raised" in the link.

client/components/common/CandidateModals/ProvideFeedbackModal/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+                        <Form.Group>
+                            <FormCheck>
+                                <FormCheck.Input
+                                    name="radio-feedback"
+                                    id="approve-input"
+                                    type="radio"
+                                    onClick={radioChange}
+                                    className={approveInputDisabled}
+                                />
+                                <FormCheck.Label
```
howard-e, line 72: I think this label and the others are missing a `for="<id>"` (or `htmlFor` in this case) to point to `<radio>`'s id.

client/components/common/CandidateModals/ProvideFeedbackModal/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+                                rows={5}
+                            />
+                        </Form.Group>
+                    </Form>
+                </div>
+            }
+            dialogClassName="feedback"
+            handleAction={handleAction}
+            title={
+                <div className="feedback-title">
```
howard-e, line 156: VO describes this modal header as having 10 items and stops. The text isn't read. I haven't tested on NVDA/JAWS as yet but this may be worth looking into.

howard-e (commented):
@evmiguel great job recreating the mockups and excited that you've got storybook back into the fold! 👍🏽

client/.storybook/preview.js
```diff
@@ -0,0 +1,4 @@
+import '../../node_modules/bootstrap/dist/css/bootstrap.css';
+import '../../node_modules/@fortawesome/fontawesome-svg-core/styles.css';
+import '!style-loader!css-loader!sass-loader!../scss/custom.scss';
```
alflennik, line 3: Maybe drop a comment to explain the use of this nonstandard syntax?

client/.storybook/main.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-    addons: ['@storybook/addon-actions/register']
+    stories: ['../stories/*.stories.jsx']
```
alflennik, line 2: Thanks for getting Storybook working, this should definitely come in handy in the future!

alflennik: Could you document the command to run storybook in the readme? Just the `yarn storybook` so people know it's there.

alflennik (commented):
Looking good, I just had some minor comments, and thanks for setting up storybook, that really makes these components easy to review, which is awesome.

client/components/common/CandidateModals/ProvideFeedbackModal/index.jsx
alflennik: Nice catch Howard, that's a good point.

client/components/common/CandidateModals/ProvideFeedbackModal/ProvideFeedbackModal.css
evmiguel: We don't need this - good catch!

client/components/common/CandidateModals/ProvideFeedbackModal/index.jsx
evmiguel: True. @isaacdurazo when 0 issues have been raised, should we remove this text? Remove the link perhaps?

isaacdurazo: Correct, if no issues have been raised, we shouldn't show any text. So the "You have raised X issues for this Test Plan" text should not appear

client/components/common/CandidateModals/ProvideFeedbackModal/index.jsx
```diff
@@ -0,0 +1,190 @@
...
+                            />
+                        </Form.Group>
+                    </Form>
+                </div>
+            }
+            dialogClassName="feedback"
+            handleAction={handleAction}
+            title={
+                <>
+                    <p className="sr-only">
```
evmiguel, line 163: @howard-e @louis4533 I made this portion screen reader accessible only and the div with the spans hidden to handle the h1 not being read. Is there a better way to do this?

client/.storybook/main.js
evmiguel: All set!

alflennik:
"consumed" 😹

Issue #437, opened 2022-06-23
Mockups for Candidate Test Plan Review


isaacdurazo:
# Candidate Test Plan Review Process - AT Developer User Journey

## Candidate Test Plans Queue

This is the page where an AT Developer begins a Test Plan review process. It has a table listing all Test Plans in the Candidate phase as well as the percentage of passing tests for each AT and Browser combination the Test Plans were run with.

###  UI Characteristics and details

*  **Heading:** Comprised of a `<h1>` with the name of the page. I'm using "Candidate Test Plans" as a placeholder.

* **Introduction:** Comprised of an `<h2>` with the word "Introduction" and a `<p>` with a short description of the page 

* **Candidate Tests table:** This table has one column for Candidate Test Plans and one column for every AT and Browser combination used.

  *  **Candidate Test Plans column:** the header cell of this column proposes a tabbed navigation, where the AT Developer can switch between Candidates and Recommended Test Plans. These two options are preceded by the number of Test Plans available under each. E.g. "5 Candidates - 3 Recommended".
  The "active" tab in this header cell is in bold, while the inactive one is in normal text. The Test Plan names in this column are links and next to them there is a label. There are three different labels depending on the step in which the review process is at:
 
     * **Ready for Review:** When a Test Plan reaches the Candidate phase, it would automatically be added to this page and a "Ready for Review" label is displayed next to its name

     * **X number of open issues:** If an AT Developer requests changes for a Test Plan, that request culminates in the creation of a Github Issue. This is also reflected in this label so users can see how many issues there are for a given Candidate Test Plan

     * **Ok:** once an AT developer approves a Candidate test plan an "Ok" label is displayed next to its name

  *  **AT and Browser column:** there is a column for every AT and Browser combination available. The cell contents of these columns could either be the word "None" if there are no test results, or a progress bar showing the percentage of passing tests for a given Test Plan under a given AT and Browser combination.

![ARIA-AT - Candidate Test Plans Queue](https://user-images.githubusercontent.com/1379244/175362997-e83424bc-be2f-4c48-a33d-7ea6cb49ea9c.png)

---

## Single Candidate Test Plan 

Once the AT Developer selects a Candidate Test Plan they would land on this page. Here we have a header containing UI elements to display different statuses and actions the user can utilize for reviewing. The main content of this page has a table listing every Assistive Technology used for every Candidate Test Plan and a column for each Browser used for each AT. 

###  UI Characteristics and details

[...]

PR #459, opened 2022-08-29
GitHub Issues Sync for Candidate Review Flow server implementation

howard-e:
Based on designs proposed for #437 as it relates to retrieving issues from the [repository's issue page](https://github.com/w3c/aria-at-app/issues) which have been created through ARIA-AT App to map to Candidate Review feedback.

When pulling the issues data, it has to have the `app` and `candidate-review` labels. It must also have the AT's label (`jaws`, `nvda` or `vo`). The app also expects an issue type to be present. Either `changes-requested` or `feedback` (if no label exists for the type, `feedback` is assumed). [Example of issues](https://github.com/howard-e/aria-at-app/issues?q=is%3Aissue+label%3Aapp+candidate-review+).

This also adds migrations that are required to know when the Phase of the Test Plan Report was last updated. A resolver has also been added to find the target date based on the [Working Mode](https://github.com/w3c/aria-at/wiki/Working-Mode#candidate-test-plan-review).


warp2nowhere (approved):
Looks good! Very excited to use this feature.

evmiguel (approved):
@howard-e the user @warp2nowhere is my testing account for vendors. Sorry for putting the review with that account. Must have been confusing!

server/graphql-schema.js
```diff
@@ -740,6 +740,30 @@ const graphqlSchema = gql`
...
+        """
+        type: String!
+        """
+        Indicates if the issue is currently open on GitHub.
+        """
+        isOpen: Boolean!
+        """
+        Test Number the issue was raised for.
+        """
+        testNumber: Int!
```
evmiguel, line 764: Is this the same as the rowNumber?

server/tests/integration/graphql.test.js
```diff
@@ -346,6 +348,11 @@ describe('graphql', () => {
                                 locationOfData
                             }
                         }
+                        phaseChangeUpdate
+                        phaseTargetDate
+                        issues {
+                            __typename
+                        }
```
alflennik, line 355: Might be nice to figure out a way to populate these in tests, possibly through a mock service or something?

alflennik (commented):
Looking great. Great job with the GitHub architecture, and the way you implemented caching was super clean. The most major comment worthy of further discussion is probably whether the candidate and recommended phases should apply to TestPlanVersions or TestPlanReports. Maybe @s3ththompson can weigh in on that?

server/graphql-schema.js
howard-e: This is slightly different to `rowNumber` in that `rowNumber` refers to the sequence number of the test from the original set of tests (the .csv file). Whereas the `testNumber` here is referring to the sequence number in the set of tests grouped by the AT if that helps to clear up anything here.

server/graphql-schema.js
```diff
@@ -740,6 +740,30 @@ const graphqlSchema = gql`
...
+        author: String!
+        """
+        Link to the GitHub issue's first comment.
+        """
+        link: String!
+        """
+        Indicates the type of issue. 'changes-requested' or 'feedback'.
+        'feedback' is the default type.
+        """
+        type: String!
```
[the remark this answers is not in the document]
howard-e: Not sure I follow this. It sounds like you're saying we should be able to expose all the labels retrieved from the GitHub issue through the GraphQL query. We're currently making use of the other labels within the resolver but I'm not what the need would be to expose all of them. Only the `changes-requested`/`feedback` label seem to matter to the client based on the latest designs.

Could you perhaps expand on what you meant here?

server/graphql-schema.js
```diff
@@ -754,6 +778,15 @@ const graphqlSchema = gql`
...
         """
+        Date of when the status was last updated.
+        """
+        phaseChangeUpdate: Timestamp!
+        """
+        The intended target date for the final TestPlanReport phase promotion.
+        Based on the ARIA-AT Working Mode.
+        """
+        phaseTargetDate: Timestamp
+        """
```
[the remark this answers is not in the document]
howard-e: > * The fact that we're calling the status both "status" and "phase" might be worth trying to avoid.

Agreed!

> * It seems conceivable that we will want to permanently save the date when each status was reached, even after the test plan has reached a recommended status (or a future deprecated status).

Makes sense.

> * I don't think the target date needs to be persisted since it's simply six months after the candidate phase starts. And I can imagine that in the real world, the date the phase is officially changed might differ from the exact six months mark, like in the case that a vendor puts in a bunch of feedback on the last day or something.

Well it isn't currently being persisted on the DB but derived from the `candidateStatusReachedAt/phaseChangeUpdate` but I do see a need where it needs to be, based on the latest designs just in terms of providing expected functionality. If it's not persisted, it would shifting the value of the `candidateStatusReachedAt` variable which doesn't seem correct to me.

> Anyway these considerations make me think of something like `candidateStatusReachedAt` and `recommendedStatusReachedAt`.

I like those!

server/tests/integration/graphql.test.js
howard-e: Definitely! Perhaps we could discuss this further as well

server/graphql-schema.js
alflennik: Ah, that's a very handy explanation. In that case I would probably vote for a more wordy variable name or at the very least putting an explanation in the documentation string. Maybe a name like `testNumberFilteredByAt`.

server/services/GithubService.js
```diff
@@ -21,9 +22,76 @@ const permissionScopes = [
...
+                      )
+                  ]
+                : []
+        };
+    }
+
+    // Check if additional pages exist
+    if (headersLink && headersLink.includes('rel="next"')) {
+        // Get result from other pages
+        const additionalResultsByAt = await constructIssuesRequest({
```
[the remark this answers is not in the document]
alflennik: Recursion never doesn't deserve applause 👏😜

server/graphql-schema.js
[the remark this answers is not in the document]
alflennik: Yeah, I was suggesting having an array of labels which would match how it's presented in GitHub. Then the client code would go something like `if (issue.labels.includes('changes-requested'))`.

If you think that's a bit too indirect, we could keep a single field but I would probably suggest having the word "label" in there since that's where the field is originating in GitHub. So `feedbackLabel` or something like that. I'm wary of the word "type" just because it gets thrown around a lot and doesn't have a lot of meaning in itself.

server/graphql-schema.js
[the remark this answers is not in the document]
alflennik: Yay great!

server/tests/integration/graphql.test.js
alflennik: Yes good idea I have no idea right now how to do it lol

s3ththompson:
> The most major comment worthy of further discussion is probably whether the candidate and recommended phases should apply to TestPlanVersions or TestPlanReports. Maybe @s3ththompson can weigh in on that?

@howard-e @alflennik I think in the short term, the phase can live on the TestPlanReport, but the outcome of the design workshop (and the mockups that Isaac is working on) is that the phase really belongs to Combobox itself, not Combobox v1234, and not Combobox for JAWS. I think this will probably mean introducing a new schema definition called TestPlan that is the parent of many TestPlanVersions.

s3ththompson:
But that change would come at a later time, after we have more time to discuss implications.

alflennik (approved):

PR #461, opened 2022-09-21
Candidate test run page

evmiguel:
This PR implements the Candidate Test Run Page and feedback modals as specified in [this mockup](https://github.com/w3c/aria-at-app/issues/437) and [this design](https://xd.adobe.com/view/1f606837-32ce-43f9-aca4-65cb99a53e09-4a33/screen/6eb3c478-1e93-488c-b28b-30d037737c3f/).

0. Run `yarn` to install dependencies

Instructions for dev server:
1. Run `yarn sequelize db:migrate`
2. Run the app, and sign out if signed in.
3. Sign in using the following `http://localhost:3000/?fakeRole=vendor`
4. Access the page by going to `http://localhost:3000/candidate-tests` and selecting one of the test plans


client/components/TestQueueRow/index.jsx
```diff
@@ -209,7 +209,7 @@ const TestQueueRow = ({
                 </>
```
evmiguel, line 209: This allows admins to keep the admin view while also assuming the vendor role.

server/resolvers/TestPlanReport/issuesResolver.js
```diff
@@ -30,8 +30,13 @@ const issuesResolver = async testPlanReport => {
                 html_url,
```
evmiguel, line 30: This is for the case of general feedback, where the title does not include `[Test #]`

server/graphql-schema.js
```diff
@@ -762,15 +766,15 @@ const graphqlSchema = gql`
         Indicates the type of issue. 'CHANGES_REQUESTED' or 'FEEDBACK'.
         'FEEDBACK' is the default type.
         """
-        feedbackType: IssueFeedbackType!
```
evmiguel, line 765: Is this supposed to be IssueFeedbackType or String?

client/components/CandidateTestPlanRun/CandidateTestPlanRun.css
```diff
@@ -0,0 +1,154 @@
...
+    left: -1.5em;
+    background: #d2d5d9;
+}
+
+.feedback-list-item .feedback-indicator {
+    background: white;
+}
+
+.issues-container {
+    max-width: 800px;
```
howard-e, line 87: A suggestion here is a `width: -14px;` should be good enough to close the gap at the bottom of the container to get it closer to the design. That's the quickest fix I'm seeing.

server/graphql-schema.js
howard-e: Yep, this is supposed to be `IssueFeedbackType`. It points to the enum defined on L747 which may have gotten ignored during some previous merge?

server/graphql-schema.js
```diff
@@ -762,15 +766,15 @@ const graphqlSchema = gql`
...
+        feedbackType: String!
```
howard-e, line 769: I'd rather moving back to using the `IssueFeedbackType` enum so it's clear what values can be used. 
Unless you had another idea in mind? If so then no problem and the enum can be deleted.

server/resolvers/TestPlanReport/issuesResolver.js
howard-e: Good catch!

howard-e (changes requested):
@evmiguel great work with adding in the Candidate Test Run page. This is so well done and certainly a heavy lift! 

I've made several notes but the truly breaking piece is around the newly introduced `Test.viewers` property. I've prepended the suggestions there with 'BREAKING:' to find the comments faster. But alternatively, making sure that property is not nullable and must at least return an empty array through GraphQL also works.

Also of note is a screen reader related issue related to the accordion components being used on the Candidate Test Run page but it could also be an issue isolated to me somehow so please let me know your thoughts there.

client/components/CandidateTestPlanRun/queries.js
```diff
@@ -0,0 +1,152 @@
...
+            statuses: [FINALIZED]
+            testPlanVersionId: $testPlanVersionId
+            atId: $atId
+        ) {
+            id
+            candidateStatusReachedAt
+            recommendedStatusTargetDate
+            vendorReviewStatus
+            issues {
+                author
```
howard-e, line 44: The mockup also includes a hyperlink on the text which says something like `x people have left feedback for this test`. My guess is that should be to the issue(s) (although I'm not sure how to account for multiple links visually, a pop up or so of applicable links? But this may be something to discuss internally).

But note that `link` is available in the query to get access to those GitHub issue links

client/components/CandidateTestPlanRun/CandidateTestPlanRun.css
evmiguel: A bit confused here. What does a negative width value mean?

client/components/common/CandidateModals/ProvideFeedbackModal/ProvideFeedbackModal.css
```diff
@@ -0,0 +1,70 @@
+.btn {
```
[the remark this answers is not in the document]
evmiguel: Good catch! Thank you!

client/components/CandidateTestPlanRun/index.jsx
```diff
@@ -0,0 +1,698 @@
...
+        const githubUrl = `https://github.com/w3c/aria-at-app/issues/new?title=${encodeURI(
+            githubIssueUrlTitle
+        )}&labels=${defaultGithubLabels},${githubAtLabelMap[at]}`;
+
+        return `${githubUrl},${type}`;
+    };
+
+    const requestChangesUrl = generateGithubUrl(true, 'changes-requested');
+    const feedbackUrl = generateGithubUrl(true, 'feedback');
+    const fileBugUrl = generateGithubUrl(true, 'bug');
```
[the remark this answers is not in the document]
evmiguel: Ah, ok! Is this something that we should in this PR? @s3ththompson

client/components/CandidateTestPlanRun/index.jsx
```diff
@@ -0,0 +1,698 @@
...
+                    }
+                })}
+            </ul>
+        </div>
+    );
+
+    const results = (
+        <>
+            <h1 className="current-test-title">{currentTest.title}</h1>
+            <Accordion className="feedback-accordion" defaultActiveKey="0">
```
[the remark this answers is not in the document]
evmiguel: I think standardizing the component would be great! (Is is not an accordion?) I interpreted the mockups as an accordion, but I'm happy to move the content to the standardized component. Is that something you would like to do?

client/components/CandidateTestPlanRun/queries.js
evmiguel: @s3ththompson @isaacdurazo how should we account for the multiple links in the case of `x people have left feedback for this test`?

client/components/CandidateTestPlanRun/CandidateTestPlanRun.css
howard-e: Apologies there! It should have been `margin-bottom: -14px` but please take a look at playing around with that value for the best fit!

client/components/CandidateTestPlanRun/queries.js
evmiguel: It is possible. We'd just have to map a link with the right query to the component.

client/components/CandidateTestPlanRun/index.jsx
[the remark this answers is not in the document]
evmiguel: Per conversation, the changes requested and leave feedback go to GitHub and the bugs go to the issue tracker.

server/controllers/AuthController.js
```diff
@@ -60,7 +60,10 @@ const oauthRedirectFromGithubController = async (req, res) => {
         roles.push({ name: User.TESTER }); // Admins are always testers
     }
 
-    if (vendors.filter(vendor => vendor.includes(githubUsername))) {
+    if (
```
howard-e, line 63: Good catch!

howard-e (approved):
@evmiguel looks good! Thanks for taking the time to incorporate some of the feedback

PR #463, opened 2022-10-04
Updater performance

evmiguel:
This PR introduces changes to help the TestPlanUpdater performance. The main change is removing the UPDATER_QUERY in favor of returning the `latestTestPlanVersion` from the server.

### DIFF
--- a/client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
+++ b/client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
@@ -13,7 +13,6 @@ import {
     SAVE_TEST_RESULT_MUTATION,
     SUBMIT_TEST_RESULT_MUTATION,
     TEST_PLAN_ID_QUERY,
-    UPDATER_QUERY,
     VERSION_QUERY
 } from './queries';
 import './TestPlanUpdaterModal.css';
@@ -31,11 +30,7 @@ const TestPlanUpdaterModal = ({
     testPlanReportId,
     triggerPageUpdate
 }) => {
-    const loadInitialData = async ({
-        client,
-        setUpdaterData,
-        testPlanReportId
-    }) => {
+    const loadInitialData = async ({ client, testPlanReportId }) => {
         const { data: currentReportData } = await client.query({
             query: TEST_PLAN_ID_QUERY,
             variables: { testPlanReportId }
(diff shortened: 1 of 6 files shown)


client/components/TestPlanUpdater/TestPlanUpdaterModal.jsx
```diff
@@ -222,7 +221,7 @@ const TestPlanUpdaterModal = ({
             return;
         }
 
-        setNewReport(true);
+        setNewReport(created.find(item => item.testPlanReport.id));
 
         for (const testPlanRun of runsWithResults) {
```
howard-e, line 226: Just a thought here. I'm wondering if there's any potential savings that can be had here if the `testResults` are `CREATE`d then (`SAVE`d or `SUBMIT`ed) locally to the new TestRun that's been created on L227-L233. I suppose a new mutation would be required however to push that batched content to the database.

The (`CREATE`+`SAVE`/`SUBMIT`)`_TEST_RESULT_MUTATION` calls back-to-back, multiplied by _'n'_ test results then multipled by the TestRun (but this is much les expensive) to the DB may be both a DB memory/network bottleneck which may be causing slower speeds in production?

This may also mean some pieces of the TestResult creation process has to be copied to the client side to replicate the encoded IDs properly (which is fine), so this may be worth considering.

howard-e (commented):
@evmiguel the process to evaluate the number of `testResults` for what's being updated is now blazing fast! Seeing some speed changes from ~19s to ~1s locally on the example which may possibly take the longest. Nice!

Just had some additional observations:
* [Blocking as different from previous behavior but okay overlooking as a minor issue for now] When making use of the "Keep older versions of test plan and results" option, after completion, the additional row isn't added to the list of items unless the page is completely refreshed which seems to be a regression (the page refreshed automatically before). Now, the updated one takes the place of the original until refreshed.

* [Non-blocking] This may have been discussed before but it seems like unknown behavior and potential loss of data (especially in the case where it's an in-place update) can happen if the user refreshes/closes/navigates away from `/test-queue`. Not sure if maybe considering having a 'fake' report hold the data on the DB until completed then it's automatically promoted as the actual report being updated or something of the sort is possible? Or maybe batching locally and then pushing when everything has been evaluated?
  * I do like that the requests continue after the modal has been closed though. Would be cool if an indication of that was shown!

* [Non-blocking] Noticed that the speed after agreeing to update the test plan is pretty much the same though so shared some thoughts in an additional comment on the code.

Issue #466, opened 2022-11-01
Candidate Tests Page Feedback


evmiguel:
Please use this issue to document feedback for the Candidate Tests feature. The page can be accessed in sandbox: https://aria-at-app-sandbox.bocoup.com/candidate-tests

Feedback from @mcking65:
- Separate start date into its own column.
- Separate target date into its own column with the change target date button in the same cell as the date.
- Put the change status button in the status column
- Separate “Assertion Pass Rate column. This could allow sorting by pass rate
- Include total assertions and tests in summary, e.g., instead of “10 assertions failed across 4 tests run with 1 browser”, use “10 of 200 assertions failed across 4 of 30 tests in 1 browser”

s3ththompson:
I'm seeing an off-by-one time zone error on the "Mark as target date". Changing the date to "20-11-2022" for example, sets the target date to November 19, 2022. (I suspect this has something to do with dates being converted to timestamps at midnight, which then fall back to the previous day due to a time zone conversion...)

s3ththompson:
From @jscholes:
- The hierarchy of titles on the results summary seem to be inverted:
  - "Test results for Chrome" is just text
  - "Test result: Pass" is an H1
  - Ideally "Test results for Chrome" would be an H1 and "Test Result: Pass" would be an H2
  - Relatedly, the collapsible accordion pattern is not actually a disclosure element. It is not currently tab-focusable or keyboard accessible

s3ththompson:
From prior discussions with @mcking65 and @jscholes:

We should update the issue titles to more closely match the current tester feedback. This will need to be done with close attention with @howard-e's GitHub issue parsing code.

Suggested format: `Apple Feedback: "Navigate to the last breadcrumb link" (Breadcrumb Example, Test 6)`
or `Apple Change Request: "Navigate to the last breadcrumb link" (Breadcrumb Example, Test 6)`

evmiguel:
From @louis4533 

- "Viewing" announcement on current test
- Previous and next navigation should announce current test as well

s3ththompson:
From @jscholes:

- During testing, encountered strange behavior: navigated to last test using “next” button, pressed “finish” button…button disappeared initially, and then went back to “next”
- Suggested copy change: after marking “not approved yet” and returning to index, the status for the test plan still shows “ready for review.” Ideally this should reflect, by some means, that the reviewer made it to the end of the review, cc @isaacdurazo

howard-e:
[...]

PR #473, opened 2022-11-28
Accessibility feedback from #466

evmiguel:
This PR implements the following accessibility changes from @louis4533:

- [x] Maybe consider having them all (disclosures) closed so a user may jump to the test Instructions or browser results for their needs.
- [x] Also, instead of the “Next” button changing to “Finish”, make the “Next” button disabled and add an extra “Finish” button
- [x] Should consider announcing that the user is on the last test

I could use some help with the following:

- [x] “main” menu navigation and “main” landmark being announced separately might be confusing
    - I'm not sure how to implement this
- [x] Number of items in navigation menu list not being announced
    - This seems to be announced in VoiceOver. Is there a better way to make this more explicit?

### DIFF
--- a/client/components/App/App.jsx
+++ b/client/components/App/App.jsx
@@ -39,7 +39,7 @@ const App = () => {
                 <Navbar
                     bg="light"
                     expand="lg"
-                    aria-label="Main Menu"
+                    aria-label="Menu"
                     expanded={isNavbarExpanded}
                     onToggle={() => setIsNavbarExpanded(previous => !previous)}
                 >
--- a/client/components/CandidateTests/CandidateTestPlanRun/index.jsx
+++ b/client/components/CandidateTests/CandidateTestPlanRun/index.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useQuery, useMutation } from '@apollo/client';
 import TestNavigator from '../../TestRun/TestNavigator';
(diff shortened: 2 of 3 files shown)


jscholes:
@evmiguel I pushed a commit to address:

> “main” menu navigation and “main” landmark being announced separately might be confusing

... specifically be changing the accessible name on the nav to just "Menu" without the "Main" prefix.  Regarding:

> Number of items in navigation menu list not being announced

... I can't reproduce this either.  It seems to just be a standard list, and with NVDA and Chrome it is announced as "list of 5 items" (or the appropriate number if signed out).

Hope this helps!

jscholes:
> Also, instead of the “Next” button changing to “Finish”, make the “Next” button disabled and add an extra “Finish” button

This one is potentially concerning to me, specifically as implemented with a `disabled` attribute.  This will mean that, if the user is focused on the Next button and it becomes disabled, keyboard focus will be lost, and e.g. a JAWS user will be returned to the top of the entire page.

Does focus move somewhere when navigating to the next test, to avoid this issue?  If not, and disabling the button is thought to be desirable, `aria-disabled` should be used instead of `disabled`.

louis4533:
@jscholes thanks for mentioning "aria-disabled." My oversight and lack of experience are to blame. @evmiguel the list items are being announced when navigating with the previous and next buttons. "Viewing Test: Navigate backwards to a collapsed disclosure button in reading mode , Test 2 of 30." @jscholes would this be desirable from your point of view?

howard-e (approved):

PR #478, opened 2022-12-22
Embed reports management

evmiguel:
This PR integrates the changes to the new reports page link structure, which accommodates multiple test plan versions.


howard-e (approved):
Looks good!

Used this to also replicate your comparisons on `/reports` as well for using `releasedAt` instead as another option.

PR #483, opened 2023-01-25
Migrate date management from momentjs to dayjs

howard-e:


server/resolvers/TestPlanReport/issuesResolver.js
```diff
@@ -1,14 +1,14 @@
...
+const dayjs = require('dayjs');
 
 const issuesResolver = async testPlanReport => {
     if (!testPlanReport.candidateStatusReachedAt) return [];
 
     const searchPrefix = `${testPlanReport.at.name} Feedback: "`;
     const searchTestPlanVersionTitle =
         testPlanReport.testPlanVersion.dataValues.title;
-    const searchTestPlanVersionDate = moment(
+    const searchTestPlanVersionDate = dayjs(
```
evmiguel, line 11: Non-blocking: Would it be better to use one of the util functions instead of `dayjs` directly?

evmiguel (approved):
One non-blocking question, but it looks good to me overall!

server/resolvers/TestPlanReport/issuesResolver.js
howard-e: Which util functions are you referring to there?

evmiguel: Looks like this is the same as `convertDateToString`

PR #488, opened 2023-02-08
Remove sass dependency

alflennik:
Removes sass, which we don't actually use.


evmiguel (commented):
Looks good, @alflennik! Can you resolve the yarn.lock conflicts before merging?

PR #498, opened 2023-02-15
Updating React Router Dependencies

aleenaloves:
This PR attempts to migrate our react router logic from v5 to v6 based on [this guide](https://reactrouter.com/en/6.8.1/upgrading/v5).

There were certain decisions made in re-writing `client/routes/index.js`. In this case it was opting for descendent routing vs nested routing. 

Certain components were simplified if the new migration made it so that certain things were unnecessary

> For example:

> If components had a `404` route in their descendent routing, this was mostly removed as the parent `routes` object now took care of that. 

But most kept to their original logic

> For example: 

> in certain cases, instead of just navigating to the `404` path, some components opted to use the `invalid-request` path that redirects to the `404` path. This could have been simplified to just navigate to `404,` but without context surrounding why that decision was made originally, that logic was not simplified and instead kept the same.

Other decisions that may require explanation are the following:
- Regular expressions no longer work in v6, but our routes that were utilizing them didn't need them in the first place
- ...anything else you would like more explanations on

Testing Notes:
To test this, you can navigate to the `client/routes/index.js` file and see if all the paths in that file still work + any descendent paths from the elements themselves. 

It is also important to check if the routes still have the same screen-reader behavior as v5.

### DIFF
--- a/client/components/Reports/SummarizeTestPlanVersion.jsx
+++ b/client/components/Reports/SummarizeTestPlanVersion.jsx
@@ -11,6 +11,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faHome } from '@fortawesome/free-solid-svg-icons';
 import styled from '@emotion/styled';
 import DisclaimerInfo from '../DisclaimerInfo';
+import { useParams } from 'react-router-dom';
 
 const FullHeightContainer = styled(Container)`
     min-height: calc(100vh - 64px);
@@ -18,14 +19,18 @@ const FullHeightContainer = styled(Container)`
 
 const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
     const { exampleUrl, designPatternUrl } = testPlanVersion.metadata;
+    let params = useParams();
+
+    if (!params.testPlanReports.length) return <Navigate to="/404" />;
+
     return (
         <FullHeightContainer id="main" as="main" tabIndex="-1">
             <Helmet>
                 <title>
-                    {getTestPlanVersionTitle(testPlanVersion)} | ARIA-AT Reports
+                    {getTestPlanVersionTitle(params.testPlanVersion)} | ARIA-AT Reports
                 </title>
             </Helmet>
-            <h1>{getTestPlanVersionTitle(testPlanVersion)}</h1>
+            <h1>{getTestPlanVersionTitle(params.testPlanVersion)}</h1>
 
             <Breadcrumb
                 label="Breadcrumb"
@@ -40,7 +45,7 @@ const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
                     </Breadcrumb.Item>
                 </LinkContainer>
                 <Breadcrumb.Item active>
-                    {getTestPlanVersionTitle(testPlanVersion)}
+                    {getTestPlanVersionTitle(params.testPlanVersion)}
                 </Breadcrumb.Item>
             </Breadcrumb>
             <h2>Introduction</h2>
@@ -100,7 +105,7 @@ const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
                         <DisclaimerInfo reportStatus={testPlanReport.status} />
                         <LinkContainer
                             to={
-                                `/report/${testPlanReport.testPlanVersion.id}` +
+                                `/report/${params.testPlanVersion.id}` +
                                 `/targets/${testPlanReport.id}`
                             }
                         >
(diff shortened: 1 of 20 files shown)


alflennik:
Thanks for getting us over the react-router v5 hump, I know there have been a lot of changes before v6!

Would it be possible to incorporate the latest merges in the update-dependencies branch, specifically the removal of nodegit? This branch still has nodegit and I can't run yarn install until it's removed 🙏

aleenaloves:
thanks for the heads up @alflennik !

@evmiguel can i merge those changes into this branch?

aleenaloves:
@alflennik this should be merged now, let me know if anything's missing ! thank you !

howard-e:
@aleenaloves looks like this was merged with `main` isntead. Because this is a PR against `update-react`, I think `update-react` should be rebased onto `update-dependencies` so this branch can then be rebased onto `update-react`.

Super excited to these changes updates be included! 🙂

howard-e:
Worked with @aleenaloves in the background to get the branches updated.

Drafting for now until newly introduced breaking changes coming from recent changes around the `/report/...` pages can be accounted for in the updated routing structure.

aleenaloves:
updated conflict + fixed test runs, removing from draft state

evmiguel (commented):
The changes look good, but I am running into an error locally. I have loaded the sandbox data into my database. When I go to http://localhost:3000/report/1090,1858,1890 -> Click the first `View Complete Results`, which is http://localhost:3000/report/1890/targets/114 -> 404.

Does this happen to you?

client/components/Reports/SummarizeTestPlanVersion.jsx
```diff
@@ -100,7 +100,7 @@ const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
                         <DisclaimerInfo reportStatus={testPlanReport.status} />
                         <LinkContainer
                             to={
-                                `/report/${testPlanReport.testPlanVersion.id}` +
```
howard-e, line 103: This change should be reverted.

client/components/Reports/SummarizeTestPlanVersion.jsx
```diff
@@ -111,7 +111,7 @@ const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
                         {skippedTests.length ? (
                             <Link
                                 to={
-                                    `/report/${testPlanReport.testPlanVersion.id}` +
```
howard-e, line 114: And this

client/components/Reports/SummarizeTestPlanVersion.jsx
```diff
@@ -151,7 +151,7 @@ const SummarizeTestPlanVersion = ({ testPlanVersion, testPlanReports }) => {
                                                 <td>
                                                     <Link
                                                         to={
-                                                            `/report/${testPlanReport.testPlanVersion.id}` +
```
howard-e, line 154: And this

howard-e (commented):
Related to @evmiguel's review:

> The changes look good, but I am running into an error locally. I have loaded the sandbox data into my database. When I go to http://localhost:3000/report/1090,1858,1890 -> Click the first View Complete Results, which is http://localhost:3000/report/1890/targets/114 -> 404.

aleenaloves:
@evmiguel 

> The changes look good, but I am running into an error locally. I have loaded the sandbox data into my database. When I go to http://localhost:3000/report/1090,1858,1890 -> Click the first `View Complete Results`, which is http://localhost:3000/report/1890/targets/114 -> 404.
> 
> Does this happen to you?

Yes, this happens to me as well, but it has to do with my database values and not the routing itself. 

In my case it's successfully routing to `/404` because that target does not exist. 

If you run the following query:

``` sql
select * from "TestPlanReport" 
where id = 114
```

Is it attached to `testPlanVersionId` 1890 in your database? 

If not, then this is the correct behavior in terms of the existing logic.

For further verification, when you run this query:
``` sql
select * from "TestPlanReport" 
where "testPlanVersionId" = 1890
```

Does `114` show up? 

If not, then this is the correct behavior in terms of the existing logic.

Although, to double check the routing itself, from my database, we can see that the following should route to something: http://localhost:3000/report/1890/target/120 

But that leads to `/404` as well. Which _could_ be a problem, if the following doesn't sound right to you:

In my case, `1890` does have a status of `DRAFT`, and `1890` does not appear on my `Test Reports` page either, so i would assume that a `DRAFT` test report should redirect to `/404` but i lack context here. 

(ps, i included all this context so that if at any point i deduced this incorrectly, feel free to correct me where i went wrong)

(pps, even though this may not be an issue, it did lead me to discover that during the merge conflicts the check for making sure `testPlanReports` weren't null or 0 also got removed which was causing bugs, so i readded that in as well. thank you !!)

aleenaloves:
reverting what @howard-e mentioned fixed @evmiguel concerns - this should be good to review

evmiguel (approved):

alflennik (approved):
Code looks good and I also ran locally, testing especially the reports pages given the comments. Everything worked for me.

PR #499, opened 2023-02-16
Test Management Page

howard-e:
Reintroducing the Test Management page with minimal features. This is located at `<base_url>/test-management`

The page now includes a `Status Summary` section which administrators can use to move the status of test plans between `DRAFT`, `CANDIDATE` and `RECOMMENDED`. The data being represented is also based on the test plan reports which are using the latest version of their respective ATs.

PR #500, opened 2023-02-21
Candidate Test Plans page row consolidation

howard-e:
This PR aims to merge multiple Test Plan reports into a single row where applicable (the same logic is being applied to the Test Reports page)

**NB. Currently based on #499**

### DIFF
--- a/client/components/TestQueueRow/index.jsx
+++ b/client/components/TestQueueRow/index.jsx
@@ -414,19 +414,28 @@ const TestQueueRow = ({
                     const candidatePhaseList = uniq.map(u => JSON.parse(u));
                     setCandidatePhaseTestPlanReports(candidatePhaseList);
 
-                    // if (candidatePhaseList > 1) {
-                    // There is a test plan report in Candidate Test Plans to be overwritten
-                    setShowCandidatePhaseSelectModal(true);
-                    // }
+                    if (candidatePhaseList.length > 0) {
+                        // There already exists a Test Plan Report which uses this pattern so
+                        // there is already a candidate phase to select from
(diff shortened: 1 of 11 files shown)


client/components/TestQueueRow/index.jsx
```diff
@@ -378,14 +385,107 @@ const TestQueueRow = ({
...
+                    const { data } = await client.query({
+                        query: TEST_PLAN_REPORT_CANDIDATE_RECOMMENDED_QUERY,
+                        fetchPolicy: 'network-only'
+                    });
+
+                    // TODO: Check to see if the proposed test plan report to be promoted to CANDIDATE
+                    //  has a test plan version date less than whatever is in CANDIDATE or RECOMMENDED
+                    //  tests already, if any, and note it may never be displayed to the admin
+
+                    // --- SECTION START: OutdatedCandidatePhaseComparison
```
evmiguel, line 400: Thank you for this documentation!

alflennik (approved):
I can confirm that this works and that the behavior seems pretty logical from what I understand about the workflow, and also I want to note that the new UI that I hadn't seen before is looking pretty nice and seems easy to use. 

Looking at all the imperative-style code, one opportunity here might be to eventually write an integration test which captures the complete end-to-end flow of a test plan through its entire life cycle, because to me it's hard to really feel like I tested it thoroughly enough and caught all the potential edge cases, but I don't think that's a blocker for this PR.

Nice work!

evmiguel (approved):
Thanks for taking care of the Candidate Test Run page and great documentation for the modal fix! This was a big body of work done over a short period of time, so kudos to you!

PR #501, opened 2023-02-21
Update Sequelize ORM from v5 to v6

Paul-Clue:
The purpose of this PR is:

- Upgrade the Sequelize ORM.
- Make sure that the test suite will work in time zones other than those in the US.
- Create a patch that fixes an error that occurs due to a deprecated function in the turndown package.


client/utils/gitUtils.js
```diff
@@ -2,7 +2,7 @@ const lc = (pattern, string) => string.replace(pattern, pattern.toLowerCase());
 const timeZone = 'UTC';
 const options = { month: 'short' };
 
-export const gitUpdatedDateToString = (dateString, locale = 'default') => {
+export const gitUpdatedDateToString = (dateString, locale = 'en-US') => {
```
howard-e, line 5: I don't think this change should be made. It would be forcing users from different regions to view dates in formats they don't expect. Any reason for this change?

patches/domino+2.1.6.patch
```diff
@@ -0,0 +1,38 @@
+diff --git a/node_modules/domino/lib/sloppy.js b/node_modules/domino/lib/sloppy.js
```
howard-e, line 1: I think the problem we were solving with the tests has already been resolved by the `update-dependencies` branch.

I went ahead and removed it locally, re-ran `yarn install` and `yarn test` and the tests were able to successfully run. Could you try the same and remove the patch file if it's all good for you as well @Paul-Clue?

server/models/index.js
```diff
@@ -30,7 +30,10 @@ fs.readdirSync(__dirname)
         );
     })
     .forEach(file => {
-        const model = sequelize['import'](path.join(__dirname, file));
+        const model = require(path.join(__dirname, file))(
```
howard-e, line 33: Nice!

howard-e (commented):
Thanks @Paul-Clue, looks good!

Had 2 minor follow up questions below but other than that, all looks well!

client/utils/gitUtils.js
Paul-Clue: First, thanks!
That change was made because the test kept failing. The formatting of the return value was different from the formatting of the expted value after running the test. I don't know why that was happening, thought. We summized that it is because I am located in Jamaica outside of the US.

patches/domino+2.1.6.patch
Paul-Clue: It worked! I removed the patch and pushed the changes.

client/utils/gitUtils.js
howard-e: Okay got it!

> We summized that it is because I am located in Jamaica outside of the US.

I still don't think this change should stick though. I forced my locale to `en-JM` and found it's the same as `en-US`, based on the expected format of the function. So maybe your device's locale could still be different somehow?

What I did notice though is that the tests at [gitUtils.test.js#L4:L8](https://github.com/w3c/aria-at-app/blob/19298fd510b35177db4bad5772cec0f21687e1b4/client/utils/gitUtils.test.js#L4:L8) and [gitUtils.test.js#L10:L14](https://github.com/w3c/aria-at-app/blob/19298fd510b35177db4bad5772cec0f21687e1b4/client/utils/gitUtils.test.js#L10:L14) are expecting a specific format by default (which seems to typically be `en-` related locales) by it not explicitly providing the locale on lines 6 & 12 to the `gitUpdatedDateToString` function. So this could cause an issue when running the tests on a device not using that format.

So I think you could try reverting the change to the `locale`'s default value and update those 2 tests to explicitly use `en-US` and see how that runs?

Paul-Clue: Ok. I'll try it out.

Paul-Clue: That work, @howard-e!

howard-e (approved):
@Paul-Clue looks good! Thanks!

Also adding @evmiguel as a reviewer here as well

evmiguel (approved):
@Paul-Clue changes look good! I tested this on locally on a vagrant machine and with the standard dev setup. I can confirm that the migrations work, and the app is functional. Looks like there are `yarn.lock` conflicts. Once those are resolved, we can merge.

During this testing, I found an issue; it is a bug that happens with the sample test data (https://github.com/w3c/aria-at-app/issues/505). @alflennik

PR #503, opened 2023-02-22
Update rest of dependencies 2

evmiguel:
This PR updates:

- prettier
- babel
- ua-parser-js
- lhci/cli

It assumes a Node 18 environment when running `yarn dev`.


client/package.json
```diff
@@ -24,6 +24,8 @@
   "homepage": "https://github.com/bocoup/aria-at-app#readme",
   "dependencies": {
     "@apollo/client": "^3.3.19",
+    "@babel/core": "^7.21.0",
```
howard-e, line 27: Seems `@babel/core` should be a `dev` dependency

client/package.json
```diff
@@ -24,6 +24,8 @@
...
+    "@babel/plugin-transform-runtime": "^7.21.0",
```
howard-e, line 28: Same with `@babel/plugin-transform-runtime`

client/package.json
```diff
@@ -90,7 +90,7 @@
     "jest-environment-jsdom": "^29.3.1",
     "jest-enzyme": "^7.1.2",
     "moxios": "^0.4.0",
-    "prettier": "^1.19.1",
+    "prettier": "^2.8.4",
```
howard-e, line 93: There's a fair amount of the existing formatting that's treated as errors due to the upgrade to `prettier`, that I think this PR could address as well. The changes can be seen when using the existing script command `yarn prettier` but that will write the changes. You can also check them with `npx eslint .`:

```
  ...
  102:29  error  Insert `,`                                          prettier/prettier
  109:25  error  Insert `,`                                          prettier/prettier
  155:23  error  Insert `,`                                          prettier/prettier

/path/to/aria-at-app/server/util/getUsersFromFile.js
   4:32  error  Replace `file` with `(file)`    prettier/prettier
   8:40  error  Replace `line` with `(line)`    prettier/prettier
   9:40  error  Replace `(line` with `((line)`  prettier/prettier
  10:35  error  Replace `line` with `(line)`    prettier/prettier

/path/to/aria-at-app/server/util/sortArrayLikeArray.js
  16:20  error  Replace `array2Item` with `(array2Item)`  prettier/prettier
  17:24  error  Replace `array1Item` with `(array1Item)`  prettier/prettier

/path/to/aria-at-app/server/util/sortArrayLikeArray.test.js
   9:38  error  Insert `,`                                                           prettier/prettier
  15:32  error  Replace `item·=>·item.id·||·item` with `(item)·=>·item.id·||·item,`  prettier/prettier
  22:41  error  Insert `,`                                                           prettier/prettier

✖ 1743 problems (1743 errors, 0 warnings)
  1743 errors and 0 warnings potentially fixable with the `--fix` option.
```

Updating `.prettierrc` with the following props gets us closer to what existed before:
```
"arrowParens": "avoid",
"trailingComma": "none"
```

But the formatting warnings that remain are opinionated whitespace handling coming from prettier which I have no real stance on. They'll also conflict with the default `eslint` rules.

howard-e:
@evmiguel note this PR now has merge conflicts

client/package.json
evmiguel: @howard-e I added the prettier changes. There might be potential workarounds if there are conventions we want to keep.

evmiguel: @alflennik I'm ok with doing that. In this PR @howard-e?

evmiguel: Oh, got it!

PR #509, opened 2023-02-28
Update testers.txt

howard-e:
Update `testers.txt`:
* Add [Carmen Cañas](https://github.com/ccanash) from Bocoup


howard-e:
cc @ccanash

Issue #505, opened 2023-02-27
Front end with sample seed data has errors


evmiguel:
When I populate a fresh database installation with `yarn workspace server db-populate-sample-data:dev;`, the Reports page and Candidate Tests page contain errors. See screenshots for errors:

- Error on Reports Page:
  - In the screenshot, the support levels are at 0%. The levels should be greater than 0.   
  - <img width="1281" alt="In the screenshot, the support levels are at 0%. The levels should be greater than 0.  " src="https://user-images.githubusercontent.com/2789098/221609236-56df99fe-6453-4455-8bb6-4841fc5b4b91.png" >

- Error on Candidate Tests Page:
  - In the screenshot, the results summaries show `NaN assertions failed across NaN tests`. The NaNs should be numbers. 
  - <img width="1271" alt="In the screenshot, the results summaries show `NaN assertions failed across NaN tests`. The NaNs should be numbers." src="https://user-images.githubusercontent.com/2789098/221609044-46e2ad11-685c-4cea-a5a8-9016a909916b.png">

alflennik:
@evmiguel Thanks for making an issue for this. It's caused by a performance enhancement @howard-e added which denormalizes some of the database data under a "metrics" field. Perhaps we can augment the model code to automatically populate the metrics field when a given row is created or updated.

Also @evmiguel would you be so kind to write some alt text for the screenshots, I'm sure that would be appreciated!

alflennik:
Thank you Erika! Now that we fixed the issue we can close this.

PR #510, opened 2023-03-01
updating candidate reports in sample seed data

evmiguel:
This PR addresses #505. In the issue, @alflennik proposed: "augment the model code to automatically populate the metrics field when a given row is created or updated". This is technically done when the status is changed on a test report. 

This solution is a quick fix on the generating the sample seed data. In this case, it's unclear to me how to mimic updating the status, so I called the `updateStatusResolver` in the populate data script. Open to other thoughts @alflennik @howard-e 

Steps to test:
- Drop the `aria-at-report` database 
- Recreate it using the steps in the [documentation](https://github.com/w3c/aria-at-app/blob/main/docs/database.md#setting-up-a-local-database-for-development)


alflennik (changes requested):
Thanks for fixing the issue! I haven't run it locally yet but it definitely looks like a practical solution that will work. My main concern is more with the fact that populate test data is currently limited to using the external API, but the code here is using resolvers directly and bypassing the external API. Is it possible to fix the issue internally within the server?

alflennik:
Hi @evmiguel I see you added a description to the PR and were wondering about other approaches you could take. Maybe you could figure out how it's currently possible in GraphQL to save a report with an inaccurate metrics.supportPercent value and fix it so that value can never fall out of sync. If that doesn't make sense let me know and maybe we can do a quick call about it. Thanks!!

server/graphql-schema.js
```diff
@@ -620,6 +620,14 @@ const graphqlSchema = gql`
         """
         See Assertion for more information.
         """
+        assertion: AssertionResultInput
```
evmiguel, line 623: @alflennik we added an assertion and priority to the schema because the [getMetrics function](https://github.com/w3c/aria-at-app/blob/main/server/util/getMetrics.js) uses the priority to do calculations

evmiguel:
@alflennik, @Paul-Clue and I tried resolving this through the server side. Let us know your thoughts!

server/graphql-schema.js
alflennik: I think one way to load that data would be to use the testResultsResolver in server/resolvers/TestPlanRun/testResultsResolver.js which will populate that association.

alflennik:
Thanks Erika! We've now fixed this issue.

Issue #493, opened 2023-02-15
Application fails to render incomplete candidate test plans


jugglinmike:
Steps to reproduce:

1. Create an incomplete "candidate" test plan
   1. Visit the "Test Queue" page
   2. Create a new plan to the test queue (these values appear to be inconsequential; I have recreated the bug with other plans, ATs, and browsers)
      - Test Plan: "Alert Example"
      - Assistive Technology: NVDA
      - Browser: Firefox
   3. Click the button labeled "Assign Yourself"
   4. Click the button labeled "Start testing"
   5. Select Assistive Technology Version "2020.4" (again this choice does not appear to matter for this report)
   6. Click "Save and Continue"
   7. Input any information into the required fields for the first test
   8. Click the button labeled "Submit Results"
   9. Visit the "Test Queue" page
   10. Click the button labeled "Mark as In Review"
   11. Click the button labeled "Mark as Candidate"
2. Review the incomplete "candidate" test plan
   1. Visit the "Candidate Tests" page
   2. Visit the new plan's page
   3. Navigate to the second test

Expected behavior: the application displays instructions and results for the second test, and no messages are printed to the browser's developer console.

Actual behavior: the application displays a blank page, and the following error is printed to the browser's developer console: "Uncaught TypeError: scenarioResult is undefined" with location `getMetrics.js:49:34`.

<details>
  <summary>complete stack trace</summary>

    countScenarioResult getMetrics.js:49
    countAssertions getMetrics.js:62
    getMetrics getMetrics.js:91
    disclosureContainerView index.jsx:492
    CandidateTestPlanRun index.jsx:487
    React 11
    unstable_runWithPriority scheduler.development.js:653
    React 21
    unstable_runWithPriority scheduler.development.js:653
    React 5
    onNewData useBaseQuery.js:17
    promise callback*onNewData useBaseQuery.js:17
    next QueryData.js:225
    notifySubscription Observable.js:135
    onNotify Observable.js:179
    next Observable.js:235
    iterateObserversSafely iteration.js:4
    iterateObserversSafely iteration.js:4
    next ObservableQuery.js:21
    iterateObserversSafely iteration.js:4
    iterateObserversSafely iteration.js:4
    next Concast.js:25
    notifySubscription Observable.js:135
    onNotify Observable.js:179
    next Observable.js:235
    promiseQueue asyncMap.js:19
    promise callback*makeCallback/< asyncMap.js:17
    notifySubscription Observable.js:135
[...]

PR #512, opened 2023-03-02
Disallowing Test Plans with Incomplete Tests

aleenaloves:
Reference Issue: https://github.com/w3c/aria-at-app/issues/493

Per the working group, incomplete test plans cannot transition to the candidacy status. This PR is a fix for this issue.

I have ran all these test cases, but have also listed them here for people to double check and/or list if i missed anything. 

Test Cases:

- Verify the test cases from the original issue no longer encounter the error
- Verify the behavior for an incomplete candidate test already existing

1. From another branch, transition an incomplete test plan to the candidate status
2. Migrate back to this branch, verify that the test navigator shows the correct icons for incomplete tests, the incomplete test also shows incomplete, and that a user cannot 'finish' this test. 

- Verify the behavior for an incomplete test plan in review on the test queue page

1. Create an incomplete test plan
2. Set it's status to 'In Review'
3. Validate that you cannot transition it to 'candidate' and that it tells you why you cannot as well.

- Verify the styling in general
- Verify that nothing else is broken

1. General Test Queue Behavior
2. General Candidate Test Behavior

### DIFF
--- a/client/components/CandidateTests/CandidateTestPlanRun/CandidateTestPlanRun.css
+++ b/client/components/CandidateTests/CandidateTestPlanRun/CandidateTestPlanRun.css
@@ -157,3 +157,7 @@
     position: relative;
     top: -5px;
 }
+
+.incomplete-test-plan-information-text {
+    margin-top: 5%;
+}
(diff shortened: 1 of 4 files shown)


aleenaloves:
pending test fixes.

aleenaloves:
tests fixed.

evmiguel:
Are we ok with the copy: `Test plans with incomplete tests cannot be finished. These tests should have their candidate status removed.`? @jugglinmike @howard-e

aleenaloves:
pending copy notes from isaac

PR #515, opened 2023-03-07
Fix For Fake Roles Login

Paul-Clue:
This PR fixes the issue that occurs when a user tries to sign in but is not on the list of testers for the app and is not a part of the github team. The user should be able to sign in as a "fake" admin or tester using the link `/?fakeRole=admin` . However, the user was not able to because of a logical error in `AuthController.js`.


evmiguel (changes requested):
Can this branch be rebased onto main / merge main into this branch?

alflennik:
This PR is no longer relevant with the redesign of the fake user system in https://github.com/w3c/aria-at-app/pull/908

PR #517, opened 2023-03-08
Regex upgrade node version fix

evmiguel:
Refining the node version check with a regex match.


alflennik (approved):
Based on our previous conversation it seems like we just need to change "18" to "v18" but this works just as well, so I'm approving.

PR #519, opened 2023-03-09
Optimize loading of embedded reports

howard-e:
This PR replaces the logic for displaying the latest version used with each respective `TestPlanReport` displayed on the embed page with an optimized query served through the `testPlanReports.latestAtVersionReleasedAt` resolver.


alflennik (approved):
Although there is an edge case you mentioned in our call, where the latest report might be stale, this is a functional implementation that we can run with for now. And the word optimize is right - the performance is much speedier, which is very satisfying.

PR #520, opened 2023-03-09
Support tables minor styling tweak

alflennik:
Following some testing I wanted to make the support tables UI less prone to shifting around when you interact with it.


howard-e (approved):
This seems to be addressing a visual flash which could happen when expanding the `<details>` disclosure element on the embedded report. 

Can confirm this addresses that issue!

PR #521, opened 2023-03-13
Address support tables 2.0 feedback

alflennik:
Implements fixes for additional issues found here: https://github.com/w3c/aria-practices/pull/2569

### DIFF
--- a/server/handlebars/public/style.css
+++ b/server/handlebars/public/style.css
@@ -71,8 +71,12 @@ h3#report-title {
     display: inline-block;
 }
 
-#candidate-title.recommended h3 {
-    width: 130px;
+#candidate-title.recommended {
+    border: 1.5px solid #7ac498;
+    background-color: #e9fbe9;
+}
+#candidate-title.recommended > span {
+    background-color: #115b11;
 }
 
 #candidate-content-container {
@@ -90,7 +94,7 @@ h3#report-title {
     margin-bottom: 3px;
 }
 
-.none {
+.no-data-cell {
     display: block;
     color: #72777f;
     font-style: italic;
--- a/server/handlebars/views/main.hbs
+++ b/server/handlebars/views/main.hbs
@@ -1,6 +1,5 @@
 <div id="main">
     {{#if dataEmpty}}
-        <h3 id="report-title">{{title}}</h3>
         <div id="no-data-container">
             <div id="no-data-content-container">
                  There is no data for this pattern.
@@ -13,7 +12,7 @@
             <summary id="candidate-title"><span>Warning! Unapproved Report</span></summary>
             <div id="candidate-content-container">
                 The information in this report is generated from candidate tests developed and run by the <a href="http://aria-at.w3.org">ARIA-AT Project</a>.
-                ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:
+                Candidate ARIA-AT tests are in review by assistive technology developers and lack consensus regarding:
                 <ol>
                     <li>applicability and validity of the tests, and</li>
                     <li>accuracy of test results.</li>
@@ -21,18 +20,17 @@
             </div>
         </details>
         {{else}}
-        <div id="embed-report-status-container">
-            <div id="candidate-title" class="recommended"><h3>Recommended Report</h3></div>
+        <details id="embed-report-status-container">
+            <summary id="candidate-title" class="recommended"><span>Recommended Report</span></summary>
             <div id="candidate-content-container">
                 The information in this report is generated from recommended tests.
-                Recommended ARIA-AT tests have been reviewed by assistive technology
-                developers and represent consensus regarding
+                Recommended ARIA-AT tests have been reviewed by assistive technology developers and represent consensus regarding:
                 <ol>
                     <li>applicability and validity of the tests, and</li>
                     <li>accuracy of test results.</li>
                 </ol>
             </div>
-        </div>
+        </details>
         {{/if}}
         <div class="responsive-table">
             <table id="embed-report-table" aria-label="Assistive Technology Support for {{title}}">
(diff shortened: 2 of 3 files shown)


server/handlebars/views/main.hbs
```diff
@@ -13,26 +12,25 @@
...
                     <li>applicability and validity of the tests, and</li>
                     <li>accuracy of test results.</li>
                 </ol>
             </div>
         </details>
         {{else}}
-        <div id="embed-report-status-container">
-            <div id="candidate-title" class="recommended"><h3>Recommended Report</h3></div>
+        <details id="embed-report-status-container">
+            <summary id="candidate-title" class="recommended"><span>Recommended Report</span></summary>
```
howard-e, line 24: I'd be curious about what the community's thoughts would on this title. Should the disclaimer also be the reverse of the `CANDIDATE` disclaimer, so `Approved Report`? 

This is also done on the application's Reports pages as well.

aleenaloves:
### Summary
Pending explanations/implementations/corrections on items I was unable to verify this is good to approve :)

### General Code Review 

Generally, everything looks awesome @alflennik ! I only had the following question, but it doesn't bar this PR from being approved.

- I know in CSS we can also use [variables/custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties), in terms of some of the feedback where we get told to use the same colors as xyz, would this be a helpful pattern to start implementing in a separate PR? 

### In reference to the [original PR](https://github.com/w3c/aria-practices/pull/2569#issuecomment-1398008506)

1. Increase the color contrast ratio of the text in the cells that contain the word "None" to the required 4.5 or better. ✅
2. Change copy in cells with data to "%X Supported", e.g. "96% Supported". ✅
3. In cells that contain the word "None", change copy to one of the two options described below. ✅ 

[Partially Complete](https://app.asana.com/0/1203924324409025/1204166626529996/f)

4. Change copy of the text above the table (warning text) as described below. ✅
5. Remove heading markup from the warning and make it a disclosure that is collapsed on load like it is on the aria-at site. ✅
6. Change the accessible name of the button "View Complete Report" to "View Complete Report in New Window". ✅
7. Change the visible name of the button "Embed" to "Copy Embed Code" and revise the confirmation announcement to be "Copied code for embedding".  ✅
8. Either by using a visible caption or by using aria-label, give the table an accessible name of "Assistive Technology Support for PATTERN_NAME", e.g., "Assistive Technology Support for Command Button". Don't allow callers to change the caption/name. 

❌❔ the preview currently reads `Assistive Technology Support for Command Button **Example**` 

9. Do not include a heading for the content, e.g., the h3 with text "Command Button", in the iframe. The iframe should contain only the warning/providence disclosure, the table, and the buttons for full report and embedding. On pages like the button example page where there are multiple examples, we will include the H3 markup in the example page content followed by the iframe. 

❔ unable to verify, could not find an example of a single one in the preview 

10. Order rows and columns in the table alphabetically. ✅
11. Increase the color contrast of the table borders (make them match other tables on the page). 

❌ [Contrast Value](https://webaim.org/resources/contrastchecker/?fcolor=F2F2F2&bcolor=FFFFFF) 

12. If possible, suppress the visibility of the table scrollbars when they are not needed. ✅

howard-e (commented):
Looks good to me @alflennik! While I do have additional questions listed below, those to me aren't blocking as they can be updated once we receive further feedback and directives there.

What I do think is blocking (which this PR doesn't actually touch or address) is around how the `CANDIDATE`/`RECOMMENDED` status is evaluated. [server/apps/embed.js lines 92:94]( https://github.com/w3c/aria-at-app/blob/99813f7e78246bd7497098fd5cab8a20405d1241/server/apps/embed.js#L92:L94) is being done against all reports which have existed and would have previously had a `CANDIDATE`/`RECOMMENDED` status. This can quite easily prevent the correct disclaimer being shown above the table. This evaluation should be done further down against `latestReports` so it is actually done against the data being shown in the table.

server/handlebars/public/style.css
```diff
@@ -71,8 +71,12 @@ h3#report-title {
     display: inline-block;
 }
 
-#candidate-title.recommended h3 {
-    width: 130px;
+#candidate-title.recommended {
+    border: 1.5px solid #7ac498;
```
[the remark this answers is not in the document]
howard-e: Okay got it! Missed that

server/handlebars/views/main.hbs
alflennik: It would be good to have them review it. Can we promote one of the reports to a recommended state on the sandbox?

alflennik:
@aleenaloves First of all thanks for the detailed review!

alflennik:
@aleenaloves About "Increase the color contrast of the table borders," you noted that the contrast value is short of the recommended 3:1 value. It seems like this is an issue for the entire site, since we're reusing the same value from the entire site. How do you think we should fix it?

alflennik:
@aleenaloves Thank you for noticing that the test titles have inconsistent names (some have the word Example and some don't). While we could definitely work around that within the embed system I think it makes sense to take this as an opportunity to improve the overall consistency of our data, and I'll bring this up with the rest of the team.

alflennik:
@howard-e Thanks for finding that issue with the candidate / recommended algorithm!

aleenaloves:
> @aleenaloves About "Increase the color contrast of the table borders," you noted that the contrast value is short of the recommended 3:1 value. It seems like this is an issue for the entire site, since we're reusing the same value from the entire site. How do you think we should fix it?

You're right that this is an issue for the entire website. Since this would then be a design concern, it may make sense to get isaac's input on it - but irregardless, irrelevant to this PR in particular. 

Since the 2nd part of matt's original step 11 also included:
> (make them match other tables on the page)

Not sure what tables he's exactly referring to, but if it's the `Role, Property, State, and Tabindex Attributes` table, then that uses a border color of `silver` or `#C0C0C0`. 

But i defer to you on what you think he meant/what you think would be best in this scenario.

alflennik:
I did make an issue about the contrast here, https://github.com/w3c/wcag/issues/3102, but for now it seems like the best move is to keep the contrast consistent with the other tables on the site which was the ask originally.

aleenaloves:
thanks @alflennik ! loved the investigation ~

i'm good to approve this, pending @howard-e's notes :D

evmiguel (approved):
LGTM! Great job!

PR #586, opened 2023-03-30
Remove IN_REVIEW status

Paul-Clue:
This pull request removes the `IN_REVIEW` status from the test plans in the test queue of the app.

### DIFF
--- a/client/components/TestQueueRow/index.jsx
+++ b/client/components/TestQueueRow/index.jsx
@@ -500,15 +500,14 @@ const TestQueueRow = ({
         const { status: runStatus, conflictsLength } = testPlanReport;
 
         let status, results;
-        const conflictsCount = conflictsLength;
 
         if (isLoading) {
             status = (
                 <span className="status-label not-started">Loading ...</span>
             );
-        } else if (conflictsCount > 0) {
-            let pluralizedStatus = `${conflictsCount} Conflict${
-                conflictsCount === 1 ? '' : 's'
+        } else if (conflictsLength > 0) {
+            let pluralizedStatus = `${conflictsLength} Conflict${
+                conflictsLength === 1 ? '' : 's'
             }`;
             status = (
                 <span className="status-label conflicts">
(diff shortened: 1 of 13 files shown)


evmiguel (commented):
Great work @Paul-Clue! I see that you've touched on all aspects of the stack that need to change and that it works!

server/migrations/20230329214527-removeInReviewStatusFromTestPlanReport.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+module.exports = {
+    up: async queryInterface => {
+        await queryInterface.sequelize.query(
+            `update "TestPlanReport" set status = 'DRAFT' where status = 'IN_REVIEW'`
```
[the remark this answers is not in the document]
evmiguel: Ah, got it. Thanks for the clarification!

client/components/TestQueueRow/index.jsx
```diff
@@ -517,39 +517,30 @@ const TestQueueRow = ({
...
-        }
+        } 
 
         return { status, results };
     };
 
     const evaluateNewReportStatus = () => {
-        const { status, conflictsLength } = testPlanReport;
+        const { conflictsLength } = testPlanReport;
         const conflictsCount = conflictsLength;
```
howard-e, line 527: nit: I think the language was being preserved based on an earlier iteration of the application but the `conflictsCount` variable may be unnecessary as it isn't (and cannot be) mutated, so we could reference `conflictsLength` directly. We can maybe add a comment that it is the `count` as well if we'd like to be explicit.

This also applies to its usage in `evaluateStatusAndResults`

client/components/TestQueueRow/index.jsx
```diff
@@ -517,39 +517,30 @@ const TestQueueRow = ({
...
 
         // If there are no conflicts OR the test has been marked as "final",
```
howard-e, line 529: nit: this comment seems to describe when the application could downgrade a Test Plan Report within the Test Queue so this could be removed.

howard-e (approved):
@Paul-Clue looks good!

I have only nits which I've mentioned in the following comments but nothing that's breaking. Additionally, there are some formatting changes which could be addressed by running `yarn prettier`

evmiguel (approved):

PR #596, opened 2023-04-04
Unexpected behavior note follow up

evmiguel:
This PR encompasses @alflennik's commits with a merge to main and a few other changes.


client/components/TestRun/index.jsx
```diff
@@ -390,34 +391,53 @@ const TestRun = () => {
...
-                            unexpectedBehaviors.push({
-                                id: 'UNEXPECTED_CURSOR_POSITION'
-                            });
-                        if (i === 2)
-                            unexpectedBehaviors.push({ id: 'SLUGGISH' });
-                        if (i === 3)
-                            unexpectedBehaviors.push({ id: 'AT_CRASHED' });
-                        if (i === 4)
-                            unexpectedBehaviors.push({ id: 'BROWSER_CRASHED' });
+                        if (i === 0) {
```
evmiguel, line 394: REVIEWERS: this is the main concern here. Why does graphql expect a string while the TestRenderer expect an object with an id?

client/components/TestRun/index.jsx
```diff
@@ -447,8 +431,23 @@ const TestRun = () => {
             !rendererState ||
             !testResult.scenarioResults ||
             rendererState.commands.length !== testResult.scenarioResults.length
-        )
-            return testResult;
+        ) {
+            // Mapping unexpected behaviors to expected TestRenderer downstream format
```
evmiguel, line 435: REVIEWERS: I had to do this so that the TestRenderer would receive consistent data.

ccanash (commented):
I tested out and it works

client/resources/aria-at-harness.mjs
```diff
@@ -371,47 +393,69 @@ function renderVirtualInstructionDocument(doc) {
...
-                  )
+            label(
+              forInput(`${failOption.description}-${commandIndex}`),
+              rich(failOption.description)
+            ),
+            br()
+          )
+        ),
+        fragment(
+          // TODO: Figure out why this isn't appearing
```
alflennik, line 443: Did I write this lol? Can we remove this comment or is it still relevant?

alflennik (changes requested):
Thank you for reviving this ancient PR, thinking of the conflicts you probably had to fix makes me sweat! I did notice one issue when testing it, which is the case where I select that "Yes, there were unexpected behaviors" and then fail to specify any of the radio buttons. When I hit submit, the toggle switches from "Yes" to "No", but more importantly the error is not focused or scrolled to, forcing me to look over all the forms trying to find the one that didn't work.

client/components/TestRun/index.jsx
```diff
@@ -447,8 +431,23 @@ const TestRun = () => {
...
+            const scenarioResults = testResult.scenarioResults.map(
+                scenarioResult => {
+                    return {
+                        ...scenarioResult,
+                        unexpectedBehaviors: scenarioResult.unexpectedBehaviors
+                            ? scenarioResult.unexpectedBehaviors.map(
+                                  behavior => behavior.id
+                              )
+                            : scenarioResult.unexpectedBehaviors
```
alflennik, line 444: Makes sense! This could be simplified slightly to `scenarioResult.unexpectedBehaviors ?? scenarioResult.unexpectedBehaviors.map(behavior => behavior.id)`

evmiguel:
> Thank you for reviving this ancient PR, thinking of the conflicts you probably had to fix makes me sweat! I did notice one issue when testing it, which is the case where I select that "Yes, there were unexpected behaviors" and then fail to specify any of the radio buttons. When I hit submit, the toggle switches from "Yes" to "No", but more importantly the error is not focused or scrolled to, forcing me to look over all the forms trying to find the one that didn't work.

The error is scrolled to and `required` is shown, when adding inputs in order. As for the toggling, the app is written in such a way that if you submit, "Yes", but there are no radio buttons, the default is set to "No". What is the intended behavior?

evmiguel:
@alflennik You should now be able to observe that when you select "Yes" with no unexpected behaviors and submit the form, the page scrolls to the required element and "Yes" is not toggled to "No".

PR #620, opened 2023-05-01
Test plan table

evmiguel:
This PR implements a higher level Test Plan table, which has a one-to-many relationship with both the Test Plan Version and Test Plan Report table. This PR also changes the GraphQL resolvers to use the new Test Plan model. Looking forward to the discussion that arises from this!

To use:
```
yarn sequelize db:migrate;
yarn sequelize db:seed:all;
```


server/migrations/20230425152008-createTestPlanColumns.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+    async up(queryInterface, Sequelize) {
+        const t = await queryInterface.sequelize.transaction();
+        try {
+            await queryInterface.addColumn(
```
howard-e, line 8: The foreign key constraints you've added declared in `models/TestPlan.js`  has to be added to the migration scripts as well

server/models/services/TestPlanService.js
```diff
@@ -0,0 +1,63 @@
...
+                association: 'testPlanReports',
+                attributes: TEST_PLAN_REPORT_ATTRIBUTES
+            }
+        ],
+        pagination,
+        options
+    );
+
+    return data.map(d => {
+        const latestTestPlanVersion = d.dataValues.testPlanVersions.sort(
```
howard-e, line 37: I believe `sequelize` provides a way for us to also declare the order with the included `association`s, which may be more optimal to do in this case.

I haven’t found a clear example for this version (`v6`) but this link may be helpful https://sequelize.org/docs/v6/core-concepts/model-querying-basics/#ordering.

server/resolvers/testPlansResolver.js
```diff
@@ -1,28 +1,9 @@
...
-        true
-    );
-
-    return getTestPlans({
-        includeLatestTestPlanVersion: includeLatest,
-        includeTestPlanVersions: includeHistoric,
-        testPlanVersionOrder: [['updatedAt', 'desc']],
-        testPlanVersionAttributes
+const testPlans = async () => {
+    const plans = await getTestPlans();
```
howard-e, line 4: Just to point out that because the default TEST_PLAN_VERSION_ATTRIBUTES are will be used in the function being called, it will force unexpected columns being queried as well. Such as the `tests` column from `TestPlanVersion`, which we know causes a huge overhead when querying from the database.

Maybe you could try using this with the `retrieveAttributes` function which was previously used here and also used in other resolvers? It attempts to map the raw attributes requested in the GraphQL query, to relevant column names for the table(s), so that only those columns are queried for.

howard-e (commented):
Thanks @evmiguel! A huge stepping stone to unlocking future functionality in the application.

I've left some follow up comments in the thread.

howard-e (approved):

alflennik (approved):
This is fantastic work Erika and also thanks Howard for figuring out the initial proposal. In addition to reviewing the code and verifying that the app works as intended, this PR led to hour-long discussion with Howard and the other members of the engineering team, where I learned a lot about how the app ACTUALLY works, as opposed to how I thought it works. At the end of that I was sold that this is the right approach and more of a future-proof solution than I even realized initially. Since we now have two approvals @howard-e do you want to merge this on Erika's behalf?

Issue #580, opened 2023-03-27
Test Management Page: 19 test plans are missing from the status summary table


mcking65:
There are 36 test plans in the [aria-at tests folder](https://github.com/w3c/aria-at/tree/master/tests) but only 17 on the [test summary page](https://aria-at.w3.org/test-management).

At the time of filing this issue, the only test plans listed in the status summary are:
1. Alert Example     
1. Breadcrumb Example
1. Checkbox Example (Mixed-State)
1. Color Viewer Slider
1. Combobox with Both List and Inline Autocomplete Example
1. Command Button Example
1. Data Grid Example 1: Minimal Data Grid
1. Disclosure Navigation Menu Example
1. Link Example 1 (span element with text content)
1. Main Landmark
1. Meter
1. Modal Dialog Example
1. Navigation Menu Button
1. Radio Group Example Using aria-activedescendant
1. Select Only Combobox Example
1. Switch Example
1. Toggle Button

mcking65:
@howard-e, Good news ... now I am not seeing the extra plan that I mentioned in our meeting. Perhaps I wasn't looking at production or was not reading where I thought I was reading.

mcking65:
Reviewed content in staging and compared to repo; this issue is resolved.

PR #622, opened 2023-05-03
Fix missing rows test management

evmiguel:
This PR addresses https://github.com/w3c/aria-at-app/issues/580. To see, run `yarn dev`, and then navigate to the Management page. Notice that all the test plan are present, even though not all of them have been tested.

---
  - https://app.asana.com/0/0/1204735509616441
  - https://app.asana.com/0/0/1204382211311636

### DIFF
--- a/client/components/TestManagement/queries.js
+++ b/client/components/TestManagement/queries.js
@@ -15,6 +15,15 @@ export const TEST_MANAGEMENT_PAGE_QUERY = gql`
             id
             name
         }
+        testPlans {
+            directory
+            id
+            title
+            latestTestPlanVersion {
+                id
+                title
+            }
+        }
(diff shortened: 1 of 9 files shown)


client/components/TestManagement/queries.js
```diff
@@ -15,6 +15,15 @@ export const TEST_MANAGEMENT_PAGE_QUERY = gql`
             id
             name
         }
+        testPlans {
```
evmiguel, line 18: REVIEWERS: even if we do not go with a separate Test Plan table, this is the query I would like to use.

howard-e (approved):

alflennik (approved):
Looks great, code looks good and I verified that the issue identified by Matt is fixed.

PR #623, opened 2023-05-09
Improve AT version and browser version queries 

Paul-Clue:
This PR improves the test queue query speed by 69X and improves the querying speeds for all the rest of pages as well.

- Before this PR, the test queue was querying the cartesian product of all tests, ATs, AT versions, browsers, and browser versions. 
- We removed the `includes` query from the `atAssociation` and `browserAssociation` functions in the  `TestPlanVersionService` and `TestPlanReportService` files. The `includes` query is what was causing the cartesian product problem.
- We then created an `AtLoader` and `BrowserLoader` that fetches all of the AT versions and browser versions, respectively.
- To test the speed improvement, use the production database. Follow the instructions in the database README to see how to use the production database.


howard-e (commented):
This is a huge amount of work, the pages are zooming! Thanks for this investigation and major optimization @Paul-Clue @alflennik 👍🏽

evmiguel:
@Paul-Clue @alflennik amazing work! Pending @howard-e's change requests, this is good to go!

PR #633, opened 2023-05-25
Encode reports URL when multiple versions are being checked

howard-e:
On the reports page (`/reports`), there is a chance that multiple versions can to determine which version has the most relevant data. Currently the url is displayed as `/report/1090,1890,2215`, where each number represents a different version of the same test plan.

This change would encode the url to `/report/1090%2C1890%2C2215` instead, to avoid unpredictable situations when navigating to the address.

### DIFF
--- a/client/components/Reports/SummarizeTestPlanReports.jsx
+++ b/client/components/Reports/SummarizeTestPlanReports.jsx
@@ -222,7 +222,9 @@ const SummarizeTestPlanReports = ({ testPlanReports }) => {
                                 <tr key={testPlanVersionId}>
                                     <td>
                                         <Link
-                                            to={`/report/${testPlanVersionId}`}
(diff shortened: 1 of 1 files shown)


alflennik:
@howard-e How about using a hyphen instead? The percentage encoding seems pretty off-putting to me. Unless ... does it still look like commas when it's in the URL bar?

howard-e:
> How about using a hyphen instead? The percentage encoding seems pretty off-putting to me.

Sure! I'm fine with that

> Unless ... does it still look like commas when it's in the URL bar?

Nope, it would include the encoding there seen here. Browser clients and the application would continue to interpret them as commas though

howard-e:
This will no longer be relevant with the separation of `TestPlanVersion`s incoming

PR #685, opened 2023-06-08
Tests which have been re-ran to update results to avoid conflicts between testers still incorrectly display conflicting results

Paul-Clue:
On the Test Queue page (`/test-queue`), there are test plans showing conflicts even though multiple testers have the same test results. The problem is that, somehow, some tests that end up with an unexpected behavior ID of "EXCESSIVELY VERBOSE" end up with `otherUnexpectedBehaviorText` in  `unexpectedBehavior` array. The `otherUnexpectedBehaviorText` should not be there.

This PR removes the `otherUnexpectedBehaviorText` from `unexpectedBehavior` arrays that are already in the database and it prevents the problem from occurring in future test plans.


server/migrations/20230523163856-remove-OtherUnexpectedBehaviorText.js
```diff
@@ -0,0 +1,163 @@
...
+                        {
+                            testPlanReportId
+                        },
+                        {}
+                    );
+
+                    const conflicts = await conflictsResolver(
+                        testPlanReport,
+                        null,
+                        { atLoader, browserLoader }
```
howard-e, line 149: Just a thought, and not suggesting as something to act on.

Right now, the GQL context could be `{ user, atLoader, browserLoader }` with the change in #623. Now that I've seen its usage a bit more, it seems `atLoader` and `browserLoader` being present is a must in most queries.

So I'd think that when `context` is null in the resolvers it's called, it may be preferred if it simply defaults to `{ atLoader, browserLoader }`. I'm unsure if sure if there's any implications to that though, and also acknowledge that it would a considerable effort in determining that. But creating the objects here to solely retrieve the conflict results here felt out of place.

server/migrations/20230523163856-remove-OtherUnexpectedBehaviorText.js
```diff
@@ -0,0 +1,163 @@
...
+                let needsUpdate = false;
+
+                let updateParams = {
+                    testResults: testPlanRun.testResults
+                };
+
+                // eslint-disable-next-line no-console
+                console.info(
+                    `=== Id for TestPlanRun results:${testPlanRunId} ===`
+                );
```
howard-e, line 50: This log may be unnecessary as it's just logging all the known TestPlanRun ids found in the database. I've suggested an alternative below.

server/migrations/20230523163856-remove-OtherUnexpectedBehaviorText.js
```diff
@@ -0,0 +1,163 @@
...
+                                    testPlanRunData[i].testResults[j];
+
+                                needsUpdate = true;
+                            }
+                        }
+                    }
+                }
+
+                if (needsUpdate)
+                    await updateTestPlanRun(testPlanRunId, updateParams);
```
howard-e, line 131:
```suggestion
                if (needsUpdate) {
                    // eslint-disable-next-line no-console
                    console.info(
                        `=== Fixing unexpectedBehavior results for TestPlanRun:${testPlanRunId} ===`
                    );
                    await updateTestPlanRun(testPlanRunId, updateParams);
                }
```

howard-e (changes requested):
Thanks @Paul-Clue, it looks good!

I've only left a comment about the info being logged.

howard-e (approved):

Issue #625, opened 2023-05-18
Define valid AT + Browser combinations


howard-e:
Following up from a discussion in #583. There are potential risks to the application not explicitly outlining valid AT and Browser combinations.

While I'm unsure if there are use cases where the application would want to intentionally avoid this restriction, there is still a need to know which combinations are considered as "valid" in the cases where the restriction is required:

> Something of note is that while this solution does work under the ideal happy case, it is also pretty easy to get through (in the absolute rarest case of an admin overlooking several reviews or the process), and this also has nothing to do in particular with this PR.

> Example being that going to `/embed/reports/<plan>`, it correctly notes that the combination of JAWS + Safari is `Not Applicable` but I am still able to submit a JAWS + Safari combination from the Test Queue, push it all the way through to `CANDIDATE` or `RECOMMENDED` and a percentage value is displayed for JAWS + Safari even though it previously said `Not Applicable`. 

> I think we recently discussed this internally and the question would be how can we explicitly define valid Browser + AT Combinations, to be ideally used at the point when a Test Plan is being added to the Queue to prevent invalid combinations from being added? And potentially added as an additional safeguard for this embedded report?

_Originally posted by @howard-e in https://github.com/w3c/aria-at-app/pull/583#pullrequestreview-1371262760_

ccanash:
This corresponds to P0: P0: Single-page view of test plan.

PR #712, opened 2023-07-18
Add At and Browser Association

Paul-Clue:
This PR finalizes the fix for the At and Browser combos embed in the APG. If a test was never written with a particular At and Browser combo, it would be marked "Not Applicable", which may not be accurate. Howard created a temporary fix for this problem (https://github.com/w3c/aria-at-app/pull/650). The original issue can be found here: https://github.com/w3c/aria-at-app/issues/625.

### DIFF
--- a/server/seeders/20230622202911-addAtBrowser.js
+++ b/server/seeders/20230622202911-addAtBrowser.js
@@ -1,0 +1,34 @@
+'use strict';
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.createTable('AtBrowsers', {
+            atId: { type: Sequelize.DataTypes.INTEGER },
+            browserId: { type: Sequelize.DataTypes.INTEGER }
+        });
+        const jawsAtId = 1;
+        const nvdaAtId = 2;
+        const voiceOverAtId = 3;
+
+        const firefoxBrowserId = 1;
+        const chromeBrowserId = 2;
+        const safariBrowserId = 3;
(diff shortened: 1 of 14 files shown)


server/seeders/20230622202911-addAtBrowser.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        await queryInterface.createTable('AtBrowsers', {
+            atId: { type: Sequelize.DataTypes.INTEGER },
+            browserId: { type: Sequelize.DataTypes.INTEGER }
```
howard-e, line 7: As it is right, the `atId` and `browserId` foreign key references which links these columns to their respective tables aren't actually created in the database when this is ran.
You should be able to copy the attribute definitions in `models/AtBrowsers.js` and get that link properly created.

server/seeders/20230622202911-addAtBrowser.js
```diff
@@ -0,0 +1,34 @@
...
+        });
```
howard-e, line 8: This seems out of place in a seeder. As this is about database structure, I'd think this `createTable` block belongs in its own migration file instead.

howard-e (changes requested):
@Paul-Clue looks good!

I've only left a few comments around the migration itself but great that you got to successfully tackle this.

howard-e (approved):

Issue #709, opened 2023-07-12
UX and UI Revisions to Test Queue and Candidate Test Plan pages


isaacdurazo:
Some of the changes proposed in issue #648 have an impact on the Test Queue and Candidate Test Plan pages. The majority of them relate to visual inconsistencies. There are also some opportunities to simplify the UI and UX. 

Before putting together a revised mockup for these two pages, I spent some time reviewing the impact of these changes. I ran into some questions as well as some ideas I would like to discuss:

**Test Queue**
1. Do we need to keep the "Manage AT Versions" and "Add Test Plans to the Test Queue" disclosures on this page? These disclosures are already under Data Management
2. We should rename the “Report Status” column to “Overall Status” to make it consistent with the Data Management page

3. The “Update Test Plan” button’s functionality only applies to a single AT/Browser combination under a given Test Plan. Shouldn’t this action apply to the Test Plan as a whole? Meaning for all the AT/Browser combinations under a given Test Plan? 

4. Do we want to keep the “Mark as Candidate” button on this page or do we want to only have it on the Data Management page?

**Candidate Tests**

1. I have notes about the “Mark as..” Button. This button currently moves a Test Plan to Recommended or Draft. 
   - We don’t have the option to move a Test plan back to Draft in the new Test Plan Status Summary table, why have it here? 
   - Similar to my question about the “Mark as Candidate” button in the test queue, do we need to keep this functionality on this page?
2. I think we should move the Review Status Summary table to the top so we have a high-level view first.
3. The current statuses under the Review Status column might not be consistent with the statuses in the new Test Plans Status Summary.

jscholes:
#### Test Queue

> Do we need to keep the "Manage AT Versions" and "Add Test Plans to the Test Queue" disclosures on this page? These disclosures are already under Data Management.

I don't think we need the first one ("Manage AT Versions"), but an "Add Test Plans to the Test Queue" button on the Test Queue page makes sense to me as a shortcut of sorts.  If there are reasons it should only live on the Data Management page, I'm also okay with its removal, but having it just do the same thing seems fine.

> We should rename the “Report Status” column to “Overall Status” to make it consistent with the Data Management page

I think this makes sense, because the test queue tracks test plan versions, not reports.  However, I want a gut check from @mcking65 on this one.

[...]

Issue #648, opened 2023-06-01
Design Missing Test Plan Status Management Functions


mcking65:
Related to #518.

**UPDATED JUNE 14**

#### June 14 revision summary

* Improved column titles.
* Changed plan version strings to "Vyy.mm.dd format so they are shorter and to prevent them from being confused with status dates.
* Minor revision to content of the "Overall Status" column, which required a significant revision to business logic for calculating status.
* Shortened and clarified label of button that opens report status dialog. Moved description of the button labeling logic to the section that describes the dialog.
* Shortened label of button that opens candidate target dialog.
* Reduce content in most cells of the phase columns.
* Added a requirement related to confirmation prompts when changing the phase of a version.
* Bumped priority of the "Test Plan Versions" page from P2 to P1. This page will be documented in a separate issue.
* Reduced priority of filters from P1 to P2.

#### Test Plan Status Table Proposal Overview

This issue describes a proposal for the test plan status summary table, including:
* The columns to include
* The content of the columns
* Default sort order
* Filters

It also introduces a  new concept of "required reports" to help manage progress of a test plan through the working mode process.

#### Development priorities

Each feature of the table is marked with a priority of P0, P1, or P2.
The assumption is that we will develop a short list of features in each sprint, with the first several sprints each delivering a set of P0 requirements and later sprints focused on P1 and P2 requirements.

#### Overview of Columns for Test Plan Status Summary Table

1. Test Plan
1. Covered AT
1. Overall Status
1. R&D Version
1. Draft Review
1. Candidate Review
1. Recommended Version

#### General Description of Phase Columns

**Note:** As described in w3c/aria-at#950, a test plan may have up to one version in each phase of the working mode. For example, there may be a recommended version and a new version that is ready for draft review at the same time.

The Hhigh-level objectives of the columns for the R&D, Draft, Candidate, and Recommended phases is to enable  users to monitor progress of a test plan through the working mode and enable admins to manage the process.

While the content of each column is different, the common objectives are to show:
* Whether the test plan has a version in that phase of the working mode.
* If there is not a version in the phase, shows if and when a version last completed that phase.
* If there is a version in the phase:
    * Show the version identifier and provide access to a view of that version of the plan
    * Show status of relevant working mode requirements and provide access to relevant reports
    * Provide admins with relevant  actions

As described in the detailed descriptions of column content that follows, each phase column may thus have:
* If the plan does not have a version in the phase, a simple status text string
[...]

Issue #632, opened 2023-05-25
Questions on proposed database changes to better support #518


howard-e:
There appears to be a need to augment the database's structure slightly, to better support what's being discussed in #518. Particularly around the proposed `Test (now Data) Management` page and its related features, once approved.

2 pages have been drafted to describe the current state of the database and the proposed changes:
* [Current Database Implementation](https://github.com/w3c/aria-at-app/wiki/ARIA-AT-App-Current-Database-Implementation)
* [Database Update Proposal](https://github.com/w3c/aria-at-app/wiki/ARIA-AT-App-Database-Update-Proposal)

I'd like to particularly raise focus on [some questions posed under the first point of the Notes and Questions section of the proposal doc](https://github.com/w3c/aria-at-app/wiki/ARIA-AT-App-Database-Update-Proposal#notes-and-questions) as they could assist with the final design proposed.

howard-e:
cc @mcking65

mcking65:
@howard-e 

Thank you again for the excellent data model documentation. The formatting you did withheadings, tables , and lists makes it easy to read with a screen reader.

I've now spent enough time with it to make some recommendations for better aligning the data model with the working mode, but first I will try to address the questions.

>1. Note that the most practical place for the phase column to be would be on the TestPlanGroup table, but it is located on TestPlanReport. 

Note that a specific version of a test plan is in a phase of the working mode, not the test plan itself. Phase should be on testPlanVersion. That would reflect the working mode process. A test plan can have a version in each phase of the working mode. V1 could be recommended while V2 is in candidate review, V3 is in draft review, and V4 is in R&D Complete.

>The following questions being discussed by the CG raises concerns on where it should be placed:
>What happens when a new Test Plan Report is introduced to the Test Queue, while a Test Plan is already under review in the CANDIDATE  phase, but there's a need for it to be included in that Test Plan?

Note that a report is not "introduced to the test queue." A test plan run is added to the test queue. A test plan run is for a specific version of a test plan, an AT, and a browser. The test plan run generates a report. That report is associated with the test plan version, not the test plan.

It is possible for the test queue to have runs for two different versions of the same test plan in it. Such a scenario could be:
* V1 of radio test plan is in the recommended phase. We add a run to the test queue for V1 to generate a report for a new version of a screen reader or browser.
* V2 of the radio test plan is in draft review. It may have several test plan runs in the queue to generate data needed for it to advance to candidate review.

[...]

PR #715, opened 2023-07-26
Updated Test Queue and Candidate Tests Page

howard-e:
This PR is based on work done in #713.

This updates the Test Queue and Candidate Tests page based on implications created #648 and ongoing discussions in #709.

**Checklist**

Test Queue
- [x] Update Test Queue rows' label for the _Mark as Candidate_ button to _Mark as Final_
- [x] Updated the functionality of the button to mark a the report item as "approved" for it to be included in any generated reports for the respective TestPlanVersion's phase
- [x] Added `approvedAt` column for `TestPlanReport`s.

Candidate Tests
- [x] Remove `Mark as ...` button
- [x] Remove `Change Target Date` button

Data Management
- [x] Added the content label to display time away or past the recommended target date. If the user is an admin, it is a button instead which displays a dialog to change the recommended phase target date for a version in CANDIDATE; this is a P1 requirement of #648

General
- [x] Remove `TestPlanReport.status` column
- [x] Update tests

___

_These changes can be viewed at https://aria-at-app-sandbox.bocoup.com_


server/resolvers/testPlanReportsResolver.js
```diff
@@ -10,12 +10,11 @@ const {
 
 const testPlanReportsResolver = async (
     _,
-    { statuses, testPlanVersionId, testPlanVersionIds, atId },
+    { phases = [], testPlanVersionId, testPlanVersionIds, atId, isApproved },
     context,
     info
 ) => {
```
alflennik, line 16: I ran this resolver in GraphQL and when I specified phases, I got an error.

I ran this query:

```
query {
  testPlanReports(phases: [RD]) {
    id
  }
}
```

Here was the error I got:

```
            "TypeError: Cannot read properties of undefined (reading 'phase')",
            "    at /Users/Shared/Projects/aria-at-app/server/resolvers/testPlanReportsResolver.js:68:60",
            "    at Array.filter (<anonymous>)",
            "    at Object.testPlanReportsResolver [as testPlanReports] (/Users/Shared/Projects/aria-at-app/server/resolvers/testPlanReportsResolver.js:67:32)",
            "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"
```

This happened no matter the phase I specified. I was using the test data.

server/graphql-schema.js
```diff
@@ -1039,14 +980,15 @@ const graphqlSchema = gql`
         testPlanVersion(id: ID): TestPlanVersion
         """
         Load multiple TestPlanReports, with the optional ability to filter by
-        status, atId and testPlanVersionId.
+        TestPlanVersionPhase, atId, testPlanVersionId and if the report is approved.
         See TestPlanReport type for more information.
         """
         testPlanReports(
-            statuses: [TestPlanReportStatus]
+            phases: [TestPlanVersionPhase]
```
alflennik, line 987: Since the phase is actually on the testPlanVersion calling it just "phases" seems like it might mislead someone looking at it. It's more wordy, but I feel like "testPlanVersionPhases" communicates that the resolver is going to examine the associated testPlanVersion to filter down the list.

We discussed this, but I feel like so much progress has been made in the UI and in the CG that we can potentially make some big improvements to the overall ergonomics of the GraphQL API now. That would obviously not apply to this PR though.

alflennik (changes requested):
This PR looks great! These are broad sweeping changes, and I can clearly see that you had to comprehensively sweep through the whole codebase redefining things to the new reality. 

I do have a few requested changes.

There is huge number of warnings getting outputted when the tests are run, would you be able to fix that?

I noticed that one of the tests seems to be failing, something about missing an test for the approvedAt field.

With the new "isFinal" functionality, it's possible to have a TestPlan on the TestPlanReports page which has zero data. I attached a screenshot below to illustrate what I mean.

The Select Only Combobox Example is showing as a row on the test reports table despite having None, None, None as data.

Despite that, I did verify that the "Mark as Final" buttons are working. That feels like a huge change!

I verified that the "Target N days away" and  "Target N days past" are working.

I went through the whole app as carefully as I could, and things are working great. I do think I should go through again after you address the feedback before I really feel confident that I tested everything and can approve.

server/migrations/20230608171911-moveTestPlanReportValuesToTestPlanVersion.js
```diff
@@ -1,14 +1,101 @@
...
+        const atLoader = AtLoader();
+        const browserLoader = BrowserLoader();
+
+        const compareTestContent = (currentTests, newTests) => {
+            const hashTest = test => hash(omit(test, ['id']));
+            const hashTests = tests => {
+                return Object.fromEntries(
+                    tests.map(test => [hashTest(test), test])
+                );
+            };
```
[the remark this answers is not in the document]
howard-e: Yep @alflennik you're right, that usage was missed. Addressed in https://github.com/w3c/aria-at-app/pull/715/commits/7bdfa8e8e5f185edf8f645f705b1a250371eb69f

server/resolvers/TestPlanVersionOperations/updatePhaseResolver.js
```diff
@@ -365,10 +366,13 @@ const updatePhaseResolver = async (
             archivedAtDate: null
```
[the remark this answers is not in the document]
howard-e: Agreed and thanks for raising that again. I've refactored the instances of `archivedAtDate` to be `deprecatedAt` in https://github.com/w3c/aria-at-app/pull/715/commits/b92d12dfebd00954c11d29638dafae6d889f0468

server/graphql-schema.js
```diff
@@ -372,9 +326,15 @@ const graphqlSchema = gql`
...
+        be a unique combination of AT + Browser + TestPlanVersion.
+
+        isApproved is used to check if a TestPlanReport has been "Marked as Final",
+        indicated by TestPlanReport.approvedAt existence.
+        None value indicates to return all.
+        True value indicates to return the reports which only have an approvedAt date.
+        False value indicates to return the reports which have no approvedAt date.
         """
-        testPlanReports(isCurrentPhase: Boolean): [TestPlanReport]!
+        testPlanReports(isApproved: Boolean): [TestPlanReport]!
```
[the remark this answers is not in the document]
howard-e: Yep! Change has been made in https://github.com/w3c/aria-at-app/pull/715/commits/b92d12dfebd00954c11d29638dafae6d889f0468

server/resolvers/testPlanReportsResolver.js
howard-e: This should now work as expected in https://github.com/w3c/aria-at-app/pull/715/commits/b92d12dfebd00954c11d29638dafae6d889f0468

server/graphql-schema.js
howard-e: Agreed with using that terminology, it's quite helpful. I've updated `server/resolvers/testPlanReportsResolver` and the schema in https://github.com/w3c/aria-at-app/pull/715/commits/b92d12dfebd00954c11d29638dafae6d889f0468 to account for this

howard-e:
> There is huge number of warnings getting outputted when the tests are run, would you be able to fix that?
> 
> I noticed that one of the tests seems to be failing, something about missing an test for the approvedAt field.
> 

This should be addressed.

> With the new "isFinal" functionality, it's possible to have a TestPlan on the TestPlanReports page which has zero data.

Thanks for catching that and I believe the same was reported in #724. I've pushed https://github.com/w3c/aria-at-app/pull/715/commits/12a176494797b1064cd0a51e7329625530166508 to account for this from 2 points. The client and server is verifying that there are valid (reports exist that are `isFinal`) reports during the `updatePhase` process. 

Additionally, the `/reports` page is also checking if there are valid testPlanReports linked to it before showing the row. This is needed since we still have reports in the system which have gotten to the CANDIDATE phase without accounting for having required reports and need to be shown after the migration.

> I went through the whole app as carefully as I could, and things are working great. I do think I should go through again after you address the feedback before I really feel confident that I tested everything and can approve.

Thanks for the thorough feedback and this should be ready for your review once again 🙂

server/graphql-schema.js
```diff
@@ -296,18 +260,8 @@ const graphqlSchema = gql`
         The title of the TestPlan at this point in time.
         """
         title: String
-        # TODO: Waiting for TestPlanVersionStatus to be needed
-        # status: TestPlanVersionStatus!
```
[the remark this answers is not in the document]
alflennik: I tried querying `isFinal: true` and I got no results:

Query for testPlanReports(isFinal: true) had zero results.

I'm not sure why this is happening. This is using a dump of the production database, and I do see results on the reports page and approvedAt is true for many reports in the database, so I wouldn't expect an empty array in this case.

`isFinal: false` did seem to work with data returned.

server/graphql-schema.js
```diff
@@ -1132,10 +1074,11 @@ const graphqlSchema = gql`
...
         deleteTestPlanRun(userId: ID!): PopulatedData!
         """
-        Update the report status. Remember that all conflicts must be resolved
-        when setting the status to CANDIDATE. Only available to admins.
+        Update the approvedAt date. This must be set before a TestPlanReport can
+        be advanced to CANDIDATE. All conflicts must also be resolved.
+        Only available to admins.
         """
-        updateStatus(status: TestPlanReportStatus!): PopulatedData!
+        updateApprovedAt: PopulatedData!
```
alflennik, line 1081: Could I suggest calling this just "approve"?

Or maybe "markAsFinal" possibly.

server/resolvers/findOrCreateTestPlanReportResolver.js
```diff
@@ -11,7 +11,7 @@ const findOrCreateTestPlanReportResolver = async (_, { input }, context) => {
     }
 
     const [testPlanReport, createdLocationsOfData] =
-        await getOrCreateTestPlanReport(input, { status: 'DRAFT' });
+        await getOrCreateTestPlanReport(input);
```
alflennik, line 14: Something that concerns me is that there is no way, even through the API, to move a report back to the queue after it's been marked final. Considering history, I can remember instances of wanting to pull content back to the queue.

I think there are multiple potential ways to solve this. One would be to add a special resolver to the testPlanReportOperations which would remove the approvedAt date. This would allow us to use the API to move content back to the queue.

Another solution could be here. It's probably too subtle, and should probably be documented with a comment, but in the past the status: "DRAFT" code above would move a report back to the queue if it already existed. That has come in handy multiple times. The equivalent with approvedAt would be setting approvedAt to null here.

client/components/TestQueueRow/index.jsx
```diff
@@ -530,30 +510,24 @@ const TestQueueRow = ({
...
-                                            );
+                                            await updateReportStatus();
                                         }}
                                     >
-                                        Mark as{' '}
-                                        {capitalizeEachWord(nextReportStatus, {
-                                            splitChar: '_'
-                                        })}
+                                        Mark as Final
                                     </Button>
```
alflennik, line 528: I noticed when I clicked this button to "Mark as Final" that it would work even if there were no actual results collected. That could cause confusion, because on the Data Management page, when advancing a TestPlanVersion to Candidate or Recommended, it does check whether any results were collected and fails. I think there should be a check before changing the status to make sure results actually exist.

server/resolvers/testPlanReportsResolver.js
alflennik: Verified, thanks.

alflennik (changes requested):
I said it before, but fabulous work. 

I said during my last review that I wanted to do a more methodical pass of the whole app, and now I've done that, so I think the comments here reflect me having gone over everything in detail. It feels pretty close now, just a few comments left.

client/components/CandidateTests/CandidateTestPlanRun/queries.js
```diff
@@ -37,7 +37,7 @@ export const CANDIDATE_REPORTS_QUERY = gql`
         }
         testPlanReports(
             atId: $atId
-            statuses: [CANDIDATE]
+            phases: [CANDIDATE]
```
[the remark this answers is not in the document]
howard-e: Good catch! This should now be resolved

server/graphql-schema.js
howard-e: Agreed and updated to make this clearer.

Related to this, there was a discussion around the column property's name at https://github.com/w3c/aria-at-app/issues/632#issuecomment-1652681110

> Option B: Change data element to "markedFinalAt"

I've updated the `approvedAt` property to be `markedFinalAt` instead for consistency across the client and server + database implementation with this operation

client/components/TestQueueRow/index.jsx
howard-e: Good catch and should have never worked like that. I've addressed this in https://github.com/w3c/aria-at-app/pull/715/commits/f904219eb943afff452f61138717d02c06184441

server/resolvers/findOrCreateTestPlanReportResolver.js
howard-e: > Something that concerns me is that there is no way, even through the API, to move a report back to the queue after it's been marked final. Considering history, I can remember instances of wanting to pull content back to the queue.
> 
> I think there are multiple potential ways to solve this. One would be to add a special resolver to the testPlanReportOperations which would remove the approvedAt date. This would allow us to use the API to move content back to the queue.

Good idea. I've pushed https://github.com/w3c/aria-at-app/pull/715/commits/9016c45f1e29c6d76ac669b6672d1ff59e3cdf87 which adds a new resolver to do that action

> Another solution could be here. It's probably too subtle, and should probably be documented with a comment, but in the past the status: "DRAFT" code above would move a report back to the queue if it already existed. That has come in handy multiple times. The equivalent with approvedAt would be setting approvedAt to null here.

I've already gone with a special resolver in this instance but also related and of note, another (and much more destructive operation here) is that when updating a TestPlanVersion's phase to "DRAFT" it will also nullify all the `approvedAt` properties for the connected `TestPlanReport`s

server/graphql-schema.js
alflennik: Great change! Thanks!

server/resolvers/findOrCreateTestPlanReportResolver.js
alflennik: Great change! I verified it's working.

server/graphql-schema.js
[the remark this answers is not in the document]
alflennik: Looks great

alflennik (approved):
Great work!

howard-e:
This PR now also includes the changes from #729 and #730 after the rebase.

PR #721, opened 2023-07-31
Fix chance for TestPlanVersion correct `updatedAt` to be wrong

howard-e:
There is a chance that should multiple updates be pushed to w3c/aria-at, only the date of the last one will be captured and that will then account for the latest update for any other tests that were updated.

Scenario:
* If w3c/aria-at/tests/alert is updated and pushed to the default branch at 12:03,
* Then w3c/aria-at/tests/checkbox is updated and pushed to the default branch at 12:07,
* Then w3c/aria-at/tests/vertical-temperate-slider is updated and pushed to the default branch at 12:44,
* But the next scheduled update for the imports-test script is at 1:00,
* Then `updatedAt` date for all 3 newly imported TestPlanVersions will be _the-date_ at 12:44
* 2 of those newly imported TestPlanVersions will be incorrect

That's because the `gitCommitDate` being used was referencing the latest found git date for entire w3c/aria-at directory at the time of the scheduled script, rather than navigating to each w3c/aria-at/tests/<pattern> and checking the most recently recorded time it was updated through git.

### DIFF
--- a/server/scripts/import-tests/index.js
+++ b/server/scripts/import-tests/index.js
@@ -51,10 +51,17 @@ const gitCloneDirectory = path.resolve(__dirname, 'tmp');
 const builtTestsDirectory = path.resolve(gitCloneDirectory, 'build', 'tests');
 const testsDirectory = path.resolve(gitCloneDirectory, 'tests');
 
+const gitRun = (args, cwd = gitCloneDirectory) => {
+    return spawn
+        .sync('git', args.split(' '), { cwd })
+        .stdout.toString()
+        .trimEnd();
+};
+
 const importTestPlanVersions = async () => {
     await client.connect();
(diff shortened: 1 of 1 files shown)


alflennik (approved):
I really appreciate this change, this was a pretty big gotcha with the TestPlanVersions and I'm impressed how small the fix was needed to address it. I ran all the code and verified that importing works. The next time the ARIA-AT team merge multiple PRs at the same time we should take a peek at the shas to verify that they are indeed split out correctly.

mcking65:
That will be happening a lot after we have the new V2 test format supported (w3c/aria-at#974) because we are going to be rewriting every test plan. So, @howard-e, this is a more important fix than you may have initially realized.

PR #722, opened 2023-07-31
Verify that the criteria for advancing the phase for test plans has been met by referencing required reports for a phase

Paul-Clue:
This PR implements a check in the `updatePhaseResolver` that verifies that the correct pairs of ats and browsers have been tested to meet the criteria for the candidate and recommended phases.


server/resolvers/TestPlanVersionOperations/updatePhaseResolver.js
```diff
@@ -253,6 +253,68 @@ const updatePhaseResolver = async (
...
+
+        const hasRequiredCandidateReports =
+            hasChromeJaws && hasChromeNVDA && hasSafariVoiceOver;
+
+        if (!hasRequiredCandidateReports) {
+            const missingCombinations = [
+                hasChromeJaws ? '' : 'Chrome and JAWS',
+                hasChromeNVDA ? '' : 'Chrome and NVDA',
+                hasSafariVoiceOver ? '' : 'Safari and VoiceOver'
+            ].flatMap(str => str);
```
howard-e, line 275:
```suggestion
            ].filter(str => str);
```

You may have wanted to use `filter` based on the truthy check callback? `flatMap` isn't doing anything here. 

Otherwise, it could end up with `missingCombinations` looking like `[ '', 'Chrome and JAWS', 'Safari and VoiceOver' ]` if reports exist for some of the combinations but not all. That would result in the error string displaying: `... have not been collected: , Chrome and NVDA, Safari and VoiceOver`, where there's a stray comma beside empty space due to the `.join()`.

Paul-Clue: Hey, @howard-e, you are 100% correct here. This has been fixed in the most recent build.

howard-e (changes requested):
@Paul-Clue thanks for this. It definitely unlocks a final piece for this new feature!

In the future, there will be a need for this to be managed by admins in some UI. So a lot of this logic I could see being managed by perhaps the `AtBrowsers` table in the future, but this makes it easy to work with that change in a future task.

Do note I've shared 2 blocking comments below.

server/migrations/20230808191138-addColumnsToAtBrowsers.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
```
howard-e, line 1: When importing the production database and running the migrations, I get the following error:

```
== 20230523163856-remove-OtherUnexpectedBehaviorText: migrating =======
=== Fixing unexpectedBehavior results for TestPlanRun:257 ===
=== Fixing unexpectedBehavior results for TestPlanRun:442 ===
=== Fixing unexpectedBehavior results for TestPlanRun:443 ===
=== Fixing unexpectedBehavior results for TestPlanRun:258 ===

ERROR: column browsers->AtBrowsers.isCandidate does not exist

error Command failed with exit code 1.
```

That's because the `remove-OtherUnexpectedBehaviorText` calls `conflictsResolver` which attempts to reference `AtBrowsers.isCandidate` (which wouldn't exist at that point) through some other connected calls and solving how to exclude that reference may not be worth it.

So I think this can be easily resolved in 2 ways:
1. Because the `AtBrowsers` table migration has not yet been included in production, you could include these columns in that migration. This way may be preferred OR
2. Renaming this file to something like `20230424152009-addColumnsToAtBrowsers` so it can be ran right after the `addAtBrowser` migration.

server/resolvers/TestPlanVersionOperations/updatePhaseResolver.js
```diff
@@ -253,6 +253,73 @@ const updatePhaseResolver = async (
         throw new Error('No test plan reports found.');
     }
 
+    // if (phase === 'CANDIDATE') {
```
howard-e, line 256: It seems this logic has been used elsewhere so you should be able to remove these commented out lines when ready.

howard-e (commented):
Hi @Paul-Clue, this looks good!

I also verified this works as expected by merging with #715 locally. I think once the failing tests are updated and my comment about the migration below is addressed, this should be good to go!

server/migrations/20230808191138-addColumnsToAtBrowsers.js
Paul-Clue: Thank you so much for this thorough review and the suggestions for fixing the issue! We decided to go with the first suggested resolution, which worked perfectly.

server/resolvers/TestPlanVersionOperations/updatePhaseResolver.js
Paul-Clue: Good catch! I deleted this huge comment.

server/resolvers/browsersResolver.js
```diff
@@ -1,9 +1,5 @@
-const { getBrowsers } = require('../models/services/BrowserService');
-
-const browsersResolver = () => {
-    return getBrowsers(undefined, undefined, undefined, undefined, undefined, {
-        order: [['name', 'asc']]
-    });
+const browserResolver = async (_, __, context) => {
```
howard-e, line 1: Minor nit: It seems the standard has been to keep these resolver functions as having the same name as their file names which is sometimes useful for doing quick searches.

Could this function be renamed back to `browsersResolver`?

Paul-Clue: Ok. Done!

howard-e (approved):
Good stuff @Paul-Clue and thanks for revising the initial implementation to make it much easier to include and support future UI to update the required reports!

PR #728, opened 2023-08-03
Test plan report status dialog

### LINKED ISSUE #649: Add Test plan report status dialog (P1)
_Taken from #648._

This page or dialog shows detailed status of AT/browser reports for a given version of a given test plan. The AT/browser combinations are limited to the AT covered by the plan and the most recent version of each in-scope browser included in the system browser list.

It includes:

- Links to completed reports
- Status of in-progress reports
- Information about missing reports
- Whether a specific report is required for the current phase
- The data is presented in a table captioned "PHASE_NAME TEST_PLAN_NAME test Plan Reports". Example: "Draft Alert Test Plan Reports".

The table has columns for:

1. required?: Shows "Yes" or "No"
2. AT: Shows AT Name and version
3. Browser: Shows browser name and version
4. Report Status: Shows:
    - If the report is complete, a link to the report with text "Report Completed on DATE"
    - If report is missing, the string "Missing", and if user is admin, button "Add to Test Queue"
    - If report is not complete and not missing and not in test queue, the string "None", and if user is admin, "Add to Test Queue" button
    - If report is in the test queue, the string "X% complete by Y testers with Z conflicts"
    - Note: Percent complete = NUMBER_COMPLETED_TESTS_BY_ALL_TESTERS / (NUMBER_ASSIGNED_TESTERS * NUMBER_TESTS_IN_PLAN) * 100

The table is sorted by required Yes, then required No, then AT, then browser.

stalgiag:
### Summary
This PR resolves #649 and implements the Test Plan Report Status Dialog as well as the Button that opens it. I am marking this as "Draft" pending clarifications for the items captured by the below notes.

**Notes**
* TestPlanReports that are in the DRAFT phase have incorrectly calculated metrics. I recall @howard-e  mentioning that this behavior is expected for now and that updates to this would need to come as part of the new changes currently being made. It is likely that `calculateTestPlanReportCompletionPercentage()` should not be a utility function (with a wildly long name) but should instead be a part of the calculation of `metrics` on both the front and backend. Should that be part of this PR?
* I am opening this PR to merge into `revised-data-management` due to being rebased on that branch. Unsure what the intended merge sequence is so please let me know if this should be changed.

**Optional notes**
* The unit testing could stand to be more thorough. Right now only static content rendering is being checked. Building the data mocks for all of  the TestPlanVersion permutations needed to correctly test the conditional renderings on this component is a big job and one that should likely start at the level of testing SummarizeTestPlanVersion. This feels out of scope for this PR but I could be wrong. Please let me know if I should expand testing beyond what is present here.

### Description of work
**Button Label**
* The label on the button that opens the dialog provides information about the current report status for the test plan. The button has one of the following labels.
  * - [x]  Label of 'Required Reports Not Started' when following condition is met: 'For every AT/browser pair that requires a report, the report is neither complete nor in the test queue.'
  * - [x] Label of 'Some Required Reports Missing' when following condition is met: 'At least one of the AT/browser pairs that requires a report neither has a complete report nor has a run in the test queue. At the same time, at least one of the AT/browser pairs that requires a report either has a complete report or has a run in the test queue.'
  * - [x] Label of 'Required Reports In Progress' when following condition is met: 'At least one AT/browser pair that requires a report does not have a complete report and is in the test queue. All other AT/browser pairs that require a report are either complete or are in the test queue.'
  * - [x] Label of 'Required Reports Complete' when following condition is met: 'All AT/browser pairs that require a report have a complete report.'

**Dialog Title**
* - [x] Should follow this template: "Report Status for TEST_PLAN_NAME Test Plan"

**Dialog Note content**
* - [x] Should follow this template: "This plan is in the PHASE_NAME phase. N AT/browser pairs require reports in this phase."

**Dialog Table Caption**
* - [x] Should follow this template: "Reports for PHASE_NAME TEST_PLAN_NAME Test Plan"

**Dialog Table Columns**
* Required
  * - [x] required?: Shows "Yes" or "No"
* AT
  * - [x] If phase is draft, shows AT name.
  * - [ ] If phase is recommended, Shows AT Name and version.
* Browser
  * - [x] If phase is draft, shows browser name.
  * - [ ] If phase is recommended, shows browser name and version.
* Report Status
  * - [x] If the report is complete, shows a link to the report with text "Report Completed on DATE"
  * - [x] If report is missing, shows the string "Missing", and if user is admin, button "Add to Test Queue"
  * - [ ] If report is not complete and not missing and not in test queue, shows the string "None", and if user is admin, "Add to Test Queue" button
  * - [x] If report is in the test queue, shows the string "X% complete by Y testers with Z conflicts"
* Sorting
  * - [x] The table is sorted by required Yes, then required No, then AT, then browser.
  

---
  - https://app.asana.com/0/0/1204829210885376


howard-e:
> * TestPlanReports that are in the DRAFT phase have incorrectly calculated metrics. I recall @howard-e  mentioning that this behavior is expected for now and that updates to this would need to come as part of the new changes currently being made. It is likely that `calculateTestPlanReportCompletionPercentage()` should not be a utility function (with a wildly long name) but should instead be a part of the calculation of `metrics` on both the front and backend. Should that be part of this PR?

Well it's not so much that it would be incorrect, but metrics done against TestPlanReports are part of a DRAFT TestPlanVersion could have unknown results. The metrics up until now only considered a "final" test plan result after all conflicts were resolved. So this requirement may be able to use that same metrics calculator (with some tweaks), but I have a feeling that we'd want it to be separate, like `calculateTestPlanReportCompletionPercentage()` function as you have it here, so that's fine.

> (with a wildly long name)

But descriptive! 🙂

> * I am opening this PR to merge into `revised-data-management` due to being rebased on that branch. Unsure what the intended merge sequence is so please let me know if this should be changed.

This is fine for now, until #713 can be merged into `main`

> * The unit testing could stand to be more thorough. Right now only static content rendering is being checked. Building the data mocks for all of  the TestPlanVersion permutations needed to correctly test the conditional renderings on this component is a big job and one that should likely start at the level of testing SummarizeTestPlanVersion. This feels out of scope for this PR but I could be wrong. Please let me know if I should expand testing beyond what is present here.

A standalone test of the `calculateTestPlanReportCompletionPercentage` utility would be good to have.

alflennik (changes requested):
Thank you for the PR Stalgia!! I'm so impressed that you were able to quickly whip this up with very little direct supervision, not just using the stack of tools well but also deftly navigating the labyrinth of technical terms associated with the project.

I also think it speaks well to the modular way you implemented this that you were able to easily port it from its original spot in the reports page over to the data management page.

Feedback-wise it's mostly just a bunch of micro-details but overall it's really usable and really far along and I'm happy with it.

- I noticed after adding a report that the table didn't automatically update and still says the report is missing. It would be sweet if we could refresh the contents of the table, what do you think?

- This is not really a request to change anything, but I felt like the loading indicator flashed really quickly. Which is a good problem to have. Probably on production it'll be slightly slower.

- I tested this on VoiceOver and NVDA. When going from the secondary modal that confirms success of adding the report back to the status modal, the focus gets out of the modal and I can't get it back to the modal. Would it be possible to focus the button or cell which originally opened the modal when it closes?

- Focus-wise, I don't know if you implemented this, but it would be great when closing the status modal to focus the button which originally opened the modal.

Side note, the fact that both members of the public and admins can click that cell and see an interactive dialog is pretty amazing interactive magic. And for admins, being able to add Test Plans to the Test Queue right there ... that's kind of a game changer and I feel like going forward that's going to be the way they add test plans in like 90% of cases.
 
- On NVDA when clicking the add to test queue button and opening the secondary modal, it said a bunch of stuff really fast, I felt like almost there were two modals flashing by really quickly. I feel like it loads a bit too fast to need to have the intermediate "Adding test plan to test queue" state.

- Visually, the button that opens the modal looks different when signed out vs. signed in. When logged out the button is floating in the middle of the cell instead of sitting at the bottom.

- I've noticed in the past that ATs would sometimes not "see" the spaces in HTML. I confused about how this happens. But this is the output in NVDA:
```
This plan is in theDRAFTReview phase. 3 AT/browser pairs require reports in this phase.
```

Do you see that the spaces are missing around the word DRAFT? I think we might need to add a `&nbsp;` instead of a `{' '}` to fix this?

- Another point about spacing, I think also there is a double space in the modal title before the bolded test plan name.

- This is a very minor point, but we did an accessibility audit around a year ago (?) and one of the notes we got was that we shouldn't use all caps for statuses like "DRAFT". The reason they gave is that people with dyslexia recognize words more by their shape than the letters themselves and so all caps would interrupt that. 

- The table "name" which spans multiple columns seemed to confuse NVDA, which treated that singular column kind of like a label for all the columns below it. Could we pull that out of the actual table?

So yeah, as you can see it's just a lot of small stuff. Overall it's really great work. Thanks Stalgia and let me know if you need any clarifications, I'm around.

stalgiag:
Okay I think I got almost everything that you listed @alflennik . There were a few items that warrant responses:

> On NVDA when clicking the add to test queue button and opening the secondary modal, it said a bunch of stuff really fast, I felt like almost there were two modals flashing by really quickly. I feel like it loads a bit too fast to need to have the intermediate "Adding test plan to test queue" state.

I agree. The loading modal generated by `triggerLoad` does flash by quickly which is disorienting with a screenreader.  This use of the `LoadingStatus` exists in `main` in several instances including the previous implementation of the button that adds a test to the queue. This is why I used it here. The problem isn't as noticeable in `main` because the buttons that cause mutations force an immediate re-render ( #731 ). 

I think a solve for this would require addressing all of the instances where `triggerLoad` causes the `LoadingStatus` component to flash on screen. Possible solutions would be to use an `Alert` instead of a `Modal` and to enforce a minimum display duration. That might be better done in a different PR but I would be happy to add to this one if preferred.

> I noticed after adding a report that the table didn't automatically update and still says the report is missing. It would be sweet if we could refresh the contents of the table, what do you think?

I made changes that fix #731 and now a rerender happens but unfortunately, due to location in the tree for the `refetch` and the way that state is maintained upon data updates (parent component resets state), this causes a rerender of the entire page. This is how this is handled elsewhere in the codebase by buttons that cause mutations in `main`. With the changes in this PR, the mutations happen after the confirmation modal is closed rather than while it is opening.

To enable more targeted rerenders, I believe we would need to do architectural changes to the components on the `/data-management/` path to make smaller components and add more data-fetching nodes in the tree. This would be a significant refactor and is probably beyond scope for this work.

stalgiag:
Note that the CI testing seems to be failing consistently today. I don't think it is specific to this branch. This was happening earlier today and I was able to pass by triggering a re-run. The only commit that is failing has nominal changes and it doesn't fail on linting or client unit tests but rather server unit tests. I will investigate tomorrow.

stalgiag:
Fixed issue where the button was rendering for Recommended phase Test plans as well @alflennik. Thanks again for catching that!

mcking65:
@stalgiag 

> Fixed issue where the button was rendering for Recommended phase Test plans as well @alflennik. Thanks again for catching that!

We do need to access report status for recommended reports. It is part of the specification in #648. Once a report is recommended, we are generating reports like crazy ... every time there is a new browser or screen reader version. When new versions are added to the system, the button will need to reflect whether there is a missing report because of that. There is one to-do noted in #648 that is on me. I have not fully defined the algorithm for calculating which reports are required for a recommended test plan. So, we probably need a stopgap for that until I write that down. I guess I better get to it!!

alflennik:
@mcking65 that's very valuable context! Thank you! We discussed this and came to the conclusion that it would be easier to get this initial version in and then address the missing case in a follow up PR.

alflennik:
@stalgiag the modals are looking great! The only thing usability-wise that's jumping out to me is what happens when you add a test plan report. It exits back to the data management page, and you have to navigate down the page and push the button again to get back to the modal. I feel like in a previous version it was possible to stay in the modal so you can add multiple reports in a boom-boom-boom one-by-one fashion. Is that still on the table?

stalgiag:
That feeling that the page is reloading is an unfortunate consequence of what I was describing here:

> I made changes that fix https://github.com/w3c/aria-at-app/issues/731 and now a rerender happens but unfortunately, due to location in the tree for the refetch and the way that state is maintained upon data updates (parent component resets state), this causes a rerender of the entire page. This is how this is handled elsewhere in the codebase by buttons that cause mutations in main. With the changes in this PR, the mutations happen after the confirmation modal is closed rather than while it is opening.

> To enable more targeted rerenders, I believe we would need to do architectural changes to the components on the /data-management/ path to make smaller components and add more data-fetching nodes in the tree. This would be a significant refactor and is probably beyond scope for this work.

That complete rerender is similar to what happens with the normal add test to queue button but I can see how it is less obtrusive in that context. If we want to both immediately rerender the updated status of a test plan report that has been added to queue AND not rerender the whole tree, the only option we have without significant changes is to store a clone of the testPlanReports in state on TestPlanReportDialog and mutate that clone when making the changes. This is usually not advised (more than one source of truth) but it would give us that immediate rerender of the modal without rerendering the entire page and would be a small change with no needed major architectural shifts.

stalgiag:
Added a data query specific to the TestPlanReportStatusDialog. This fixes the issue outlined in the above conversation and allows re-rendering on test queue mutation without re-rendering the entire page. This will duplicate the fetching of some data but it is acceptably small ( about 1kb per test plan version).

client/components/TestPlanReportStatusDialog/index.jsx
```diff
@@ -0,0 +1,303 @@
...
+                );
+        }
+    };
+
+    const renderReportStatus = testPlanReport => {
+        const { metrics, draftTestPlanRuns, at, browser, id } = testPlanReport;
+        if (metrics) {
+            const percentComplete =
+                calculateTestPlanReportCompletionPercentage(testPlanReport);
+            if (percentComplete === 100) {
```
howard-e, line 185: We may want to check for the `testPlanReport.markedFinalAt` property here instead.

And it may also play a little better into the process because of a blocking bug I noticed. I was testing this out by:
1. completing all the results for testers assigned to a row in the Test Queue and then,
2. coming back to this dialog, where I'd now see that the `Report completed on <date>` link was available
3. clicked it and it went to the 404 page

That's because the link won't render unless the test plan report's conflicts are resolved, which the `markedFinalAt` property here would represent.

client/components/TestPlanReportStatusDialog/index.jsx
```diff
@@ -0,0 +1,303 @@
...
+                let currentDate = new Date(currentResult.completedAt);
+                return currentDate > latestDate ? currentResult : latestResult;
+            },
+            { completedAt: new Date(0) }
+        );
+
+        return mostRecentTestResult;
+    };
+
+    const getFormattedDate = dateString => {
```
howard-e, line 129: Could we make use of the `utils/formatter#convertDateToString` function instead? You should be able to pass the same `MMM DD, YYYY` format to it.

howard-e (changes requested):
Thanks @stalgiag. The modals and the logic here is looking great!

I've only requested changes because of a blocking issue I noticed below, along with some thoughts on that.

client/components/TestPlanReportStatusDialog/index.jsx
howard-e: A follow up thought I just had (and verified) is about what happens if a test plan report is marked as final, and the TestPlanVersion is still in the DRAFT phase?

We could theoretically show the report page it generates, with the same "Unapproved Report" disclaimer but I'm unsure if we want to do that, and it doesn't feel like that should work. I can't tell with certainty from the issue description if this is expected, unless I've missed something but If that's the case, the condition must also check that the version's phase is CANDIDATE or RECOMMENDED as well.

client/components/TestPlanReportStatusDialog/index.jsx
```diff
@@ -0,0 +1,303 @@
...
+        let options = { year: 'numeric', month: 'short', day: 'numeric' };
+        let formattedDate = new Intl.DateTimeFormat('en-US', options).format(
+            date
+        );
+        return formattedDate;
+    };
+
+    const renderCompleteReportStatus = (draftTestPlanRuns, id) => {
+        const mostRecentRun = getMostRecentTestPlanRun(draftTestPlanRuns);
+        const formattedDate = getFormattedDate(mostRecentRun.completedAt);
```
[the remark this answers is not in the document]
stalgiag: Great to know that is there now. I like it much better as well.

client/components/TestPlanReportStatusDialog/index.jsx
stalgiag: I check for `markedFinalAt && (phase === RECOMMENDED || phase === CANDIDATE)`. Thanks for the suggestion!

client/components/TestPlanReportStatusDialog/index.jsx
stalgiag: Good catch. Didn't see that one but very happy to get to use moment.js here.

stalgiag:
@howard-e I addressed your feedback in the latest commit. Thanks for the helpful tips! I always love when a review suggests a path to simplifying code.

alflennik (approved):
I'm really excited to see this make it in!

howard-e:
> @howard-e I addressed your feedback in the latest commit. Thanks for the helpful tips! I always love when a review suggests a path to simplifying code.

Thank you for addressing it! 💯

howard-e (approved):

PR #739, opened 2023-08-07
Update deduplicate migration

### LINKED ISSUE #579: Add Test Plans to the Test Queue: Plan version dropdown shows commits that do not include a change in the chosen test plan
Using the [sandbox on March 25, 2023](https://aria-at-app-sandbox.bocoup.com/test-queue).

When adding a plan to the test queue, the system is surfacing commits that do not include changes to the selected plan.
This enables generation of significant downstream data integrity problems.

Test plans are versioned by their last date of change as derived from the aria-at commit history.
The test plan version is critical information when running a plan through the working mode.
For example, it is esential to know what version of a plan was run to generate a specific result set and when the plan was last changed.

To reproduce the problem:
1. Open the add test plan experience
2. Choose alert
3. Note the list of versions includes every commit to the main branch in w3c/aria-at.
4. Look at the file tree for aria-at.
    As of March 25, 2023, The last commit that change the
    [Alert Test Plan Files](https://github.com/w3c/aria-at/tree/master/tests/alert)
    was [0928bcf on December 8, 2022](https://github.com/w3c/aria-at/commit/0928bcf530efcf4faa677285439701537674e014).

Actual behavior: Commits that do not include changes to the alert plan are listed in the version dropdown.

Expected behavior: Only commits to main with changes to the alert plan are listed.

Because of this problem, I was able to add multiple versions of the alert test plan to the test queue for VoiceOver with Chrome. The VoiceOver/Chrome table shows two rows for alert, one with a March 20 version and another with a March 8 version. This is a serious problem, because it indicates the plan has changed since December 8, but it has not changed. The ramifications of signaling change when in fact there was no change poses major risk to our ability to execute the working mode.

howard-e:
Resolves #579.

This updates the de-duplication migration to check the batches of git commits, and only save the test plan versions which match up with a valid git sha, referenced through the GitHub API.

This will make it so ONLY known git commits (in the database at this time) are shown for respective Test Plans, which addresses #579. Note that this PR does not back-populate any commits that may have been missed.

~**NOTE**
This causes data to be lost for the entire rows of Action Menu Button Example Using aria-activedescendant and Toggle Button when comparing against the [production /reports page](https://aria-at.w3.org/reports), because those reports were created with test plan versions that weren't valid and couldn't be easily updated by this migration.~

~Drafting to determine how the migration can avoid causing the data loss, or we may need to consider manually updating that data separately~

### DIFF
--- a/server/migrations/20230501220810-deduplicateTestPlanVersions.js
+++ b/server/migrations/20230501220810-deduplicateTestPlanVersions.js
@@ -557,11 +557,11 @@ module.exports = {
                 // Update TestPlanVersion -> TestPlanReport fkey to add cascade deletion on
                 // TestPlanVersion row deletion
                 await queryInterface.sequelize.query(
-                    `alter table public."TestPlanReport"
-                        drop constraint "TestPlanReport_testPlan_fkey";
+                    `alter table "TestPlanReport"
+                            drop constraint "TestPlanReport_testPlan_fkey";
 
-                     alter table public."TestPlanReport"
-                        add constraint "TestPlanReport_testPlan_fkey" foreign key ("testPlanVersionId") references public."TestPlanVersion" on update cascade on delete cascade;`,
+                         alter table "TestPlanReport"
+                            add constraint "TestPlanReport_testPlan_fkey" foreign key ("testPlanVersionId") references "TestPlanVersion" on update cascade on delete cascade;`,
                     {
                         transaction
                     }
@@ -579,11 +579,11 @@ module.exports = {
                 // Update TestPlanVersion -> TestPlanReport fkey to remove cascade delete on
                 // TestPlanVersion row deletion
                 await queryInterface.sequelize.query(
-                    `alter table public."TestPlanReport"
-                        drop constraint "TestPlanReport_testPlan_fkey";
+                    `alter table "TestPlanReport"
+                            drop constraint "TestPlanReport_testPlan_fkey";
 
-                     alter table public."TestPlanReport"
-                        add constraint "TestPlanReport_testPlan_fkey" foreign key ("testPlanVersionId") references public."TestPlanVersion" on update cascade;`,
+                         alter table "TestPlanReport"
+                            add constraint "TestPlanReport_testPlan_fkey" foreign key ("testPlanVersionId") references "TestPlanVersion" on update cascade;`,
                     {
                         transaction
                     }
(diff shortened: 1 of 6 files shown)


howard-e:
Importing a commit entry for https://github.com/w3c/aria-at/commit/022340081280b8cafb8ae0716a5b67e9ab942ef4 with the import script, into `TestPlanVersion` before running the migration will allow the data to be preserved for `Toggle Button`. Haven't found the same for `Action Menu Button Example Using aria-activedescendant` as yet.

server/migrations/20230501220810-deduplicateTestPlanVersions.js
```diff
@@ -278,6 +282,215 @@ module.exports = {
                 );
         };
 
+        /**
```
alflennik, line 285: Beast mode starts here lol

alflennik (commented):
I'll wait for you to mark this as ready to review but the direction looks good 👍

howard-e:
After making sure an entry exists for [the Toggle Button version](https://github.com/w3c/aria-at/commit/022340081280b8cafb8ae0716a5b67e9ab942ef4) and [the Action Menu Button Example Using aria-activedescendant](https://github.com/w3c/aria-at/commit/7c4b5dce23c74fcf280ed164bdb903e02e0e7726), which the migrated reports will be linked with, this should work without any data loss.

alflennik (approved):
It's actually pretty magical to see that every single commit is now directly referring to the test plan it's paired with. This seems like a big win for the usability of the app. I did find one commit that seemed to be miscategorized, but when I looked into it I saw that it did indeed change the source code for two test plans, so the behavior was expected. I verified it working with a fresh database and with the clone of the production database. Thanks for getting this PR over the finish line, it's even more polished than I imagined possible when I implemented the first version of this.

howard-e:
@alflennik thanks for the review!

I also noticed something else which is that with `TestPlanVersion.tests[].viewers` also being updated, it had to be ignored in the `hashTests` function or we could risk an import being duplicated again as the hash could be different if viewers existed for that row.

Using the following query on the latest production data, 5 rows were found before the migration and 2 rows after. 3 of those rows were meant to be correctly removed so we don't need to do anything additional here. But I mention it in case any future migrations or functions require moving of existing viewers on a TestPlanVersion to another:

```sql
select *
from "TestPlanVersion"
where exists ( select 1 from jsonb_array_elements(tests) as test where jsonb_array_length(test -> 'viewers') > 0 );
```

I also think in the future, we may want to have viewers being its own column but that's a separate discussion.

PR #749, opened 2023-08-21
Explicitly support 'DEPRECATED' phase for `TestPlanVersion.phase`

### LINKED ISSUE #741: Should not be able to add deprecated test plan versions to the test queue
Problem observed in staging.

In staging, the color viewer slider test plan has only one active version (V22.04.14). All earlier versions are deprecated.

image of data management page showing color viewer slider row

It should not be possible to add versions earlier than V22.04.14 to the test queue because they are all deprecated. Yet, the deprecated versions show up in the add to queue dialog.

image of add to queue dialog showing deprecated versions of slider test plan

howard-e:
Closes #741

Currently, the 'deprecated' phase can be derived within the client. But it can be be easier to have that explicitly set so the state of the database is represented better, and so special queries and resolvers can be avoided.

Deriving in-app evaluations such as figuring out when a deprecation happened, such as for the Test Plan Version History page also proves to be more manageable in this instance.

This PR updates the usage of the 'DEPRECATED' phase for:
- [x] TestPlanVersionsPage
- [x] Preventing deprecated (and R&D) test plan versions from being shown in the data management dropdown

### DIFF
--- a/client/components/DisclaimerInfo/index.jsx
+++ b/client/components/DisclaimerInfo/index.jsx
@@ -70,6 +70,14 @@ const recommendedMessageContent = (
     </>
 );
 
+const deprecatedTitle = 'Deprecated Report';
+const deprecatedMessageContent = (
+    <>
+        The information in this report is generated from previously set
+        candidate or recommended tests.
+    </>
+);
+
 const content = {
     CANDIDATE: {
         title: candidateTitle,
@@ -78,6 +86,10 @@ const content = {
     RECOMMENDED: {
         title: recommendedTitle,
         messageContent: recommendedMessageContent
+    },
+    DEPRECATED: {
+        title: deprecatedTitle,
+        messageContent: deprecatedMessageContent
     }
 };
 
--- a/client/components/DisclaimerInfo/index.jsx
+++ b/client/components/DisclaimerInfo/index.jsx
@@ -84,8 +84,9 @@ const content = {
 const DisclaimerInfo = ({ phase }) => {
     const [expanded, setExpanded] = useState(false);
 
-    const title = content[phase].title;
-    const messageContent = content[phase].messageContent;
+    const title = content[phase]?.title || content.CANDIDATE.title;
+    const messageContent =
+        content[phase]?.messageContent || content.CANDIDATE.messageContent;
 
     return (
         <Container>
--- a/client/components/DataManagement/queries.js
+++ b/client/components/DataManagement/queries.js
@@ -25,7 +25,7 @@ export const DATA_MANAGEMENT_PAGE_QUERY = gql`
             directory
             title
         }
-        testPlanVersions {
+        testPlanVersions(phases: [RD, DRAFT, CANDIDATE, RECOMMENDED]) {
             id
             title
             phase
--- a/client/components/ManageTestQueue/index.jsx
+++ b/client/components/ManageTestQueue/index.jsx
@@ -206,13 +206,18 @@ const ManageTestQueue = ({
                 item =>
                     item.title === retrievedTestPlan.title &&
                     item.testPlan.directory ===
-                        retrievedTestPlan.testPlan.directory
+                        retrievedTestPlan.testPlan.directory &&
+                    item.phase !== 'DEPRECATED' &&
+                    item.phase !== 'RD'
             )
             .sort((a, b) =>
                 new Date(a.updatedAt) > new Date(b.updatedAt) ? -1 : 1
             );
         setMatchingTestPlanVersions(matchingTestPlanVersions);
-        setSelectedTestPlanVersionId(matchingTestPlanVersions[0].id);
+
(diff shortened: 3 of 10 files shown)


client/components/DisclaimerInfo/index.jsx
```diff
@@ -84,8 +84,9 @@ const content = {
 const DisclaimerInfo = ({ phase }) => {
     const [expanded, setExpanded] = useState(false);
 
-    const title = content[phase].title;
-    const messageContent = content[phase].messageContent;
+    const title = content[phase]?.title || content.CANDIDATE.title;
```
howard-e, line 87: Needed in the instance that a `TestPlanVersion`'s phase is marked as `DEPRECATED`, but reports did exist if it was deprecated during the candidate or recommended phase.

It could be considered as 'unapproved' in that case but should another copy be defined for this instead? Or some alert that this set of reports has been deprecated?

stalgiag: My personal opinion is that different copy should be defined for `TestPlanVersion`'s with a phase of `DEPRECATED`. The copy could be similar to the `CANDIDATE` copy but could mention the deprecation, and potentially one day point to the current related `TestPlanVersion` if it exists.

Making different copy for `DEPRECATED` would also make this code easier to understand without context.

stalgiag (commented):
With these changes, the related code is much easier to understand with limited context. Great work!

I manually updated two test plan versions in my DB to have a phase of `DEPRECATED` and a `deprecatedAt` of `NOW()`. I did some manual testing and everything appears to be rendering with one exception, the "Overall Status" in the Data Management page table renders nothing. Since this PR adds explicit support for the `DEPRECATED`, it might be good to handle that case in `renderCellForPhase()` in `DataManagementRow`. The same `phasePill` color used on the `TestPlanVersionsPage` and a `deprecated at ...` string would feel natural here.

Another small comment inline in response to your question.

client/components/DisclaimerInfo/index.jsx
howard-e: > My personal opinion is that different copy should be defined for `TestPlanVersion`'s with a phase of `DEPRECATED`. The copy could be similar to the `CANDIDATE` copy but could mention the deprecation, and potentially one day point to the current related `TestPlanVersion` if it exists.

Yep, agreed. That would be good.

> Making different copy for `DEPRECATED` would also make this code easier to understand without context.

Thanks for your thoughts there. I agree, will draft something share.

howard-e:
> ...with one exception, the "Overall Status" in the Data Management page table renders nothing. Since this PR adds explicit support for the `DEPRECATED`, it might be good to handle that case in `renderCellForPhase()` in `DataManagementRow`. The same `phasePill` color used on the `TestPlanVersionsPage` and a `deprecated at ...` string would feel natural here.

Thanks for catching that and the suggestion. Will look into it!

howard-e:
> I did some manual testing and everything appears to be rendering as expected with one exception, the "Overall Status" in the Data Management page table renders nothing. Since this PR adds explicit support for the `DEPRECATED`, it might be good to handle that case in `renderCellForPhase()` in `DataManagementRow`. The same `phasePill` color used on the `TestPlanVersionsPage` and a `deprecated at ...` string would feel natural here.

@stalgiag looking over #648, and considering the flow of information, I don't think this condition should ever be possible. Versions are ideally only deprecated by being "overwritten" by a more recent version. So a deprecated test plan version should never be able to be shown at the top level (unless something was deleted directly from the database I suppose and that shouldn't happen). With this in mind, do you still think we should support displaying this?

If you agree with this, I'm thinking to be even safer, we could make sure the DataManagement page excludes the `DEPRECATED` phases during the query so there's no chance of any being displayed

stalgiag:
> > I did some manual testing and everything appears to be rendering as expected with one exception, the "Overall Status" in the Data Management page table renders nothing. Since this PR adds explicit support for the `DEPRECATED`, it might be good to handle that case in `renderCellForPhase()` in `DataManagementRow`. The same `phasePill` color used on the `TestPlanVersionsPage` and a `deprecated at ...` string would feel natural here.
> 
> @stalgiag looking over #648, and considering the flow of information, I don't think this condition should ever be possible. Versions are ideally only deprecated by being "overwritten" by a more recent version. So a deprecated test plan version should never be able to be shown at the top level (unless something was deleted directly from the database I suppose and that shouldn't happen). With this in mind, do you still think we should support displaying this?
> 
> If you agree with this, I'm thinking to be even safer, we could make sure the DataManagement page excludes the `DEPRECATED` `TestPlanVersion`s during the query so there's no chance of any being displayed.

This makes sense to me. I would be in favor of excluding `DEPRECATED` `TestPlanVersion`s from being displayed.

stalgiag (approved):

mcking65:
that is correct, deprecated versions should never show on data management. they will only be visible from the test plan versions page.

Issue #678, opened 2023-06-08
Add filters for the test plan status summary table on the data management page (P2)


lolaodelola:
_Related to #648: Filters._ 

Add a set of quick filter buttons above the table. the label on Each filter will show the filter name and number of table rows it will render.

The filter buttons are:

1. All Plans
2. R&D Complete
3. In Draft Review
4. In Candidate Review
5. Recommended Plans

Note: Because each plan may have multiple versions moving through the process, i.e., a version active in more than one phase at the same time, the set of rows shown by each filter is not unique. Thus, if the label on the all plans button were "All Plans 65", the sum of the numbers shown in the labels of the other buttons will likely be greater than 65.

mcking65:
@stalgiag and @howard-e, thank you; the filters are working well in staging. We can close this issue after deployment.

Note that the data in staging doesn't enable full testing. Because none of the test plans have multiple active versions, I can't test all the filter requirements. We will have sufficient conditions in production sometime in October.

ccanash:
Released to prod

PR #750, opened 2023-08-21
Filter and sort functionality for Data Management table

stalgiag:
## Overview

This PR adds support for sorting and filtering Test Plans in the Data Management table as outlined in #678 and #648. The design of the UI follows [the mockups provided in this comment](https://github.com/w3c/aria-at-app/issues/648#issuecomment-1678065550).

## Notes

- The sorting and filtering logic was broken out into custom hooks to avoid dramatically inflating the size of the `DataManagement` component.
- The [APG sortable table example ](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) was followed for the creation of the `SortableTableHeader` component. When testing with VoiceOver for macOS, I noticed that there is no dynamic announcement when the`aria-sort` attribute of the currently focused header is changed. I believe this is an issue with VoiceOver as the output is identical on the APG example.
- All sorting does primary active sorting first and then alphabetical by title secondarily. The sorting by AT is currently a stub that ends up doubling the behavior of the title sorting. I believe this can be changed once other work lands and test plans do not all have an identical list of ATs.
- This PR relies heavily on string comparison. To avoid problems created by string typos, I added a `constants.js` file with enums. I did not see this pattern anywhere else in the codebase but in my experience it helps avoid bugs that are difficult to diagnose with string comparison. That said, the source of truth for some of those constants, especially `TEST_PLAN_VERSION_PHASES`, should likely be on the server once the dust has settled and other work has been merged. Let me know if this pattern is unwelcome or previously decided against, I would be happy to change it.


howard-e:
@stalgiag this looks super good! I've been super eager to see filters and sorting being included since the initial days of the Test Queue page so thank you! 🙂

I took an initial pass at it, functionality wise and noticed a few things that may need to be addressed first:

Screenshot showing filter buttons and top of Data Management Page; additional details follow this image

In the above screenshot, there are 2 things to note:
* The values being calculated for the filter buttons may need to be checked again? It notes that there are `34 R&D Complete` rows but just `1 In Candidate Review` row. The screenshot shows I have at least 2 Candidate rows and my database instance during this test has 10 Candidate Test Plan Versions overall.
* This is a fresh load of the page, but it seems the default sort order hasn't been maintained according to:

  > **Default Table Sort** (from #648)
  > * First sort by overall status descending: recommended plans, then candidate plans, then draft plans, then R&D complete plans.
  > * Then sort by test plan name ascending.
  > * Then sort by Covered AT ascending
  * Shown in the screenshot is that the initial table rows are:
    * `Link Example 1 ...` in Candidate, then 
    * `Action Menu Button Using aria-activedescendant` in Candidate, then
    * not in the screenshot is `Action Menu Button Example Using element.focus()` in R&D, then
    * `Alert Example` in Candidate
    * ... and so on with mixing of the R&D Candidate rows
  * It does seem the alphabetical order is maintained outside of the first row though

Screenshot shows Recommended Plans filter option being selected when there are 0 known results. With no data to show, the column titles of the first 4 columns intersect

The above screenshot shows the `Recommended Plans filter option` being selected but there are 0 known results. With no data to show, the column titles of the first 4 columns are intersecting. This could be a consequence of the css doing anything along the lines of using fit-content or auto width or some variation of having the cell's content set the width, with 1 or more of those columns.

I do think a better experience here could be to show a message that there is no data to show, rather than rendering an empty table though.

stalgiag:
Thanks for checking it out! Looks like I missed most of these issues because of the limited variety present in my local DB which is usually just the blank reseeded default. This likely won't be a problem once I am able to use the migrated production DB but in the meantime I might go ahead and make myself a few different test DBs so I can catch problems like these sooner.

> * The values being calculated for the filter buttons may need to be checked again? It notes that there are `34 R&D Complete` rows but just `1 In Candidate Review` row. The screenshot shows I have at least 2 Candidate rows and my database instance during this test has 10 Candidate Test Plan Versions overall.

Interesting. Will modify my DB to reproduce.

> This is a fresh load of the page, but it seems the default sort order hasn't been maintained according to ...

The code follows the sort order so I think that is a bug specific to that first Test Plan Version. I'll check it out.

> The above screenshot shows the `Recommended Plans filter option` being selected but there are 0 known results. With no data to show, the column titles of the first 4 columns are intersecting. This could be a consequence of the css doing anything along the lines of using fit-content or auto width or some variation of having the cell's content set the width, with 1 or more of those columns.

> I do think a better experience here could be to show a message that there is no data to show, rather than rendering an empty table though.

I agree. Perhaps we should even take it a step further and not show the filter button for a phase with out any test plan versions.

stalgiag:
Okay updated the functionality as per our discussion @howard-e . I broke out a lot of logic for sorting by phase and deriving overall phase for a test plan into custom hooks. This could open up some opportunities for refactoring in `DataManagementRow` but I will leave that code as-is for now.

I still want to write new unit tests for the additional hooks but this can be reviewed at any time.

client/components/DataManagement/hooks.js
```diff
@@ -0,0 +1,232 @@
+import { useMemo, useState } from 'react';
```
alflennik, line 1: Great idea splitting out the logic into hooks. I definitely appreciate that idea.

I was thinking maybe can rename the file to refer to the idea of filtering and sorting instead of the more general "hooks.js"? 

If we end up another complex feature like this where it would be too complicated to dump it into the component, in that case it seems like we would want to create a separate file instead of intermingling it in this hooks file.

client/utils/constants.js
```diff
@@ -0,0 +1,22 @@
...
+    RD: 'RD',
+    DRAFT: 'DRAFT',
+    CANDIDATE: 'CANDIDATE',
+    RECOMMENDED: 'RECOMMENDED'
+};
+
+export const DATA_MANAGEMENT_TABLE_FILTER_OPTIONS = {
+    ALL: 'ALL',
+    ...TEST_PLAN_VERSION_PHASES
+};
```
alflennik, line 22: I'm down with using enums like this. I do think there's a place on the server that uses this same pattern. Since this is only used in the sorting, I wonder what you think about moving these constants to the data management hooks file or somewhere like that (maybe the data management utils.js file?) until they're needed in multiple places?

client/components/DataManagement/hooks.js
```diff
@@ -0,0 +1,232 @@
...
+                }
+            }
+        }
+        return derivedOverallPhaseByTestPlanId;
+    }, [testPlans, testPlanVersions]);
+
+    return { derivedOverallPhaseByTestPlanId };
+};
+
+export const useTestPlansByPhase = (testPlans, testPlanVersions) => {
```
alflennik, line 88: I'm not 100% sure about this, and I think you should do whatever way you prefer, but here's a rough suggestion, which is to use plain functions instead of hook functions for these three internal helpers. It would interfere with the use of useMemo though. It's very subjective but the reason for that suggestion is that hooks require the reader to mentally track the reactivity of the hooks so it feels more intimidating to read.

alflennik (changes requested):
Damn, nice filtering and sorting!

I went through all the code and I don't have much to comment on, beyond some very minor points below. Here are a couple observations I had from testing it out:

- Minor point: the spacing seems to be greater between the filter buttons than I would expect, I like the spacing between the "Filter" text and the first button, can that spacing be repeated for the following buttons?
- I'm struggling to articulate this, but when I click on the sort buttons it seems like sometimes they go from asc to desc and sometimes they don't. Would it be possible for the first click to activate the sorting for that column without switching from asc to desc and then the second click switch from asc to desc? If that doesn't make sense I can try to articulate it more clearly and possibly record a video of my screen.

client/components/DataManagement/hooks.js
stalgiag: Agreed! I opted for renaming the file for now. A more common solution if hooks become more prevalent in the future would be a hooks folder in the component folder with a separate file for each hook.

client/utils/constants.js
stalgiag: I tried to localize each of them to where they are most heavily used. One in specific can and is used in many places, `TEST_PLAN_VERSION_PHASES` so I kept that in this more global file.

client/components/DataManagement/hooks.js
stalgiag: I can definitely see your point. My (not super strongly held) opinion is that since they are doing operations on potentially large arrays it is worth the trade off in readability to be able to avoid redoing those operations on unrelated re-renders. I am also hoping to one day get to refactor some code in DataManagement to use these internal helpers. They were designed to replace similar non-memo'd code in that component and DataManagementRow at a later date. I am just leaving that out of this PR for the sake of this feature making it before release cut-off.

But point definitely taken on not being inviting to readers. I wonder if this would be alleviated somewhat by each being in a dedicated file which is pretty standard practice. I am going to leave as-is for now but keep thinking on this.

stalgiag:
Thank you for checking it out!

Some inline responses to individual points. Also addressed sort click UX feedback and smaller margins between filter buttons.

alflennik (approved):
Looking great, I like the way the filters are working now, seems super intuitive.

howard-e:
@stalgiag functionally this looks really good, and even though it slightly deviates from the original proposal which we'll still need some further clarity on, this feels pretty intuitive.

> * The [APG sortable table example ](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/) was followed for the creation of the `SortableTableHeader` component. When testing with VoiceOver for macOS, I noticed that there is no dynamic announcement when the`aria-sort` attribute of the currently focused header is changed. I believe this is an issue with VoiceOver as the output is identical on the APG example.

After looking at this, it's super jarring. There's no output on VO and NVDA only announces the sort order the first time the header is clicked for me. I can confirm the same happened with VO in the example, but the announcements were consistent with NVDA there.

What do you think about using a (_probably hidden_) aria-live element instead and updating the content of that when the sort happens?

stalgiag:
> What do you think about using a (_probably hidden_) aria-live element instead and updating the content of that when the sort happens?

@howard-e I have taken this approach. I made a generic AriaLiveRegionProvider that can wrap a set of elements and provide access to a hidden live region that can be updated with a polite announcement. Now changes are announced in a similar way to the example you shared with me. I was only able to test on VoiceOver. It is a little verbose but very clear what is happening.

howard-e (approved):

PR #751, opened 2023-08-22
Checkbox for selecting "No Output" for Output field in Test Run form

stalgiag:
## Overview

This PR adds a new component `OutputTextArea` which renders the label, feedback, textarea, and new "no output" checkbox for the output field in a Test Run form.

## Details

- [x] If the "No output" is checked then the text area field should be disabled 
- [x] We should make it impossible to submit the form if the text area is blank without clicking the checkbox. We need to display an error message. 
- [x] The report should unambiguously describe this input from the tester 
- [x] This shouldn't require changes in the database.

When the checkbox is checked, the textarea is given the value of 'No output was detected.' and the textarea is disabled. Upon unchecking the box, the text area is cleared. Once any value is entered into the text area, the checkbox is disabled. Upon loading, if the form previously had a value of 'No output was detected.', the checkbox is displayed as checked on load.

## Screenshots
A screenshot of a section of Action Menu Button test plan run. The header reads 'After Tab'. A label for the text area reads 'VoiceOver for macOS output after Tab'. Displayed inline to the label on the right side, is a checkbox with a label that says 'No Output'. The checkbox is unchecked and the text area is empty.

A screenshot of the same section as the previous screenshot but the 'No output' checkbox is now checked. The textarea is a slightly lighter grey and it reads 'No output was detected.'

A screenshot of the same section as the previous two screenshots but the textarea now has a value that reads 'Output was detected'. The text is black and the textarea has a focus highlight. The checkbox is unchecked and a very faint gray.

---
  - https://app.asana.com/0/0/1204491043205338


alflennik (changes requested):
Thanks for this! It works well when editing the results, but when I hit submit and came back to the page the checkbox seems like it got desynced from the input box. I don't remember the exact steps let me know if I can help reproduce.

No output unchecked even though text box says no output was detected

client/components/TestRenderer/OutputTextArea/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+const OutputTextAreaWrapper = styled.div`
+    > textarea {
+        width: 100%;
+    }
+`;
+
+const NoOutputCheckbox = styled(Form.Check)`
+    display: inline-block;
+    float: right;
+    color: ${props => (props.disabled ? 'lightgray' : 'inherit')};
```
alflennik, line 17: I felt like the lightgray might be TOO lightgray, I'm curious if you try 4:1 contrast (which is #7F7F7F) if it will look good and also sufficiently disabled? If not I think I'm okay dipping below 4:1 contrast, since the control is pretty irrelevant when there's text in the text box.

stalgiag:
> Thanks for this! It works well when editing the results, but when I hit submit and came back to the page the checkbox seems like it got desynced from the input box. I don't remember the exact steps let me know if I can help reproduce.
> 
> No output unchecked even though text box says no output was detected

I was able to reproduce. Thanks for catching this!

stalgiag:
There was a state issue on mount with the checkbox. Now fixed.

alflennik (approved):
Great! I verified that the feedback I raised has been addressed.

ccanash:
Hi @mcking65  this was included on Dec 5th production release

PR #752, opened 2023-08-22
Adjust semantic structure on Data Management Page

### LINKED ISSUE #718: Content of test plan status summary table cells is not semantically structured so is difficult for screen reader users to understand
This bug is present in the sandbox on July 28.

To reproduce:
1. Open the data management page.
2. Examine any cell that contains multiple data elements, such as in the candidate review column for the alert test plan.

**Actual result:**

The candidate review column content for the alert test pattern is presented to screen readers as a link followed by a string with words that are run together. Users cannot distinguish one data element from another.

>V22.12.08
>0 Open IssuesTarget 295 Days Past

**Expected Result:**

The content of the cell is semantically structured so it is possible to understand it contains three separate data elements. For example, it could be presented as a list:

>* V22.12.08
>* 0 Open Issues
>* Target 295 Days Past

The HTML for this cell is:

~~~
<td>
  <div class="css-i0ihz8"><span class="version-string"><a
        href="/aria-at/0928bcf530efcf4faa677285439701537674e014/build/review/alert.html" target="_blank"
        rel="noreferrer"><svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="circle-check"
          class="svg-inline--fa fa-circle-check " role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"
          color="#2BA51C">
          <path fill="currentColor"
            d="M256 512c141.4 0 256-114.6 256-256S397.4 0 256 0S0 114.6 0 256S114.6 512 256 512zM369 209L241 337c-9.4 9.4-24.6 9.4-33.9 0l-64-64c-9.4-9.4-9.4-24.6 0-33.9s24.6-9.4 33.9 0l47 47L335 175c9.4-9.4 24.6-9.4 33.9 0s9.4 24.6 0 33.9z">
          </path>
        </svg><b>V22.12.08</b></a></span><span class="more"><span class="more-issues-container"><span
          class="issues css-1ghojnj"></span> 0 Open Issues</span><span class="target-days-container">Target <b>295
          Days</b> Past</span></span></div>
</td>
~~~

There are several options for coding this. Typically, a `ul` with 3 `li` elements is preferable. However, if that complicates styling, you could add `role="list"` on the div containing the cell content and then put `role="listitem"` on child spans that represent each item. Note: It is important that list semantics are not used in cells that contain only one data element. So, the list markup must be used conditionally. Here is how the above HTML could be modified.

~~~
<td>
  <div class="css-i0ihz8" role="list">
    <span role="listitem" class="version-string">
      <a
        href="/aria-at/0928bcf530efcf4faa677285439701537674e014/build/review/alert.html"
        target="_blank"
        rel="noreferrer">
          <svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="circle-check"
            class="svg-inline--fa fa-circle-check " role="img" xmlns="http://www.w3.org/2000/svg"
            viewBox="0 0 512 512"
          color="#2BA51C">
            <path fill="currentColor"
              d="M256 512c141.4 0 256-114.6 256-256S397.4 0 256 0S0 114.6 0 256S114.6 512 256 512zM369 209L241 337c-9.4 9.4-24.6 9.4-33.9 0l-64-64c-9.4-9.4-9.4-24.6 0-33.9s24.6-9.4 33.9 0l47 47L335 175c9.4-9.4 24.6-9.4 33.9 0s9.4 24.6 0 33.9z">
          </path>
        </svg>
        <b>V22.12.08</b>
      </a>
    </span>
    <span class="more">
      <span role="listitem" class="more-issues-container">
        <span class="issues css-1ghojnj"> </span> 0 Open Issues
      </span>
      <span role="listitem" class="target-days-container">Target <b>295 Days</b> Past</span>
    </span>
  </div>
</td>
~~~

howard-e:
Resolves #718.

### DIFF
--- a/client/components/DataManagement/DataManagementRow/index.jsx
+++ b/client/components/DataManagement/DataManagementRow/index.jsx
@@ -517,7 +517,7 @@ const DataManagementRow = ({
                 // "active"
                 insertActivePhaseForTestPlan(latestVersion);
                 return (
-                    <PhaseCell role="list">
+                    <PhaseCell role="list" aria-setsize={isAdmin ? 2 : 1}>
                         <VersionString
                             role="listitem"
                             date={latestVersionDate}
@@ -526,7 +526,6 @@ const DataManagementRow = ({
                         />
                         {isAdmin && (
                             <Button
-                                role="listitem"
                                 ref={ref => setFocusRef(ref)}
                                 className="advance-button"
                                 variant="secondary"
@@ -655,7 +654,7 @@ const DataManagementRow = ({
                     // Phase is "active"
                     insertActivePhaseForTestPlan(latestVersion);
                     return (
-                        <PhaseCell role="list">
+                        <PhaseCell role="list" aria-setsize={isAdmin ? 3 : 2}>
                             <VersionString
                                 role="listitem"
                                 date={latestVersionDate}
@@ -665,7 +664,6 @@ const DataManagementRow = ({
                             />
                             {isAdmin && (
                                 <Button
-                                    role="listitem"
                                     ref={ref => setFocusRef(ref)}
                                     className="advance-button"
                                     variant="secondary"
@@ -707,10 +705,11 @@ const DataManagementRow = ({
                                     Advance to Candidate
                                 </Button>
                             )}
-                            <TestPlanReportStatusDialogWithButton
-                                role="listitem"
-                                testPlanVersionId={latestVersion.id}
-                            />
+                            <span role="listitem">
+                                <TestPlanReportStatusDialogWithButton
+                                    testPlanVersionId={latestVersion.id}
+                                />
+                            </span>
                         </PhaseCell>
                     );
                 }
@@ -877,8 +876,7 @@ const DataManagementRow = ({
                     return (
                         <PhaseCell
                             role="list"
-                            aria-level="1"
-                            aria-setsize={shouldShowAdvanceButton ? 4 : 3}
+                            aria-setsize={shouldShowAdvanceButton ? 5 : 4}
                         >
                             <VersionString
                                 role="listitem"
@@ -889,7 +887,6 @@ const DataManagementRow = ({
                             />
                             {shouldShowAdvanceButton && (
                                 <Button
-                                    role="listitem"
                                     ref={ref => setFocusRef(ref)}
                                     className="advance-button"
                                     variant="secondary"
@@ -918,17 +915,17 @@ const DataManagementRow = ({
                                     Advance to Recommended
                                 </Button>
                             )}
-                            <TestPlanReportStatusDialogWithButton
-                                role="listitem"
-                                testPlanVersionId={latestVersion.id}
-                            />
-                            <span role="list" className="more" aria-level="2">
+                            <span role="listitem">
+                                <TestPlanReportStatusDialogWithButton
+                                    testPlanVersionId={latestVersion.id}
+                                />
+                            </span>
+                            <span className="more">
                                 <span
                                     role="listitem"
                                     className="more-issues-container"
                                 >
                                     <ReportStatusDot className="issues" />
-                                    &nbsp;
                                     {issuesCount} Open Issue
                                     {`${issuesCount === 1 ? '' : 's'}`}
                                     {`${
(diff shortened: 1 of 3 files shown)


client/components/common/VersionString/index.js
```diff
@@ -41,11 +41,13 @@ const VersionString = ({
     linkRef,
     linkHref
 }) => {
+    const dateString = convertDateToString(date, 'YY.MM.DD');
+
     const body = (
-        <>
+        <span aria-label={`V${dateString}`}>
```
stalgiag, line 47: In VoiceOver on MacOs, the non-link version of this markup reads with more complexity now. Here is what I experienced:

On update-database-impl:
On first focus, I hear: "Draft Review, V22.03.17, Column 5 of 7"
When I navigate to the next item, "Review Completed" immediately gets the focus.

On  this branch:
On first focus, I hear "V22.03.17, group, 1 of 2"
When I navigate to the next item, "V" is focused and read.
When I navigate to the next item, "22.03.17" is focused and read
When I navigate to the next item, the parent is focused again and "end of V22.03.17, group" is read.
When I navigate to the next item, "Review Completed" gets focus

stalgiag: Note that the link version has identical output on both branches.

client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -918,10 +927,14 @@ const DataManagementRow = ({
...
+                                role="listitem"
                                 testPlanVersionId={latestVersion.id}
                             />
-                            <span className="more">
-                                <span className="more-issues-container">
+                            <span role="list" className="more">
+                                <span
+                                    role="listitem"
+                                    className="more-issues-container"
+                                >
```
stalgiag, line 937: This definitely sounds better and it broken up into fewer parts. The only part that was challenging on VoiceOver was when I first focus, I only hear "List 2 Items, Level 2". These two items are connected to one another visually for the sake of the layout but it doesn't feel like they are conceptually linked. If this is the case, could they be separated out of a list so that they read separately without the additional layer of having the list read? If they are conceptually linked, maybe the list element could have a label that described the combination of listitems?

stalgiag (commented):
Exciting! Thanks for getting this work started. Curious to know which screen reader you are working with. I am hearing different but not necessarily cleaner output on VoiceOver with macOS but that may just be the nature of the reader.

comments inline.

howard-e:
> Curious to know which screen reader you are working with. I am hearing but not necessarily cleaner output on VoiceOver with macOS but that may just be the nature of the reader.

I tested with NVDA and didn't get the chance to follow up with VO at the time. So I'll definitely pull that up tomorrow so I can better respond to your feedback. Thanks much for what you've shared, so I know what to look out for!

client/components/common/VersionString/index.js
alflennik: Okay, so this is fascinating. In JSX, if you put `<b>{'V' + dateString}</b>` instead of `<b>V{dateString}</b>`, the issue where the screen reader reads the V separate from the date goes away.

If you do that the issue necessitating the aria-label goes away.

I had no idea until I started messing around with it.

client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -541,14 +541,16 @@ const DataManagementRow = ({
                 // "active"
                 insertActivePhaseForTestPlan(latestVersion);
                 return (
-                    <PhaseCell>
+                    <PhaseCell role="list">
                         <VersionString
+                            role="listitem"
```
alflennik, line 546: In the VersionString component, you should take the rest props `...props` and then insert them in the first child, or else the VersionString will be missing the listitem role.

alflennik (changes requested):
Having the list is a huge improvement, it seems much more intuitive.

client/components/common/VersionString/index.js
howard-e: That's bizarre and kudos for finding that! We should definitely note this down somewhere. I tested it out and it sounds like expected!

client/components/DataManagement/DataManagementRow/index.jsx
howard-e: Good catch, and I also did the same with `TestPlanReportStatusDialogWithButton`

alflennik (approved):

stalgiag (approved):
Works fantastically on VO now. Thanks for this!

PR #766, opened 2023-08-30
Remove superfluous header from Test Plan Report Status Dialog

### LINKED ISSUE #756: Reporting Status Dialog contains misleading, superfluous heading
The report status dialog has 2 level 2 headings. The first heading is correct. the second heading that appears just before the table contains incorrect content that is misleading. It needs to be removed.

For example, the Report Status dialog for the Color Viewer Slider Test Plan has:

Correct heading:

>Report Status for the Color Viewer Slider Test Plan

Incorrect heading to remove:

>Reports for Draft Alert Test Plan

image

stalgiag:
see title.

This PR removes the superfluous and incorrectly static heading from the TestPlanReportStatusDialog table. 

resolves #756


howard-e (approved):
LGTM!

PR #767, opened 2023-08-30
Update Version History Page

### LINKED ISSUE #761: Test plan version history page is incomplete
On the test plan version page:
1. Version summary tables are missing versions.
2. Tables of 'Timeline for All Versions' are missing both events and versions.
3. GitHub issues are not shown.
4. Version headings should show most recent versions first.
5. Some heading levels are incorrect.
6. Tables are missing captions.
7. Strings in headings and table cells are run together.

As I go through this page, given the length of sections, I wonder if it would be beneficial for all level 2 headings after the version summary to be disclosures that are collapsed by default, e.g.:

`<h2><button aria-expanded="false">GitHub Issues</button></h2>`

Then people could easily see what is on the page without scrolling down.

Note that details/summary can't be used for this because you can't wrap a summary in a heading, and the heading navigation is super important.

Here is more information about each of the above problems using [Radio Group Example Using aria-activedescendant Test Plan Versions | ARIA-AT](https://aria-at-staging.w3.org/data-management/radiogroup-aria-activedescendant) as an illustrative example.

#### Version summary tables are missing versions

The radio version summary table shows data for:
* V23.04.10
* V22.05.18

According to the commit history, the following versions are missing:
* V2023.08.23
* V2022.12.8
* V2022.05.5
* V2022.04.7
* V2022.03.17
* V2021.10.13
* V2021.09.22
* V2021.07.22

#### Tables of 'Timeline for All Versions' are missing both events and versions

The radio timeline shows more versions that the summary, but several are still missing. In addition, many events are missing.

Deprecations are not shown. For example, it shows V21.10.13 R&D Complete, but that version never shows a deprecation event, which most likely occurred on march 17, 2022 when R&D was completed for V2022.03.17 (one of the versions missing from the timeline.

Similarly, there is an event for V22.05.18 Draft Review Started, but there is no event in the table for R&D complete of that version.

There is an event for V23.04.10 Candidate Review Started, but that version doesn't have R&D complete or draft review started.

In the timeline tables that are for one specific version, I failed to mention that if two events occur on the same day, the second sort factor should be phase. It is confusing to see draft review started listed in the table before R&D complete because a plan has to complete R&D before it can go to draft review.

#### GitHub issues are not shown

I couldn't find a plan that lists any GitHub issues. We definitely have issues for several of the plans that should be listed. Is it the case that this feature is still being developed? If so, it would be better to leave out the heading until we have the data pipeline for it. I assumed it was complete because #719 is marked as in the sandbox and we don't have a separate issue tracking this part of #719 nor a reference to one from #719.

#### Version headings should show most recent versions first

My apologies for showing this requirement only in the mockup and not documenting it in the text of #719. As one navigates by heading through the history, like the version summary table, the most recent plan version sections should be at the top.

#### Some heading levels are incorrect.

The information architecture of the page is that a version heading should be level 2 and the timeline information for that version is a subsection. This is currently backward. Currently:

>V23.04.10Candidateon Apr 9, 2022

is a level 3, but it should be level 2.

The heading:

>Timeline for V23.04.10

is currently a level2, but this is a subsection of information for that version so should be level 3.

#### Tables are missing captions

Issue #719 specifies captions. That is primarily so the tables will have accessible names. If visible captions are redundant due to the presence of the headings, then please use aria-labelledby or aria-label to provide the equivalent accessible label.

#### Strings in headings and table cells are run together

This is an issue similar to #718. Example of a heading problem:

>V22.05.18Drafton May 18, 2022

Note that "Draft" and "on" are run together; they need a space between them.

Similarly in the all versions timeline tables:

>V22.05.18Draft Review Started

"V22.05.18" and "Draft" are run together as a single word. They need to be separated with a space, a `br` or semantically separated into list items.

howard-e:
Close #761.

### DIFF
--- a/client/components/TestPlanVersionsPage/index.jsx
+++ b/client/components/TestPlanVersionsPage/index.jsx
@@ -13,6 +13,7 @@ import {
 import VersionString from '../common/VersionString';
 import PhasePill from '../common/PhasePill';
 import { convertDateToString } from '../../utils/formatter';
+import { derivePhaseName } from '../../utils/aria';
 import styled from '@emotion/styled';
 import {
     faArrowUpRightFromSquare,
@@ -165,7 +166,7 @@ const TestPlanVersionsPage = () => {
         }
     };
 
-    const getDerivedDeprecatedDuringPhase = testPlanVersion => {
+    const deriveDeprecatedDuringPhase = testPlanVersion => {
         let derivedPhaseDeprecatedDuring = 'RD';
         if (testPlanVersion.recommendedPhaseReachedAt)
             derivedPhaseDeprecatedDuring = 'RECOMMENDED';
@@ -252,14 +253,14 @@ const TestPlanVersionsPage = () => {
                                         {(() => {
                                             // Gets the derived phase even if deprecated by checking
                                             // the known dates on the testPlanVersion object
-                                            const derivedPhase =
-                                                getDerivedDeprecatedDuringPhase(
+                                            const derivedDeprecatedAtPhase =
+                                                deriveDeprecatedDuringPhase(
                                                     testPlanVersion
                                                 );
 
                                             const phasePill = (
                                                 <PhasePill fullWidth={false}>
-                                                    {derivedPhase}
+                                                    {derivedDeprecatedAtPhase}
                                                 </PhasePill>
                                             );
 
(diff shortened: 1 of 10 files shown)


client/components/TestPlanVersionsPage/index.jsx
```diff
@@ -409,18 +420,19 @@ const TestPlanVersionsPage = () => {
...
                     <div key={testPlanVersion.id}>
-                        <H3>
+                        <H2>
                             <VersionString
                                 date={testPlanVersion.updatedAt}
                                 iconColor={getIconColor(testPlanVersion)}
                                 fullWidth={false}
                                 autoWidth={false}
                             />
+                            &nbsp;
```
stalgiag, line 428: I think this is just a problem between different screenreaders but on VO this outputs:

> heading level 2 5 items V22.09.28  **space** Deprecated  on  Oct 24, 2022

If I remove this `&nbsp;` the "space" is no longer read but I assume this is necessary to prevent the running together issue on other screenreaders (which doesn't happen with VO on the target branch). Just wanted to add a note.

stalgiag (approved):
This is great. Thanks @howard-e for addressing this so quickly!

client/components/TestPlanVersionsPage/index.jsx
howard-e: Very odd. Double checked on NVDA and can confirm it comes out pressed against each other without adding in the `nbsp` like so:

> heading    level 2  V22.10.24R&D on Oct 24, 2022

VO then reads it properly though like that. So pretty puzzling there. Going to spend some time looking for a better approach before merging. Thanks for the review!

howard-e: @stalgiag it seems to be a known issue with VO so I've gone ahead and used an aria-label on the `h2` instead

stalgiag: Exactly the kind of problem this project will hopefully prevent in the future. Thanks for looking into it!

PR #771, opened 2023-09-11
Address edge case scenarios during the updating of TestPlanVersion phase with old TestPlanVersion results

howard-e:
This PR addresses some edge cases found during the update phase process when using old results from an existing TestPlanVersion.

Some likely scenarios being tested:
* What happens when a TestPlanVersion is being updated and there is an older version with results
* What happens when a TestPlanVersion with a non-required report that's unmarked as final is being updated and there is an older version with results
* What happens when a TestPlanVersion with a required report that's unmarked as final is being updated and there is an older version with results


alflennik (approved):
I appreciate this change! And given the complex scenarios at play here, I appreciate the new test suite you added. And the fact you added good sample data that can allow us to easily simulate this deprecation situation is much appreciated as well.

In addition to running the tests successfully I tested the three cases you outlined using the sample data. I also exercised the app a bit using both the sample data and a recent database clone. It all worked for me, which is awesome.

I read through the code. Looks good, not sure I have any changes I'd want to make.

One documentation change which helped me test was to duplicate the sample data that's being used in test environments into the dev environments. I pushed a commit for the database.md file with that change. Let me know if you foresee any issues with that change.

Nice job!

howard-e:
> I appreciate this change! And given the complex scenarios at play here, I appreciate the new test suite you added. And the fact you added good sample data that can allow us to easily simulate this deprecation situation is much appreciated as well.
> 
> In addition to running the tests successfully I tested the three cases you outlined using the sample data. I also exercised the app a bit using both the sample data and a recent database clone. It all worked for me, which is awesome.
> 
> I read through the code. Looks good, not sure I have any changes I'd want to make.

Thanks for checking it out!

> One documentation change which helped me test was to duplicate the sample data that's being used in test environments into the dev environments. I pushed a commit for the database.md file with that change. Let me know if you foresee any issues with that change.


@alflennik yep that change looks good to me, thanks for catching that! Noticed a code block formatting being slightly off, so I've fixed that as well.

PR #772, opened 2023-09-12
Implement protocol for communication with Response Scheduler 

stalgiag:
### Summary

This PR adds the protocol for communication with the Response Scheduler. All related state management for automated jobs is also included. Unit tests and a mock of the Response Scheduler are used to confirm functionality.

### Important notes for review:
- There is no final authentication solution in place. A key is checked into the testing environment variables as a demonstration of one possible straightforward approach.
- The atVersion and browserVersion that the job is running with are not addressed until results are delivered. This is a placeholder while we wait for the finalization of the Response Scheduler. It is more likely that this will be communicated during initialization with possible affordances for requesting specific configurations as a future improvement.

### Misc. Notes:
- CollectionJob is the new table introduced by this PR. It is a relatively thin wrapper around a TestPlanRun that simply stores a reference to the workflow run and the current status of that run.
- In order to make the backend organization clearer for this new paradigm in which database mutations come from the client and also from within the backend, I moved all DB manipulation logic out of resolvers and into services where relevant. This matched the primary pattern in the codebase and changes nothing about the related code functionality but it adds some additional noise to this PR.
- Similarly, I wanted to be able to simplify the resolvers where possible. The `GraphQLContext` was being used to pass instances of `AtLoader` and `BrowserLoader` throughout the codebase. I changed `AtLoader` and `BrowserLoader` to singletons. This way they can be initialized/accessed where used without passing. This adds noise to this PR.


server/controllers/AutomationController.js
```diff
@@ -0,0 +1,329 @@
...
+
+        const automationSchedulerResponse = await axios.post(
+            `${process.env.AUTOMATION_SCHEDULER_URL}/jobs/new`,
+            {
+                testPlanVersionGitSha: gitSha,
+                testIds: tests.map(t => t.id),
+                testPlanName: directory
+            },
+            axiosConfig
+        );
```
jugglinmike, line 70: Please pardon my inexperience with GraphQL, but it seems as though the user authentication logic is implemented solely via Apollo resolvers. Since the HTTP resources introduced in this patch send requests to the automation scheduler before calling into the Apollo framework, I'm wondering if unauthenticated users could trigger privileged actions in the response collection system (e.g. initiating an AT response collection job) even though the subsequent bookkeeping in the ARIA-AT database would fail.

stalgiag: Good catch! I am authenticating the gql queries but only authenticating the routes that the response scheduler will use. Will update.

server/controllers/AutomationController.js
```diff
@@ -0,0 +1,329 @@
...
+    const { testPlanVersion } = testPlanReport;
+
+    // To be considered "Approved", a test plan run must be associated with a test plan report
+    // that is associated with a test plan version that is in "CANDIDATE" or "RECOMMENDED" or
+    // "DRAFT" phase and has been marked as final.
+    if (
+        !testPlanVersion ||
+        testPlanVersion.phase === 'RD' ||
+        (testPlanVersion.phase === 'DRAFT' && !testPlanReport.markedFinalAt)
+    )
```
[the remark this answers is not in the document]
stalgiag: `DEPRECATED` test plan versions shouldn't be selectable for new test plan runs through the UI but I can also make checks for that.

server/controllers/AutomationController.js
stalgiag: On second thought, I don't think I will provide a route for scheduling a new job.

server/graphql-schema.js
```diff
@@ -70,6 +70,36 @@ const graphqlSchema = gql`
         atIds: [ID]!
     }
 
+    """
+    The possible statuses for a CollectionJob.
+    """
+    enum CollectionJobStatus {
```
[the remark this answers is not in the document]
stalgiag: That was my understanding but I do not recall any specific conversations or documentation around 'SKIPPED'.

stalgiag:
I believe that I addressed all of the feedback. Thanks @jugglinmike!

stalgiag:
Awesome! I notice that their mocks have "Stories tagged XXXX" at the top and this PR just has "XXXX" at the top. Other than that this is a great quick fix!

server/controllers/AutomationController.js
```diff
@@ -25,130 +26,157 @@ const axiosConfig = {
     timeout: 1000
 };
 
-const scheduleNewJob = async (req, res) => {
-    try {
```
jugglinmike: I was getting ready to mention that Express 4 doesn't recognize rejected promises, so you can imagine my surprise to learn that [this project relies on a pre-release version of the framework](https://github.com/w3c/aria-at-app/commit/dc4edc33a)!

server/routes/automation.js
```diff
@@ -11,21 +11,32 @@ const {
 const {
     verifyAutomationScheduler
 } = require('../middleware/verifyAutomationScheduler');
+const { handleError } = require('../middleware/handleError');
 
 const router = Router();
 
-router.get('/:jobID/log', getJobLog);
+router.get('/:jobID/log', getJobLog, handleError);
```
jugglinmike, line 14: Something like `router.use(handleError);` will probably communicate our intent more clearly.

server/routes/automation.js
```diff
@@ -17,8 +16,6 @@ const router = Router();
 
 router.get('/:jobID/log', getJobLog, handleError);
 
-router.post('/new', scheduleNewJob, handleError);
```
jugglinmike: This change and the related ones appear to replace the HTTP API we've designed with an HTTP+GraphQL API. Is that right?

stalgiag: Yes. This should've been the approach from the beginning but I read a little too literally into the protocol design document. As far as I can tell, we don't need exposed endpoints for new, cancel, delete, and restart. If we ever want to manage jobs from somewhere other than the ARIA-AT App but still tracked those jobs in the Aria-AT App, then this could be changed.

server/controllers/AutomationController.js
stalgiag: Yeah I checked for Express 5 before I made these changes.

server/tests/integration/automation-scheduler.test.js
```diff
@@ -8,6 +8,7 @@ const { getAtVersions } = require('../../models/services/AtService');
 const { getBrowserVersions } = require('../../models/services/BrowserService');
 const dbCleaner = require('../util/db-cleaner');
 const { default: axios } = require('axios');
+const { COLLECTION_JOB_STATUS } = require('../../util/enums');
```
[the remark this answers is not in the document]
stalgiag: I am a little on the fence. My thinking is that reduced friction in editing and a single source of truth would ensure that any changes to the external systems could be quickly matched and verified on the app. The original motivation for this was that I had a typo with a status of "CANCELED" and it wasn't being caught because semantic satiation + copy/paste was causing me to repeat the mispelling elsewhere. It feels safer to me to have a single source of truth that ensures consistent breakdown in all uses if deviation ever with external system occurs. Not a strongly held opinion, so would be happy to change.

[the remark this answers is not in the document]
jugglinmike: Let's keep it as-is, then!

(Since I get the sense that this kind of thing interests you, too: we agree on the value of a single source of truth, it might just be that my definition is contained to the application logic. My concern taken ad absurdum might look like a test suite that simply verifies that internal functions are being called.)

server/controllers/AutomationController.js
jugglinmike: You also have the test coverage to prove it :) I didn't mean to suggestion otherwise--I was the one caught off-guard.

server/routes/automation.js
jugglinmike: For what it's worth, I don't think you read "too literally;" I think we just spec'd the wrong thing :P I didn't even recognize that until your response just now, so thanks!

jugglinmike (approved):
Thanks, @stalgiag, this looks good to me! @howard-e this is targeting your branch, so I'd like to leave merging to you.

server/tests/integration/automation-scheduler.test.js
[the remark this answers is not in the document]
stalgiag: Ah! Yes, now I better understand your comment and am in complete agreement with you. I took your comment to be on the existence of the enum largely but now I have re-read your first sentence closely and see the word test. I agree with using literal values in the tests. That is useful friction. Changed it 😆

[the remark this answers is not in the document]
jugglinmike: Oh, cool! Thanks for continuing to consider this so thoughtfully

howard-e:
@stalgiag I tried to be very careful with this merge because of how many files it changed. And because there are commits being oddly duplicated which should already exist on `main` which cause the merge to be a bit more involved than it needed to be, by conflicting with a change already existing in `main`.

So what I did was to look at the compare of when this branch was first created vs the latest commit and [found 67 changed files](https://github.com/w3c/aria-at-app/compare/e31ecab...c6221a7). I went through each _actually conflicting_ file between this branch and main and made sure to all the correct lines were being captured.

I’ve only HAD to make 2 additional changes here with the changes coming from `main`. I’ve also stepped through all the functionality and pages of the app and haven’t noticed anything **sinister jumping out** at me. I also verified that the tests, both existing and the new ones added here are running as expected, locally. But I’d love if you could re-verify again when you get the chance that all still looks fine to you.

Hopefully this won’t have to happen again!

howard-e:
> * In order to make the backend organization clearer for this new paradigm in which database mutations come from the client and also from within the backend, I moved all DB manipulation logic out of resolvers and into services where relevant. This matched the primary pattern in the codebase and changes nothing about the related code functionality but it adds some additional noise to this PR.
> * Similarly, I wanted to be able to simplify the resolvers where possible. The GraphQLContext was being used to pass instances of AtLoader and BrowserLoader throughout the codebase. I changed AtLoader and BrowserLoader to singletons. This way they can be initialized/accessed where used without passing. This adds noise to this PR.

Also really loving the changes made for this

stalgiag:
> @stalgiag I tried to be very careful with this merge because of how many files it changed. And because there are commits being oddly duplicated which should already exist on `main` which cause the merge to be a bit more involved than it needed to be, by conflicting with a change already existing in `main`.
> 
> So what I did was to look at the compare of when this branch was first created vs the latest commit and [found 67 changed files](https://github.com/w3c/aria-at-app/compare/e31ecab...c6221a7). I went through each _actually conflicting_ file between this branch and main and made sure to all the correct lines were being captured.
> 
> I’ve only HAD to make 2 additional changes here with the changes coming from `main`. I’ve also stepped through all the functionality and pages of the app and haven’t noticed anything **sinister jumping out** at me. I also verified that the tests, both existing and the new ones added here are running as expected, locally. But I’d love if you could re-verify again when you get the chance that all still looks fine to you.
> 
> Hopefully this won’t have to happen again!

@howard-e thank you so much for doing that! I am very impressed that you were able to sort through the tangle. This all looks good to me. And, yes, here is hoping that we have an easier branching and merging strategy in the future.

stalgiag:
I made a new current work branch named `automation-mvp` and modified this PR to target that branch as discussed in Slack.

stalgiag:
Merging this into `automation-mvp` after approved review by Mike. This work will receive additional review when `automation-mvp` is ready for review.

PR #773, opened 2023-09-14
Update `Timeline for All Versions` table on Test Plan Versions Page

howard-e:
Address:
* https://github.com/w3c/aria-at-app/issues/761#issuecomment-1719015358
> The timeline for all versions should show all events from all versions. The timeline for each individual version should show all events for that specific version. Thus, the number of rows in timeline for all versions should be the sum of all rows from the other timeline tables.
* https://github.com/w3c/aria-at-app/issues/761#issuecomment-1719021693
> It would make the tables more screen reader friendly if we used th instead of td in the version column of the version summary and the date column of the timeline tables.

### DIFF
--- a/client/components/TestPlanVersionsPage/index.jsx
+++ b/client/components/TestPlanVersionsPage/index.jsx
@@ -304,7 +304,7 @@ const TestPlanVersionsPage = () => {
                         <tbody>
                             {testPlanVersions.map(testPlanVersion => (
                                 <tr key={testPlanVersion.id}>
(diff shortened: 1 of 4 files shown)


alflennik (approved):
Code looks good, it makes sense to just make sure the timeline occurs in the proper order, it's a clean solution. I appreciate that most of the complexity is contained to the migrations. I tested it out and saw the fix in action. All the test plan version pages seem to be working properly.

Issue #768, opened 2023-08-30
Update `w3c/aria-at` Feedback Issues


howard-e:
#753 includes functionality to support a standard `Raise an issue` operation throughout the app and is expected to land soon. With this change, the app will expect to show issues from [w3c/aria-at/issues](https://github.com/w3c/aria-at/issues) once they have included a specific formatted string in the body and included applicable labels.

The following body content (preferably hidden) and labels is expected in the following cases:
* Issue raised against a specific test in a Test Plan:
  * Body includes: `<!-- Feedback: "{TEST_NAME}" ({TEST_PLAN_NAME}, Test {NUMBER}, {VERSION_STRING}) -->`
  * Expected labels: `app`,` feedback`, `nvda | jaws | vo` (matching to the AT being filed against), `candidate-review` (if this was raised from the Candidate Review Page)
  * Eg. _`<!-- Feedback: "Navigate backwards to a menu button" (Action Menu Button Example Using element.focus(), Test 6, V21.10.13) -->` in body, with the labels: `app`, `feedback` and `vo`_

* Issue raised as general feedback from the Candidate Review Page:
  * Body includes: `<!-- {AT_NAME} General Feedback: {TEST_PLAN_NAME} {VERSION_STRING} -->`
  * Expected labels: `app`,` feedback`, `nvda | jaws | vo` (matching to the AT being filed against), `candidate-review`
  * Eg. _`<!-- JAWS General Feedback: Action Menu Button Example Using aria-activedescendant V22.03.17 -->` in body, with the labels: `app`, `feedback`, `candidate-review` and `jaws`_

* Issue raised with changes requested from the Candidate Review Page:
  * Body format: `<!-- {AT_NAME} Changes Requested: "{TEST_NAME}" ({TEST_PLAN_NAME}, Test {NUMBER}, {VERSION_STRING}) -->`
  * Expected labels: `app`,` changes-requested`, `nvda | jaws | vo` (matching to the AT being filed against), `candidate-review`
  * Eg. _`<!-- VoiceOver for macOS Changes Requested: "Read information about a pressed toggle button" (Toggle Button, Test 18, V22.05.18) -->` in body, with the labels: `app`, `changes-requested`, `candidate-review` and `vo`_

[...]

PR #775, opened 2023-09-18
Add hidden JSON metadata to issues created by the app

alflennik:
When users raise issues through the app we insert a hidden comment into the issue which is used to later track the issue within the app. This PR improves the system by converting the hidden comment from a string, which was difficult to parse, to JSON, allowing us to remove most of the string parsing logic from the issue system. It also fixes a problem caused by the fact that most users who raise issues do not have the ability to attach the required issue labels - the system now relies on the hidden metadata and missing labels should no longer cause issues. Related to https://github.com/w3c/aria-at-app/issues/768.

### DIFF
--- a/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
@@ -302,12 +302,14 @@ const CandidateTestPlanRun = () => {
     // Assumes that the issues are across the entire AT/Browser combination
     const changesRequestedIssues = testPlanReport.issues?.filter(
         issue =>
+            issue.isCandidateReview === true &&
             issue.feedbackType === 'CHANGES_REQUESTED' &&
             issue.testNumberFilteredByAt === currentTest.seq
     );
 
     const feedbackIssues = testPlanReport.issues?.filter(
         issue =>
+            issue.isCandidateReview === true &&
             issue.feedbackType === 'FEEDBACK' &&
             issue.testNumberFilteredByAt === currentTest.seq
     );
@@ -316,6 +318,7 @@ const CandidateTestPlanRun = () => {
         isCandidateReview: true,
         isCandidateReviewChangesRequested: true,
         testPlanTitle: testPlanVersion.title,
+        testPlanDirectory: testPlanVersion.testPlan.directory,
         versionString,
         testTitle: currentTest.title,
         testRowNumber: currentTest.rowNumber,
@@ -428,7 +431,9 @@ const CandidateTestPlanRun = () => {
     );
 
     const feedback = testPlanReport.issues.filter(
-        issue => issue.testNumberFilteredByAt == currentTest.seq
+        issue =>
+            issue.isCandidateReview &&
+            issue.testNumberFilteredByAt == currentTest.seq
     ).length > 0 && (
         <div className="issues-container">
             <h2>
--- a/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
@@ -302,14 +302,14 @@ const CandidateTestPlanRun = () => {
     // Assumes that the issues are across the entire AT/Browser combination
     const changesRequestedIssues = testPlanReport.issues?.filter(
         issue =>
-            issue.isCandidateReview === true &&
+            issue.isCandidateReview &&
             issue.feedbackType === 'CHANGES_REQUESTED' &&
             issue.testNumberFilteredByAt === currentTest.seq
     );
 
     const feedbackIssues = testPlanReport.issues?.filter(
         issue =>
-            issue.isCandidateReview === true &&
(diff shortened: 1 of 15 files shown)


client/components/CandidateReview/CandidateTestPlanRun/index.jsx
```diff
@@ -302,12 +302,14 @@ const CandidateTestPlanRun = () => {
     // Assumes that the issues are across the entire AT/Browser combination
     const changesRequestedIssues = testPlanReport.issues?.filter(
         issue =>
+            issue.isCandidateReview === true &&
```
howard-e, line 305:
```suggestion
            issue.isCandidateReview &&
```

nit

client/components/CandidateReview/CandidateTestPlanRun/index.jsx
```diff
@@ -302,12 +302,14 @@ const CandidateTestPlanRun = () => {
...
             issue.feedbackType === 'CHANGES_REQUESTED' &&
             issue.testNumberFilteredByAt === currentTest.seq
     );
 
     const feedbackIssues = testPlanReport.issues?.filter(
         issue =>
+            issue.isCandidateReview === true &&
```
howard-e, line 312:
```suggestion
            issue.isCandidateReview &&
```
nit

server/services/GithubService.js
```diff
@@ -29,28 +29,30 @@ const permissionScopes = [
...
+const getAllIssues = async () => {
     let currentResults = [];
     let page = 1;
 
     // eslint-disable-next-line no-constant-condition
     while (true) {
-        const issuesEndpoint =
-            `${GITHUB_ISSUES_API_URL}/issues` +
-            `?labels=app&state=all&per_page=100`;
+        const issuesEndpoint = `${GITHUB_ISSUES_API_URL}/issues?state=all&per_page=100`;
```
howard-e, line 39: Was this the only place where the `app` label was being used? If that's the case, you could stop it being added during `createIssueLink`, to have less 'noise' in the labels. I suppose it could be pretty useful to quickly note what was created by the app, but I also have no strong feelings for either.

howard-e (approved):
@alflennik looks good! I checked through to make sure everything worked as it did before and this definitely makes it a bit easier to update how the issues are queried by the App in the future!

alflennik:
@howard-e would you be willing to take another look? The big change since you last saw it is that I added an issues field to the test plan type to get a list of all issues ever raised for a test plan (including issues for reports that no longer exist).

server/graphql-schema.js
```diff
@@ -230,11 +230,15 @@ const graphqlSchema = gql`
...
-
         """
         Gets all historic versions of the test plan.
         """
         testPlanVersions: [TestPlanVersion]!
+        """
+        A list of all issues which have filed through "Raise an Issue" buttons
+        in the app. Note that results will be cached for at least ten seconds.
+        """
+        issues: [Issue]!
```
howard-e, line 241: Having this being at the `TestPlan` level makes a lot of sense to me, thanks for that change!

server/resolvers/TestPlanReport/issuesResolver.js
```diff
@@ -45,6 +61,11 @@ const issuesResolver = async testPlanReport => {
                     ? 'CHANGES_REQUESTED'
                     : 'FEEDBACK';
 
+            const at = ats.find(at => at.name === hiddenIssueMetadata.atName);
+            const browser = browsers.find(
+                browser => browser.name === hiddenIssueMetadata.browserName
+            );
```
howard-e, line 67: Let's make these name checks case insensitive just in case the GitHub issue is ever edited and someone accidentally does `NVDa` instead of `NVDA`, for example.

server/resolvers/TestPlanReport/issuesResolver.js
```diff
@@ -54,9 +75,12 @@ const issuesResolver = async testPlanReport => {
                 isCandidateReview: hiddenIssueMetadata.isCandidateReview,
                 feedbackType,
                 isOpen: state === 'open',
-                testNumberFilteredByAt: hiddenIssueMetadata.testRowNumber
+                testNumberFilteredByAt: hiddenIssueMetadata.testRowNumber,
+                at,
+                browser
```
howard-e, line 80: Similar to my concern above, if a GitHub issue's AT or Browser name is edited and it doesn't match against any known `At` or `Browser`, this query would fail because those 2 fields are required. Can we guard against this here? Maybe we just don't want to have them be required and default to `N/A`, in those columns on the frontend and share that this is a hint that the test structure in the issue may be malformed to issue authors? Or perhaps we'd want to just not return this item?

howard-e (changes requested):

alflennik:
@howard-e all the feedback should be addressed now, thanks for the review!

howard-e (approved):
@alflennik looks good to me! Thanks for addressing the feedback

PR #776, opened 2023-09-19
Promote unit test isolation

jugglinmike:
The unit testing infrastructure for the project's server logic is currently built around global mutable state. Prior to this commit, some tests did not explicitly clear this state, and as a result, a failure or timeout in one test would cause failures in subsequent tests.

Insert previously-omitted invocations of the state-management function `dbCleaner` as per convention in this test suite.


howard-e (approved):
Good catch! This should resolve some irregular instances of failed tests when the GH Actions are ran for these files.

PR #780, opened 2023-09-19
Correct deprecation date

evmiguel:
As discussed with @howard-e, this PR changes the migration to compare against the most recent version in the database and deprecate against that version.


howard-e (approved):
@evmiguel thanks, LGTM!

Issue #778, opened 2023-09-19
Test plan version history shows incorrect text when a plan in R&D Complete phase is deprecated


mcking65:
If a test plan is deprecated before starting draft review, the text in the "Latest Phase" column of the "Version Summary" table is:

>Deprecated during R&D review

![image](https://github.com/w3c/aria-at-app/assets/7241781/da3c1957-586e-4718-a00f-a759a4a3e4f1)

There is not a phase named "R&D Review".

However, following the spec I wrote in #719 would have not been much better. The spec says to use the text "Deprecated during PHASE_NAME", which would have created the text:

>Deprecated during R&D

That wouldn't have been accurate and would have been confusing. The spec I wrote in in #719 is fine if the phase is draft review or candidate review, but it is not very good for R&D or Recommended.

I recommend using the following text strings for when plans are deprecated in either of these two phases.

If a test plan is deprecated after R&D is complete and before draft review is started, use the string "Deprecated before draft review started".

If a test plan is deprecated after it has been approved as recommended, use the string "Deprecated after being approved as recommended".

mcking65:
Thank you @alflennik  and @howard-e! This is working well in staging now.

ccanash:
Released to prod

PR #781, opened 2023-09-20
Deprecated phase copy

alflennik:
Addresses https://github.com/w3c/aria-at-app/issues/778 by changing the way phases are referred to when a deprecation occurs.

### DIFF
--- a/client/components/TestPlanVersionsPage/index.jsx
+++ b/client/components/TestPlanVersionsPage/index.jsx
@@ -288,9 +288,9 @@ const TestPlanVersionsPage = () => {
                                                     return (
                                                         <>
                                                             {deprecatedPill}
-                                                            before&nbsp;
+                                                            {` before `}
                                                             {draftPill}
-                                                            review
+                                                            {` review `}
                                                         </>
                                                     );
                                                 }
@@ -302,8 +302,7 @@ const TestPlanVersionsPage = () => {
                                                     return (
                                                         <>
                                                             {deprecatedPill}
-                                                            after being approved
-                                                            as&nbsp;
+                                                            {` after being approved as `}
                                                             {phasePill}
                                                         </>
                                                     );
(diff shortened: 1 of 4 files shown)


client/components/TestPlanVersionsPage/index.jsx
```diff
@@ -273,10 +273,48 @@ const TestPlanVersionsPage = () => {
...
+                                                        DRAFT
+                                                    </PhasePill>
+                                                );
+
+                                                if (
+                                                    derivedDeprecatedAtPhase ===
+                                                    'RD'
+                                                ) {
+                                                    return (
+                                                        <>
```
howard-e, line 289: With NVDA, the text comes back as `Latest Phase  column 2  Deprecatedbefore Draftreview`. The same doesn't happen with VO though, and I haven't checked JAWS.

client/components/TestPlanVersionsPage/index.jsx
```diff
@@ -273,10 +273,48 @@ const TestPlanVersionsPage = () => {
...
+                                                    );
+                                                }
+
+                                                if (
+                                                    derivedDeprecatedAtPhase ===
+                                                    'RECOMMENDED'
+                                                ) {
+                                                    return (
+                                                        <>
+                                                            {deprecatedPill}
```
howard-e, line 304: Like above, NVDA returns the speech as `Latest Phase  column 2  Deprecatedafter being approved as Recommended`. The same doesn't happen with VO

client/components/TestPlanVersionsPage/index.jsx
```diff
@@ -273,10 +273,48 @@ const TestPlanVersionsPage = () => {
...
+                                                            {phasePill}
+                                                        </>
+                                                    );
+                                                }
+
                                                 return (
                                                     <>
-                                                        {deprecatedPill} during{' '}
-                                                        {phasePill} review
+                                                        {deprecatedPill}
```
howard-e, line 314: With one that has been deprecated in Candidate, NVDA announces it as `Latest Phase  column 2  Deprecatedduring Candidatereview`. The same doesn't happen with VO for me.

howard-e (commented):
@alflennik the copy and changes look good to me!

But I did notice this causes the same issue with NVDA where the text is being announced "joined together". I described what I'm experiencing below. I haven't experienced the same on VO though.

howard-e (approved):
LGTM! Thanks for addressing the feedback

Issue #763, opened 2023-08-30
Clashing version string titles


howard-e:
The design doesn't speak on how we could differentiate between between clashing version strings for the same test plan (and will likely be covered by https://github.com/w3c/aria-at-app/issues/571). This will happen if multiple updates are made to the same test plan on the same day.

For example [Banner Landmark](https://aria-at-staging.w3.org/data-management/banner) has 3 pushes on [January 4, 2022](https://github.com/w3c/aria-at/commits/master/tests/banner).

Would something like `VYY.MM.DD_X`, where `X` is the most recent instance of the version date found, help support this? (Eg. V22.01.04_1, V22.01.04_2, etc)

howard-e:
@mcking65 said in #761:
> In the timeline tables that are for one specific version, I failed to mention that if two events occur on the same day, the second sort factor should be phase. It is confusing to see draft review started listed in the table before R&D complete because a plan has to complete R&D before it can go to draft review.

mcking65:
@howard-e , originally I didn't think this would matter; we would just have the latest merge of the day be the one that is the version for that day. However, that would be a problem in the timelines if instance 1 was advanced to draft and then on the same day, an error was discovered and so instance 2 was created and merged on the same day. In that case, the timeline would need to show instance 1 being deprecated by instance 2.

So, I like your _x proposal. however, we don't want _1 on every version; it is just more verbosity and could get annoying. So, I propose that we start with _2. Such version numbers will be uncommon, but as you pointed out, they do happen.

So, _1 is assumed without being present in the version string. If a second version is created same day, it will be _2, and so on.

alflennik:
Hi @mcking65! I wanted to highlight a small superficial change that I mentioned in the PR unless you had any feedback with it. From PR https://github.com/w3c/aria-at-app/pull/804.

> When @mcking65 proposed a solution, he suggested using a underscore. For whatever reason, I found a hyphen looks noticeably better. It bugs me because I can't really articulate why. It just does. So I went with a hyphen. I'd be willing to go back to an underscore if that's a problem.

Let me know if you have any feedback about that! Thanks.

ccanash:
Hi @mcking65  this has been released to production so I am closing it. If you would like an update to the hyphing.

ccanash:
Hi, closing this as it was released yesterday. @mcking65  if you would like to change the hyphen we can create a separate issue.

PR #782, opened 2023-09-20
Address clashing version string titles

evmiguel:
This PR addresses issue #763, clashing version string titles. To implements the proposed ordering where a duplicate but more recent version will have the format of `YYYY.MM.DD_X`, where X is the more recent version found.

### DIFF
--- a/client/components/TestPlanVersionsPage/index.jsx
+++ b/client/components/TestPlanVersionsPage/index.jsx
@@ -200,6 +200,32 @@ const TestPlanVersionsPage = () => {
         item => item.link
     );
 
+    // Hash the test plan version version strings. If found, record the order
+    const testPlanVersionDuplicates = new Map();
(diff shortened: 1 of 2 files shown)


alflennik:
Nice job implementing this! I can confirm that it works on the Test Plan Versions page which is awesome. Since there are a lot of places the version strings are used, and actually more every day (for example I just added version strings to the issues raised by the app), I want to make sure that the solution works across the board, so I wanted to suggest adding a new versionString field to the TestPlanVersion type, originating on the server. This could use logic similar to what you already implemented. Then on the frontend the versionString could be queried directly. Any thoughts on that or does that make sense?

alflennik:
Closing because this is addressed in https://github.com/w3c/aria-at-app/pull/804 thanks for getting this one going Erika!

Issue #786, opened 2023-09-22
Labels on sort direction for overall status are inverted and sort has undesirable side-effect on test plan name column in Test Plan Status Summary Table


mcking65:
From #648, referenced by #654:
>Default Table Sort
>* First sort by overall status descending: recommended plans, then candidate plans, then draft plans, then R&D complete plans. 
>* Then sort by test plan name ascending. 
>* Then sort by Covered AT ascending 

Note that the requirements define:
* **DESCENDING ORDER** as: recommended , candidate, draft, R&D complete. 
* **ASCENDING ORDER** as: R&D complete, draft, candidate, recommended.

There are two problems.

#### Problem 1: sort direction label

The table has the correct default sort order of overall status descending and test plan name ascending. However, the column header on overall status incorrectly represents the sort order as ascending. At least that is how it is communicated via aria-sort, and I assume the sort icon is matching aria-sort.

#### Problem 2: side effect on test plan sort.

The specification in #648 has a gap. It didn't state what should happen to the test plan column when changing the sort order of the overall status. It should have stated that test plan should stay sorted alphabetically ascending when changing the direction of sort for overall status. Currently, toggling the sort on overall status is also toggling the sort on test plan. It should not do that.

This should also be true for covered AT. However, we have the problem that we only have one set of covered AT for now.

stalgiag:
Thank you for this. Both issues should be addressed by #790 .

ccanash:
Hi @mcking65  flagging this was included in the release from yesterday.

mcking65:
I have verified the fix in prod.

PR #790, opened 2023-09-25
Fix inverted data management status sort order, always sort ascending by name on columns other than name

stalgiag:
This PR changes the following for the Data Management table:
- Changes the interpretation of phase order in phase-based sorting
- Inverts the starting sort sequence to ascending
- Ensures that rows are always sorted in ascending order alphabetically as a second sort on columns other than name. Descending alphabetical sort is only used when explicitly set on the Test Plan name column.

Resolves issue #786


howard-e (approved):
@stalgiag looks good to me!

PR #794, opened 2023-09-26
Add collapsible areas for Test Plan Versions History page

### LINKED ISSUE #769: Add collapsible areas to Test Plan Versions History Page
@mcking65 said in #761:

> As I go through this page, given the length of sections, I wonder if it would be beneficial for all level 2 headings after the version summary to be disclosures that are collapsed by default, e.g.:
> 
> `<h2><button aria-expanded="false">GitHub Issues</button></h2>`
> 
> Then people could easily see what is on the page without scrolling down.
> 
> Note that details/summary can't be used for this because you can't wrap a summary in a heading, and the heading navigation is super important.

Paul-Clue:
see #769

This PR turns all level 2 headings after the version summary into disclosures that are collapsed by default on the `TestPlanVersionsPage`. This was done to curtail the need for excessive scrolling on the page.


client/components/common/VersionString/index.js
```diff
@@ -31,6 +31,11 @@ const StyledPill = styled.span`
         width: auto;
         margin: 0.75rem;
     }
+
+    // .check {
+    //     margin: 0;
+    //     color: #969696;
+    // }
```
howard-e, line 38:
```suggestion
```

client/components/TestPlanVersionsPage/index.jsx
```diff
@@ -497,28 +600,13 @@ const TestPlanVersionsPage = () => {
...
-                        report => report.isFinal
-                    ).length;
-
-                return (
-                    <div key={testPlanVersion.id}>
-                        <H2
+            <PageSpacer />
+            <DisclosureComponent
+                componentId="versionHistory"
+                title={testPlanVersions.map(testPlanVersion => {
```
howard-e, line 606: The console throws the following errors:

```
Warning: Failed prop type: Invalid prop `title` supplied to `DisclosureComponent`, expected one of type [string, string].
```

and 

```
Warning: Encountered two children with the same key, `disclosure-[object Object]-key`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

And that's because the DisclosureComponent has been accepting plain string as title values in the array and the constructed key is also based on that string but here, the title array is being passed as `[ <React.Element>, ... ]` so you may want to:

1. Relax the constraints on the title props being accepted by the DisclosureComponent, and
2. Allow for explicitly key values to be used, to be set

howard-e (commented):
Looks good and definitely makes navigating this page feel less noisy. Once you can get in those changes mentioned here, this should soon be ready to go!

client/components/common/VersionString/index.js
```diff
@@ -91,12 +90,13 @@ const VersionString = ({
 };
 
 VersionString.propTypes = {
+    date: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)])
+        .isRequired,
```
stalgiag, line 94: @Paul-Clue this is the source of the failing test. Your PR seems to be causing a regression from #804 which was merged while you were working on this. [Here is the relevant changed file.](https://github.com/w3c/aria-at-app/pull/804/files#diff-5a6ef0a58a10458be0d847550d91f47e200c5f0922c30226f9e7d927d69194b5). It looks like that PR was merged since you opened this PR. I don't think you have any actual changes in this file.

Paul-Clue: Got it. Thanks!

Paul-Clue: Hey, @stalgiag, I just fixed issues that were causing the error. Check it out, when you have time.

client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -717,21 +717,15 @@ const DataManagementRow = ({
...
+                        getVersionData(otherTestPlanVersions);
 
                     const completionDate =
                         otherLatestVersion.recommendedPhaseReachedAt;
 
                     return (
                         <PhaseCell role="list">
-                            <VersionString
-                                role="listitem"
-                                date={otherLatestVersionDate}
```
stalgiag, line 732: It looks like this `date` was accidentally left behind during the changes in #804. Is that right?

client/components/TestPlanVersionsPage/index.jsx
```diff
@@ -65,13 +145,24 @@ const ThemeTableHeader = styled(UnstyledThemeTableHeader)`
...
     const { testPlanDirectory } = useParams();
 
     const { loading, data, error } = useQuery(TEST_PLAN_VERSIONS_PAGE_QUERY, {
         variables: { testPlanDirectory },
         fetchPolicy: 'cache-and-network'
     });
+    const forceUpdate = useForceUpdate();
+
+    const expandedVersionSections = useRef();
+    const toggleVersionSections = useRef();
```
stalgiag, line 165: I wonder if this could be simplified by having these two values be a single `useState`? You could have `[expandedVersionSectionId, setExpandedVersionSectionId]`. This way you would only need to check if the `expandedVersionSectionId` matches the `testPlanVersion` when looping over for rendering. The `onClick` could run `setExpandedVersionSectionId(testPlanVersion.id)`. The nice thing about using `useState` is that it would automatically trigger a re-render and you wouldn't need the `useForceUpdate` at all.

client/components/TestPlanVersionsPage/index.jsx
```diff
@@ -534,125 +629,182 @@ const TestPlanVersionsPage = () => {
...
+                            {getEventDate(testPlanVersion)}
+                        </span>
+                    );
+                })}
+                disclosureContainerView={testPlanVersions.map(
+                    testPlanVersion => {
+                        const vString = `V${convertDateToString(
+                            testPlanVersion.updatedAt,
+                            'YY.MM.DD'
+                        )}`;
```
stalgiag, line 642: Is this needed after #804? Test plan versions have a `versionString` property now.

stalgiag (changes requested):
Glad the unit tests are passing. I added a few small inline questions and suggestions.

client/components/TestPlanVersionsPage/index.jsx
alflennik: Thanks for the note! Yeah, this is something Paul and I were hashing out together. We started with useState, but found that it was producing stale data, in effect, we would set one of the sections to open, but it wouldn't open, and when logging the values, we would see things set to "false" even right after trying to set them to "true".

The advantage of useRef is that its data can't ever be stale, but the disadvantage is exactly as you noted, you have to introduce a useForceUpdate function to trigger a rerender.

If you have an alternative idea, that would be welcome. We went back and forth a few times trying to figure out the most elegant way to do this.

client/components/DataManagement/DataManagementRow/index.jsx
Paul-Clue: Hey @stalgiag this is fixed.

client/components/TestPlanVersionsPage/index.jsx
Paul-Clue: Got it!

client/components/TestPlanVersionsPage/index.jsx
stalgiag: > Thanks for the note! Yeah, this is something Paul and I were hashing out together. We started with useState, but found that it was producing stale data, in effect, we would set one of the sections to open, but it wouldn't open, and when logging the values, we would see things set to "false" even right after trying to set them to "true".

Weird! That sounds like a bug with how the state was being updated. The state will only update after the rerender but it will also trigger a render of the component and its children so any immediate logs will log a stale value but after the render finishes the value will change and a rerender will happen. `useState` theoretically should be perfectly designed for this precise problem.

Anyway, maybe there is something specific about this situation that I'm not seeing with my shallow window in. This could easily be a side effect of the strange array-based interface for `DisclosureComponent`. I would personally like to put this in a tech debt inventory but we don't need to address it now.

stalgiag (approved):
Looks good to me. Thanks for the hard work!

alflennik (approved):

Issue #732, opened 2023-08-03
Mockups for AT Automation MVP


isaacdurazo:
### Visual Mockup
![ARIA-AT - Test Queue](https://github.com/w3c/aria-at-app/assets/1379244/e894924f-c6b9-44ac-84c2-0b957ff72fa5)

### Text-based Mockup
From this page, a Test Admin is able to start a Response Collection Job. The Response Collector Bot should first be assigned to a Test Plan the same way we currently assign human testers. Once the Bot appears under the Testers column, the action column will display the button "Start Response Collection", which when triggered will start the Collection Job. Once the Job gets started, this button will turn into a "Pause"  button.

This Design also suggests the display of different statuses under the Report status column, such as X Canceled, X Errors, X Skipped, X Complete, and X Queued. Not necessarily in that order

### Visual Mockup
![ARIA-AT Test Run Page](https://github.com/w3c/aria-at-app/assets/1379244/e6b4bd91-ff69-4421-b0a0-cb5029f102dd)

### Text-based Mockup
This design re-uses the existing structure and layout of the Test Run Page. On the left-hand side, we repurposed the timeline to display different statuses related to the response collection job. The statuses and their icons are:
1. Queued - a blue circle with a three horizontal line icon where the one in the middle is an arrow pointing to the right.
2. Skipped - a black circle with a semi-circle arrow suggesting a "jump"
3. Canceled - a purple circle with a  white crossed-out circle
4. Complete - a green circle with a white double-check mark
5. Error - a red circle with a triangle with an exclamation mark inside
6. Running - this state doesn't use an icon, but rather an outer green circle around whatever status that test is. This is what we currently use to indicate the test in this test plan is being looked at.

On top of the main Test Area, we are repurposing the toast that currently reads "Reviewing tests of USERNAME" to indicate that one is reviewing the Responses of the Collection Bot. This design suggests: "Reviewing tests of Response Collector Bot"

The main Test area is read-only and it will display the Bot responses collected during the job in the text area currently dedicated to the AT output

The assertions section is where the tester will provide the verdicts but it will be until the tester gets assigned that they are able to modify these.

Lastly, on the right-hand side, we have repurposed the Test Options buttons to have the following actions: 
1. Raise an issue
2. Restart - Available only to admins and for jobs with the "canceled" or "error" status.
3. Cancel - Available only to admins and for jobs with the "queued" or "running" status.
4. View Log - Available for jobs with "skipped" status

[...]

PR #802, opened 2023-10-03
UI for Automation MVP

stalgiag:
see #732.

### Overview

This PR adds a user interface for running and managing automated test plan runs. Apologies for the size of this PR, I understand that this makes reviews challenging. The scope is a byproduct the feature's impact and newness.

### Checklist

A checklist of changes, based on @jugglinmike 's wonderful breakdown of the design:

<details>
  <summary>Tasks handled by this PR</summary>

### Test plan report status dialog

- Button labeled "Add to Test Queue"
  - [x] Location: the cells in the column labeled "Report Status"
  - [x] Present when: always present
  - [x] Enabled when: always enabled
  - Behavior:
    - [x] if the row's Test Plans can be run in automation (only NVDA at the time of writing), opens a dialog titled "Adding TEST_PLAN_NAME Test Plan to the Test Queue"
    - [x] otherwise, opens the existing dialog for adding a manual test run
- Dialog with heading, "Adding TEST_PLAN_NAME Test Plan to the Test Queue"
  - [x] Modal: yes
  - Contents
    - [x] text: "Choose how the report for AT_NAME and BROWSER_NAME will be generated. Add it to the queue so it can be assigned to a tester at a later time or start running automated response collection with BOT_NAME."
    - Button labeled "Cancel"
      - [x] Present when: always present
      - [x] Enabled when: always enabled
      - [x] Behavior: closes the dialog
    - Button labeled "Add and run later"
      - [x] Present when: always present
      - [x] Enabled when: always enabled
      - [x] Behavior: adds the Test Plan to the Test Queue with zero Testers assigned
    - Button labeled "Add and run with bot"
      - [x] Present when: always present
      - [x] Enabled when: always enabled
      - [x] Behavior: adds the Test Plan to the Test Queue with a bot assigned and starts a collection job

### Test Queue

- Button labeled: "Manage [AT_Name] Bot run"
  - [x] Location: the cells in the column labeled "Report Status"
  - [x] Present when: the row's Test Plan includes a bot-assigned Test Plan Run
  - [x] Enabled when: always enabled
  - [x] Behavior: opens a dialog titled "Manage [AT_Name] Bot run"
- Prose describing the status of the collection
  - [x] Text: "Responses for x of y tests recorded. Verdicts for x of y assertions assigned."
  - [x] Location: the column labeled "Testers" (below the entry representing the bot)
  - [x] Present when: the row's Test Plan includes an "unfinished" bot-assigned Test Plan Run
- Dialog with heading "Manage [AT_Name] Bot run"
  - [x] Modal: yes
  - Contents
    - Select-only combobox labeled "Assign to..."
      - [x] Options: the account names of other testers who are not already assigned to another run of the same test plan with the same AT and browser
      - [x] Enabled when: always enabled
      - [x] Behavior: re-assigns the Test Plan Run
    - Button labeled "Stop running"
      - [x] Present when: always present
      - [x] Enabled when: at least one job is "queued"
      - [x] Behavior: updates the status of all "queued" tests to "cancelled"
    - Button labeled "Retry cancelled collection jobs"
      - [x] Present when: always present
      - [x] Enabled when: all collection jobs are "complete" or "cancelled"
      - [x] Behavior: changes all "cancelled" collection jobs to "queued" and initiates automated collection for just those tests
    - Button labeled "Mark as finished"
      - [x] Present when: always present
      - [x] Enabled when: all collection jobs are "complete" and all assertions have a verdict assigned
      - [x] Behavior: opens a dialog titled "Are you sure you wish to mark [AT_Name] Bot run as finished?"
- Dialog with heading: "Are you sure you wish to mark [AT_Name] Bot run as finished?"
  - [x] Modal: yes
  - Contents
    - Button labeled "Yes"
      - [x] Present when: always present
      - [x] Enabled when: always enabled
      - [x] Behavior: sets the Test Plan Run's testResults' `completedAt` value to the current time (this is how the system tracks the "finished" state), closes the current dialog, and closes the dialog titled, "Manage [AT_Name] Bot run"
    - Button labeled "No"
      - [x] Present when: always present
      - [x] Enabled when: always enabled
      - [x] Behavior: closes the current dialog

### Test Plan Run page

- Button labeled "Raise an issue"
  - [x] Present when: the Test Plan Run is assigned to a bot
  - [x] Enabled when: always enabled
  - [x] Behavior: opens a new window on the GitHub.com "file an issue" page
- Button labeled "View Log"
  - [x] Present when: the Test Plan Run is assigned to a bot
  - [x] Enabled when: always enabled
  - [x] Behavior: opens a new window with the text that has been output by the response collection process up to the moment that the window was opened (the text is static)
- Prose with invitation to report problems (no change from the prose which is currently deployed)
  - [x] Text: "Need Help? E-mail Us"
  - [x] Link: `mailto:` URL
  - [x] Present when: always present

</details>

### Tech Debt

- More frontend unit tests are needed
- Unit tests for retrying cancelled jobs
- Let DB define the CollectionJob id. This protocol design originally had the Response Collector defining the ID based on the GH workflow ID. Since this is no longer the case, the CollectionJob table should have a UUID type. Rewriting tests that would allow this to be set on the table rather than by the server would've been time-consuming so leaving for debt.
- Move all CollectionJob graphql mutations into CollectionJobOperations. At first, I avoided adding an umbrella "XOperations" into the schema but it became necessary. Since it exists, it would be better to move most (if not all) mutations under it.
- Status exclusively exists at the CollectionJob level. This was intended as minimally intrusive. Much complexity could be refactored if statuses exist at the test result level as well since this a 'status' is often derived and frequently the relevance of CollectionJob status is only in how we derive status at the test level. I believe this will be looked into during v2 revisions.
- Proptypes.object was used in places. It would be better to create reusable proptype fragments for complex objects.
- When writing unit tests for updating jobs with a new (as of yet unused) at or browser version, the `atVersion` and `browserVersion` for the `testPlanRun.testResults` is returning `null` in the space of the unit test transaction. I was not able to pinpoint why this is and I was losing too many cycles to it. The functionality works in all real world scenarios that I could think to test so I am leaving as debt so that someone with fresh perspective (perhaps me on another day) can look at it.


client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -31,17 +53,86 @@ function AddTestToQueueWithConfirmation({
...
 
+    const closeWithUpdate = async () => {
+        setShowConfirmation(false);
+        await triggerUpdate();
+        setTimeout(() => {
+            if (buttonRef?.current) {
+                buttonRef.current.focus();
+            }
+        }, 0);
+    };
```
jugglinmike, line 64: It seems like it'd be safer to defer resolution until the timeout fires.

client/tests/BasicModal.test.jsx
```diff
@@ -0,0 +1,128 @@
...
+                content="Test content"
+                actions={[
+                    { label: 'Save', onClick: saveFunction },
+                    { label: 'Delete', onClick: deleteFunction }
+                ]}
+            />
+        );
+        fireEvent.click(screen.getByText('Save'));
+        fireEvent.click(screen.getByText('Delete'));
+        expect(saveFunction).toHaveBeenCalledTimes(1);
```
jugglinmike, line 54: Let's swap these two lines (or better yet, duplicate the assertion) so the test verifies the correct label-onClick association.

client/components/TestQueueRow/index.jsx
```diff
@@ -495,23 +491,7 @@ const TestQueueRow = ({
                     </div>
                     {isSignedIn && isTester && (
                         <div className="secondary-actions">
-                            {isAdmin &&
-                                !isLoading &&
-                                !testPlanReport.conflictsLength && (
```
jugglinmike, line 500: This part of the condition (concerning the existence of conflicts) isn't present in the refactored rendering code. Is that intentional?

client/utils/automation.js
```diff
@@ -0,0 +1,21 @@
...
+export const isAssertionValidated = assertion => {
+    return (
+        assertion.passed ||
+        (assertion.failedReason &&
+            assertion.failedReason !== 'AUTOMATED_OUTPUT')
+    );
+};
+
+// TODO: Stub, support for more bot users should be added
+export const getBotUserFromAtBrowser = (at, browser) => {
```
jugglinmike, line 15: Since the object returned by this function doesn't fully conform to the "user" type, and since the only call site just needs the name, I think it'd be more clear to define `getBotUsernameFromAtBrowser`, instead.

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -31,17 +53,86 @@ function AddTestToQueueWithConfirmation({
...
                 variables: {
                     testPlanVersionId: testPlanVersion.id,
                     atId: at.id,
                     browserId: browser.id
                 }
             });
+            const testPlanReport =
+                res?.data?.findOrCreateTestPlanReport?.populatedData
+                    ?.testPlanReport ?? null;
+            tpr = testPlanReport;
```
jugglinmike, line 121: Maybe another piece of tech debt: update `useTriggerLoad`'s callback-accepting function to forward the callback's return value. (Also test `useTriggerLoad` while we're at it.)

client/components/BotRunTestStatusList/index.js
```diff
@@ -0,0 +1,139 @@
...
+
+    const [getCollectionJobStatus, { data: collectionJobStatusQueryResult }] =
+        useLazyQuery(COLLECTION_JOB_STATUS_BY_TEST_PLAN_RUN_ID_QUERY, {
+            fetchPolicy: 'cache-and-network'
+        });
+
+    const [collectedData, setCollectedData] = useState([]);
+    const requestedTestRunIds = useRef(new Set());
+
+    const filteredTestPlanRuns = useMemo(() => {
```
jugglinmike, line 40: A name like `botTestPlanRuns` might be more helpful since it documents what the data describes rather than how it was derived.

client/components/BotRunTestStatusList/index.js
```diff
@@ -0,0 +1,139 @@
...
+        if (!testPlanRunsQueryResult?.testPlanRuns) {
+            return [];
+        }
+        return testPlanRunsQueryResult.testPlanRuns.filter(testPlanRun =>
+            isBot(testPlanRun.tester)
+        );
+    }, [testPlanRunsQueryResult?.testPlanRuns]);
+
+    useEffect(() => {
+        if (filteredTestPlanRuns && filteredTestPlanRuns.length > 0) {
```
jugglinmike, line 50: We can shrink the checks in this component by taking for granted that `filteredTestPlanRuns` always stores an array value.

client/components/BotRunTestStatusList/index.js
```diff
@@ -0,0 +1,139 @@
...
+        startPolling,
+        stopPolling
+    } = useQuery(TEST_PLAN_RUNS_TEST_RESULTS_QUERY, {
+        variables: { testPlanReportId },
+        fetchPolicy: 'cache-and-network',
+        pollInterval: 2000
+    });
+
+    const [getCollectionJobStatus, { data: collectionJobStatusQueryResult }] =
+        useLazyQuery(COLLECTION_JOB_STATUS_BY_TEST_PLAN_RUN_ID_QUERY, {
```
jugglinmike, line 33: Beyond the context of this patch, I'd like to pick your brain about this pattern.

client/components/BotRunTestStatusList/index.js
```diff
@@ -0,0 +1,139 @@
...
+        !collectedData ||
+        !(collectedData.length === filteredTestPlanRuns.length)
+    ) {
+        return null;
+    }
+    return (
+        <BotRunTestStatusUnorderedList className="text-secondary fs-6">
+            <li className="m-2">
+                <ReportStatusDot className="tests-complete" />
+                {numTestsCompleted} Tests Completed
```
jugglinmike, line 120: Watch out for "1 Tests Completed"

client/components/ManageBotRunDialog/MarkBotRunFinishedButton/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+        data: testPlanRunCompletionQuery,
+        loading,
+        refetch
+    } = useQuery(TEST_PLAN_RUN_RESULT_IDS, {
+        variables: {
+            testPlanRunId: testPlanRun.id
+        }
+    });
+
+    const { triggerLoad } = useTriggerLoad();
```
jugglinmike, line 25: Here and elsewhere, the `useTriggerLoad` hook doesn't seem useful without a reference to the `loadingMessage` state it provides. It seems like we could forgo it in favor of immediately-invoked function expressions, but I wonder whether we actually want to employ the `LoadingStatus` component for these new asynchronous operations.

client/components/ManageBotRunDialog/index.jsx
```diff
@@ -0,0 +1,168 @@
...
+                    onClick: () => setShowDeleteDialog(true)
+                }
+            }
+        ];
+    }, [testPlanReportId, testPlanRun, possibleReassignees, onChange]);
+    const deleteConfirmationContent = (
+        <>
+            <p>
+                The test plan run for <b>{testPlanRun.tester.username}</b> will
+                be deleted:
```
jugglinmike, line 116:
```suggestion
                be deleted.
```

jugglinmike (commented):
Nice work, @stalgiag! I still have a lot left to review, but here's what I got to today.

client/components/AddTestToQueueWithConfirmation/index.jsx
stalgiag: I agree with this. This code exists in the app in a different location prior to this PR so this might not necessarily be automation tech debt but general app tech debt. Not sure if this distinction is meaningful.

client/components/BotRunTestStatusList/index.js
stalgiag: Good catch! Thanks :-)

client/components/BotRunTestStatusList/index.js
stalgiag: Sounds good! This was an initial pass on "real-time" updates before I found out this would be a focus in v2. I do think the pattern should be improved and possibly refactored. I didn't spend additional cycles on it due to the aforementioned v2 focus.

client/components/BotRunTestStatusList/index.js
stalgiag: Good callout!

client/components/BotRunTestStatusList/index.js
stalgiag: Nice!

client/components/ManageBotRunDialog/MarkBotRunFinishedButton/index.jsx
stalgiag: Good point. I am going to update to use the existing pattern correctly with a LoadingStatus component. I think this pattern would benefit from refactoring app-wide. The loading status could be a singleton in the app so that one doesn't have to remember to include it in children components. The hook could update its state via context.

client/tests/BasicModal.test.jsx
stalgiag: Wow! Color me impressed that you detected that.

client/components/TestQueueRow/index.jsx
stalgiag: Wow thanks so much for catching this!

client/hooks/useTestPlanRunValidatedAssertionCounts.js
```diff
@@ -0,0 +1,99 @@
...
+            const {
+                testPlanRun: { testResults }
+            } = testPlanRunAssertionsQueryResult;
+            for (let i = 0; i < testResults.length; i++) {
+                const scenarios = testResults[i].scenarioResults;
+                for (let j = 0; j < scenarios.length; j++) {
+                    const assertions = scenarios[j].assertionResults;
+                    for (let k = 0; k < assertions.length; k++) {
+                        count++;
+                    }
```
jugglinmike, line 49:
```suggestion
                    count += assertions.length;
```

client/hooks/useTestPlanRunValidatedAssertionCounts.js
```diff
@@ -0,0 +1,99 @@
...
+    const testResultsLength = useMemo(() => {
+        if (testPlanRunAssertionsQueryResult) {
+            const length =
+                testPlanRunAssertionsQueryResult.testPlanRun?.testResults
+                    ?.length;
+            return length;
+        } else {
+            return 0;
+        }
+    }, [testPlanRunAssertionsQueryResult, stopPolling, startPolling]);
```
jugglinmike, line 32: This code returns `undefined` if an optional property is undefined. If the properties are always present, then we should remove the optional chaining. Otherwise, it seems like we'd want the zero fallback:

```suggestion
    const testResultsLength = useMemo(() => {
        return testPlanRunAssertionsQueryResult?.testPlanRun?.testResults?.length || 0;
    }, [testPlanRunAssertionsQueryResult, stopPolling, startPolling]);
```

Separately, it doesn't seem like `stopPolling` and `startPolling` are dependencies for this calculation. Is it safe to remove them?

client/components/TestQueueCompletionStatusListItem/PreviouslyAutomatedTestCompletionStatus/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+        const { testPlanRun } = testPlanRunAssertionsQueryResult;
+
+        if (!testPlanRun) {
+            return 0;
+        }
+
+        const { testResults } = testPlanRun;
+
+        return testResults?.reduce((acc, testResult) => {
```
jugglinmike, line 35: This component generally handles unavailable data by rendering a zero, except for this final condition where it would render the string "undefined". For consistency, how about:

```suggestion
        const { testResults = [] } = testPlanRun;

        return testResults.reduce((acc, testResult) => {
```

client/components/common/AtAndBrowserDetailsModal/index.jsx
```diff
@@ -215,6 +215,12 @@ const AtAndBrowserDetailsModal = ({
...
                 }
+                actions={[
+                    {
+                        label: 'Ok',
+                        onClick: () => navigate('/test-queue')
+                    }
+                ]}
                 actionLabel="Ok"
                 closeLabel="Cancel"
                 handleAction={() => navigate('/test-queue')}
```
jugglinmike, line 226: We can remove `actionLabel` and `handleAction` here, too.

client/components/common/BasicModal/index.jsx
```diff
@@ -36,6 +36,27 @@ const BasicModal = ({
         headerRef.current.focus();
     }, [show]);
 
+    const renderAction = (action, index) => {
+        if (action.component) {
+            return React.createElement(
+                action.component,
+                { key: `CustomComponent_${index}`, ...action.props },
+                null
+            );
```
jugglinmike, line 45: Fancy!

client/tests/AssignTesterDropdown.test.jsx
```diff
@@ -0,0 +1,193 @@
...
+            </MockedProvider>
+        );
+
+        const button = await screen.getByRole('button', {
+            name: /assign testers/i
+        });
+        fireEvent.click(button);
+
+        const items = await screen.findAllByText(/john/);
+        fireEvent.click(items[0]);
```
jugglinmike, line 156: Asserting the value of `items.length` here will make diagnosing failure easier when no element is present (and also prevent false positives if a future change causes multiple items to be rendered).

```suggestion
        const items = await screen.findAllByText(/john/);
        expect(items.length).toBe(1);
        fireEvent.click(items[0]);
```

client/tests/AssignTesterDropdown.test.jsx
```diff
@@ -0,0 +1,193 @@
...
+
+        await waitFor(() => {
+            expect(useMutation).toHaveBeenCalledWith(ASSIGN_TESTER_MUTATION);
+            expect(mockProps.onChange).toHaveBeenCalledTimes(1);
+        });
```
jugglinmike, line 161: The alert message seems worth testing.

client/tests/AssignTesterDropdown.test.jsx
```diff
@@ -0,0 +1,193 @@
...
+    it('renders without crashing', () => {
+        render(
+            <MockedProvider mocks={mocks}>
+                <AssignTesterDropdown {...mockProps} />
+            </MockedProvider>
+        );
+        expect(screen.getByLabelText('Assign testers')).toBeInTheDocument();
+    });
+
+    it('assigns tester correctly and calls assignTester mutation', async () => {
```
jugglinmike, line 143: How about a test for "assigning" a bot user (since that involves a different mutation)?

client/tests/BotRunTestStatusList.test.jsx
```diff
@@ -0,0 +1,179 @@
...
+                    id: testRun.id
+                }
+            }
+        }
+    }));
+
+    return [testPlanRunMock, ...collectionJobStatusMocks];
+};
+
+test('correctly displays statuses for single COMPlETED test run', async () => {
```
jugglinmike, line 41:
```suggestion
test('correctly displays statuses for single COMPLETED test run', async () => {
```

server/models/services/AtService.js
```diff
@@ -553,6 +553,40 @@ const getUniqueAtVersionsForReport = async testPlanReportId => {
...
+        { atId, name },
+        atVersionAttributes,
+        atAttributes,
+        options
+    );
+
+    if (!version) {
+        /* TODO: releasedAt manually entered by users submitting new version, support actual dates in automation */
+        version = await createAtVersion(
+            { atId, name, releasedAt: releasedAt ?? new Date(Date.now()) },
```
jugglinmike, line 580:
```suggestion
            { atId, name, releasedAt: releasedAt ?? new Date() },
```

server/models/services/CollectionJobService.js
```diff
@@ -407,5 +480,6 @@ module.exports = {
     // Custom for Response Scheduler
     scheduleCollectionJob,
     restartCollectionJob,
-    cancelCollectionJob
+    cancelCollectionJob,
+    retryCancelledTests: retryCancelledCollections
```
jugglinmike, line 484: Just out of curiosity, why the rename?

jugglinmike (commented):
Hey @stalgiag! I've completed the review. This includes some in-line comments about low-level details along with some notes about where my functional testing didn't match my expectations. I'm happy to elaborate on any of this in whatever channel works best for you.

---

The `<textarea>` elements in the Test Plan Run page should have the attribute `readonly` for bot-assigned Test Plan Runs. (The intent is to highlight the fact that this data was provided by the system and to promote intentionality when overriding the bot's data--Test Admins will have to explicitly re-assign the Test Plan Run to allow this.)

The input elements for verdicts are spec'd to be `disabled`, but I think Matt's changes to the "manage bot run" dialog have obsoleted that design. (Specifically, Matt expects to be able to "Mark as finished" when managing the bot run. If we disallowed assigning assertions, then this would only be possible for collections that had 100% parity with historic results. I think this still fits with the underlying design principle of "ownership" because verdict assignment is never performed by bots.)

##### Viewing tests in a bot-assigned Test Plan Run alters their state

Steps to reproduce:

1. Add the "Action Menu Button Example Using aria-activedescendant" Test Plan to the Test Queue for NVDA and Chrome via the button labeled "Add and run with bot"
2. Assign the "NVDA Bot" user to the new Test Plan
3. Immediately activate the "Manage NVDA Bot Run" button and select "Stop Running" before the simulated results collection process completes
4. Use the "Open run as..." button to open the Test Plan Run as the "NVDA Bot" user
5. View one of the tests labeled as "Cancelled by Bot"

Expected behavior: the status of the selected test remains "Cancelled by Bot"

Actual behavior: the status of the selected test changes to "Completed by Bot"

##### Simulated AT response collection is rendered inconsistently

Steps to reproduce:

1. Add the "Action Menu Button Example Using aria-activedescendant" Test Plan to the Test Queue for NVDA and Chrome via the button labeled "Add and run with bot"
2. Wait for the simulated results collection to complete for all 15 tests
3. Use the "Open run as..." button to open the Test Plan Run as the "NVDA Bot" user

Expected behavior: all 15 tests should be in the "completed by bot" state, and they should all have simulated data in the `<textarea>` elements designated for AT response data

Actual behavior: only the first 10 tests are presented as expected; the remaining 5 are indefinitely reported as "in progress by bot" and include no AT response data

Additional notes

- This also occurs when assigning the bot via the "Add Testers" menu
- The "Stop Running" button in the "Manage [AT Name] Bot Run" dialog is disabled in this state, corroborating the progress information in the Test Queue
- Reassigning the Test Plan Run to a human from this state causes the Test Plan Run's status to be updated to "5 of 15 tests evaluated", though opening the Test Plan Run displays the first 10 tests as "in progress" and the final 5 tests as "not started"
- The number of collections which are truly reported as "completed by bot" appears to be deterministic but dependent on the total number of tests in the Test Plan.)

##### "Assign Testers" dropdown includes bot user in ineligible Test Plans

1. Add the "Action Menu Button Example Using aria-activedescendant" Test Plan to the Test Queue for JAWS and Chrome
2. Navigate to the "Test Queue" page
3. Open the "Assign testers" menu

Expected behavior: only human humans are listed in the menu

Actual behavior: human users and the bot user are listed in the menu

Additional notes

- Selecting the bot in this situation causes the menu to close with no further updates to the user interface. The browser developer console shows a HTTP 500 response to a POST request

client/hooks/useTestPlanRunValidatedAssertionCounts.js
stalgiag: 🤣 very good point

client/components/TestQueueCompletionStatusListItem/PreviouslyAutomatedTestCompletionStatus/index.js
stalgiag: Nice and clean. I sometimes forget that destructured properties can have default values.

server/models/services/AtService.js
stalgiag: Oh nice! Didn't know that the constructor would fallback to Date.now()

client/tests/AssignTesterDropdown.test.jsx
stalgiag: Added!

stalgiag:
Wow thanks so much for the review! This was very generous in attention to detail and organized bug reporting.

I believe I have addressed all of the feedback. Each inline comment was addressed and resolved separately. I was also able to reproduce all of the described bugs and fixed each.

The only part that was a bit unclear was this:

> The input elements for verdicts are spec'd to be `disabled`, but I think Matt's changes to the "manage bot run" dialog have obsoleted that design. (Specifically, Matt expects to be able to "Mark as finished" when managing the bot run. If we disallowed assigning assertions, then this would only be possible for collections that had 100% parity with historic results. I think this still fits with the underlying design principle of "ownership" because verdict assignment is never performed by bots.)

My reading of that leads me to believe that no update is needed here and that we want to be able to modify verdicts for bot runs that have no historic results without reassigning. Let me know if I am misreading.

Some notes:
- While working on this, I noticed an unrelated bug in which `testPlanRun.browserVersion` is occasionally null on first render in the test run. This seems like an issue with the order of operations with the AtAndBrowserDetailsModal. For now, I am allowing null values in the graphql schema. This is another item of tech debt, though it is general to the entire app not just automation.
-  If you are curious, the issue with not all tests completing was due to using `TestPlanVersion.tests`for mapping csv row to test id instead of `TestPlanReport.runnableTests`. As a coincidence some of the tests did map correctly (and in certain test plan versions all did) so that is why there were some completions.

jugglinmike (approved):
> Wow thanks so much for the review! This was very generous in attention to detail and organized bug reporting.

It's the least I could do given all the effort that went in to this patch.

> I believe I have addressed all of the feedback. Each inline comment was addressed and resolved separately. I was also able to reproduce all of the described bugs and fixed each.

Seems that way to me!

> The only part that was a bit unclear was this:
>
> > The input elements for verdicts are spec'd to be `disabled`, but I think Matt's changes to the "manage bot run" dialog have obsoleted that design. (Specifically, Matt expects to be able to "Mark as finished" when managing the bot run. If we disallowed assigning assertions, then this would only be possible for collections that had 100% parity with historic results. I think this still fits with the underlying design principle of "ownership" because verdict assignment is never performed by bots.)
>
> My reading of that leads me to believe that no update is needed here and that we want to be able to modify verdicts for bot runs that have no historic results without reassigning. Let me know if I am misreading.

That's right (I could have been more clear about that; I just wanted to acknowledge that the design spec was not entirely relevant any longer.)

> Some notes:
>
>   * While working on this, I noticed an unrelated bug in which `testPlanRun.browserVersion` is occasionally null on first render in the test run. This seems like an issue with the order of operations with the AtAndBrowserDetailsModal. For now, I am allowing null values in the graphql schema. This is another item of tech debt, though it is general to the entire app not just automation.

Good catch! Could you add that to the list of items in the description of this pull request? That'll make it easier for us to create new tasks once this lands (cc @ccanash).

>   * If you are curious, the issue with not all tests completing was due to using `TestPlanVersion.tests`for mapping csv row to test id instead of `TestPlanReport.runnableTests`. As a coincidence some of the tests did map correctly (and in certain test plan versions all did) so that is why there were some completions.

"Passing Spuriously: A Memoir by Mike Pennisi"

stalgiag:
Thank you @jugglinmike 

I am going to merge this. Note that this will be merged into `automation-mvp` which is our "dev" branch for the automation work. A secondary round of reviews may happen before `automation-mvp` makes it into `main` or is released to production.

@ccanash I have a list of tech debt items in the top comment of this PR. I ranked the items in relative priority but none of them take precedent over getting the MVP into main. These can be added to Asana/GH Projects at some point but no rush. I can also flag again when we are ready to merge into main if that would be useful.

Issue #785, opened 2023-09-22
Dialog for changing candidate review completion target date has several critical problems


mcking65:
The dialog that opens with a button for changing the target date in the candidate review column of the test plan status summary table on the data management page has the following problems:
1. It allows invalid dates: Earliest allowed date should be 180 days after the day that candidate review started. Note that this may still be in the past and might also be earlier than the R&D complete date of the version currently in candidate review.
2. When the date is saved, one day is subtracted. Example, set the date to Sep 30, 2023 and then save. Re-open the dialog , and the date in the field will be Sep 29, 2023.
3. Accessibility issue: The date input should receive focus when the dialog opens.
4. Accessibility issue: The label of the input is not associated with the input.
5. Accessibility issue: The dialog is not labeled. It has aria-labelledby, but the idref is invalid. It should reference the heading.
6. Accessibility issue: Like several of the other dialogs, the tab sequence is broken. Tabbing from the save button should move focus to the date input. Back tabbing from the date input should move focus to the save button.

howard-e:
> 6. Accessibility issue: Like several of the other dialogs, the tab sequence is broken. Tabbing from the save button should move focus to the date input. Back tabbing from the date input should move focus to the save button.

@mcking65 I think this is the same as what was reported in #757. The team found that this is a side-effect of the underlying bootstrap component and are tracking that work to submit against that.

The other items will be assigned

mcking65:
@howard-e wrote:

> > 6. Accessibility issue: Like several of the other dialogs, the tab sequence is broken. Tabbing from the save button should move focus to the date input. Back tabbing from the date input should move focus to the save button.
> 
> @mcking65 I think this is the same as what was reported in #757. The team found that this is a side-effect of the underlying bootstrap component and are tracking that work to submit against that.
> 
> The other items will be assigned


Thanks. Yes, it is the same tabbing issue.

Please create a separate issue in this repo to track/link to  the work on the bootstrap component.

howard-e:
> Please create a separate issue in this repo to track/link to the work on the bootstrap component.

https://github.com/w3c/aria-at-app/issues/805 covers that

mcking65:
@howard-e, @ccanash 

In staging, I can still reproduce 5 of the 6 bugs I reported in this issue. Plus, there is one additional side-effect that I either didn't previously notice or is new. I don't want to create a new issue; I wonder if this should be moved back to in-progress?

Here are the details of my current observations.

[...]

PR #806, opened 2023-10-04
Fix for focus trapping in modal/dialog components

### LINKED ISSUE #757: Keyboard and focus related accessibility issues in report status dialog
The report status dialog is very close to the mark on accessibility, but there are a few issues that are rather consequential:

1. Focus is initially placed on the dialog.
2. You can't tab backward through the dialog.
3. Tab is not fully trapped; it leaves the dialog and goes into browser chrome, e.g., address bar.

Please remove tabindex=-1 from the dialog element. Set tabindex=-1 on the heading and initially place focus there when the dialog opens. If you don't like how that looks from a design perspective, It would be OK to visually draw focus around the entire dialog instead of just the heading.

Make sure that it is possible to both Ttab backwards and forwards through the dialog tab sequence, and ensure that focus does not leave the dialog. For an example, see [Modal Dialog Example | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/).

### LINKED ISSUE #805: React bootstrap modal does not prevent focus from entering browser interface
This issue was surfaced by #785 and #757. The current version of the React Bootstrap modal component does not prevent focus from entering the browser interface. Our dialog components extend the React Bootstrap modal component so our dialog components also have this behavior. This is expected behavior on React Bootstrap's side but it does not match the APG dialog example.

Browsers (tested on Chrome) prevent modification of focus during an event that would lead to focusing on browser interface elements. The to-do here is to follow the APG example and insert focusable divs before and after the ~~dialog~~ dialog content*. We can then use these barrier divs to detect focus events and wrap within the boundaries of the dialog.

*Edited for clarity. Thanks @jscholes !

stalgiag:
### Overview

This PR adds a `FocusTrapper` component that traps focus in children regardless of whether those children are rendered in a portal. This component side steps browser restrictions on modifying focus during events that would lead to browser interface elements becoming focused. It does this by adding focusable divs before and after the child element. Focus events that would lead to focusing on the barrier divs are used to wrap the focus around to the correct focusable element within the child.

This component is added directly into the `BasicModal` component that extends the React Bootstrap modal component.

Unit tests are added for `FocusTrapper`.

### Problems addressed

- From #757 
  - [x] Focus is initially placed on the dialog.
  - [x] You can't tab backward through the dialog.
  - [x] Tab is not fully trapped; it leaves the dialog and goes into browser chrome, e.g., address bar.
- From #785 
  - [x] Accessibility issue: Like several of the other dialogs, the tab sequence is broken. Tabbing from the save button should move focus to the date input. Back tabbing from the date input should move focus to the save button. 

### Testing

Since this modifies a common component, the potential for side effects are substantial with this PR. I have done a pass of testing for all of the instances of `BasicModal` use listed below. In all instances focus wrapped to interactive elements as expected. I'd greatly appreciate any additional testing during review.

- [x] TestPlanReportStatusDialog
- [x] 'Are you sure you want to exit?' Test Plan Run
- [x] Candidate Review Completion Date Edit Dialog
- [x] UpdateTargetDateModal
- [x] Advancing Test Plan Dialog
- [x] UpdateVersionModal
- [x]  "Start Over?" Dialog

closes #805
closes #757 
partially addresses #785

### DIFF
--- a/client/components/TestPlanReportStatusDialog/TestPlanReportStatusDialog.css
+++ b/client/components/TestPlanReportStatusDialog/TestPlanReportStatusDialog.css
@@ -1,0 +1,4 @@
+.test-plan-report-status-dialog {
+    max-width: 1200px;
+    min-width: max-content;
+}
--- a/client/components/TestPlanReportStatusDialog/index.jsx
+++ b/client/components/TestPlanReportStatusDialog/index.jsx
@@ -1,6 +1,5 @@
 import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
-import { Modal } from 'react-bootstrap';
 import styled from '@emotion/styled';
 import { getRequiredReports } from './isRequired';
 import AddTestToQueueWithConfirmation from '../AddTestToQueueWithConfirmation';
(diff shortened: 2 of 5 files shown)


alflennik (approved):
Thanks for implementing these fixes, I verified they are working. I tested different access methods like mouse, keyboard, VoiceOver and NVDA. As always, it's great to see some new tests appearing 🙏 Code looks good! I'm glad this fix will apply to all modals going forward.

Issue #787, opened 2023-09-23
Mark as final button should not be shown in the test queue report status column unless at least one run of a plan is complete


mcking65:
#755 suppresses display of the "mark as Final" button when there are conflicts, but the button still appears in other circumstances where a report can't be finalized. In those other circumstances, an error dialog is shown when the button is activated. This creates odd inconsistencies, such as the following, where one row has the button and another does not, but it is not useful or needed in either row.

![image](https://github.com/w3c/aria-at-app/assets/7241781/4dd63c9a-4554-4276-aaa3-78996a7010b0)

Change the display logic so that "Mark as Final" is shown only if all the following are true:
* The user is an admin
* There is at least one complete run
* There are no incomplete runs
* There are 0 conflicts

An incomplete run is a run that is assigned and does not have a complete set of results.

If a report has one complete run and no incomplete runs, the button could appear. Assigning another person to run the test plan for the same report would cause the button to disappear.

If a report has some incomplete runs, deleting all the incomplete runs would cause the button to appear as long as there is at least one complete run and no conflicts.

I don't think we want to force two complete runs even though that is what we normally require. The admin should be able to decide if one run is sufficient, which we may allow in the event of automated re-runs where 0 differences from prior comparables were present.

mcking65:
@jscholes  would like to make sure you concur with all the above requirements.

jscholes:
@mcking65 Yes, these look good to me.

alflennik:
PR #807 will make the button appear and disappear at the correct times, but only with the current behavior of the system, it does not implement a requirement that all tests be filled out. We should keep this issue open until that larger feature is implemented.

PR #807, opened 2023-10-05
Fix Issue 787 Mark as Final Button Before Any Results

alflennik:
Closes https://github.com/w3c/aria-at-app/issues/787. While it is not currently possible to mark results as final before there are any filled out test results (an error message is shown), we still are showing a mark as final button. This PR removes the button in that case.


stalgiag (approved):
Approving as partially addressing the issue.

Issue #798, opened 2023-09-27
System incorrectly permits advancement of draft test plan to candidate review before required reports are complete


mcking65:
To reproduce:
1. As an admin, go to data management page and advance a test plan version from R&D complete to draft review.
2. Open the report status dialog and add required report runs to the test queue.
3. Go to the test queue, assign yourself to the test plan run for one report, complete the report, and finalize it.
4. Go to data management and note that the draft review column shows the required reports are in progress and the advance to candidate button is shown.
5. Activate the advance to candidate button.

Expected Results: The advance to candidate button is not available because there are missing required reports.

Actual result: Activating the advance to candidate button permits the user to move the test plan version into candidate review.

![screenshot of data management page showing meter in draft review](https://github.com/w3c/aria-at-app/assets/7241781/c205340a-0549-42ad-8310-e8d83d04851b)

![screenshot of advance to candidate confirmation dialog](https://github.com/w3c/aria-at-app/assets/7241781/0524993f-d3e8-46f3-8d22-13035d61e932)

Issue #784, opened 2023-09-22
Only show "Advance to Candidate" button if reports required during draft review are complete


mcking65:
The "Advance to Candidate" button is being displayed in the "Draft Review" column of the "Test Plan Status Summary" table before required reports are complete. If required reports are not complete, an error dialog is displayed. That prevents invalid conditions, but it would be better if the button were only displayed if conditions required for advancement have been met.

Tested on Sep 22, 2023 in staging.

mcking65:
Elivating to P1 because of #798. Rather than fix #798 by showing an error, it would be better to implement this change.

PR #812, opened 2023-10-13
Hide advance button when cannot advance

alflennik:
See https://github.com/w3c/aria-at-app/issues/784 and https://github.com/w3c/aria-at-app/issues/798.

Fixes an issue that the "Advance to Candidate" and "Advance to Recommended" buttons show on the Data Management page even when the required reports have not yet been collected.

### DIFF
--- a/client/components/DataManagement/DataManagementRow/index.jsx
+++ b/client/components/DataManagement/DataManagementRow/index.jsx
@@ -212,13 +212,6 @@ const DataManagementRow = ({
     }, [testPlanVersions]);
 
     const completedRequiredReports = testPlanVersion => {
-        // if (
-        //     testPlanVersion.title ===
-        //     'Action Menu Button Example Using element.focus()'
-        // ) {
-        //     debugger;
-        // }
-
         const reportAtBrowsers = testPlanVersion.testPlanReports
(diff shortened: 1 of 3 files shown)


client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -211,6 +211,39 @@ const DataManagementRow = ({
         );
     }, [testPlanVersions]);
 
+    const completedRequiredReports = testPlanVersion => {
```
stalgiag, line 214: Better to not prematurely optimize but just wanted to add a note that I can imagine a scenario in which we have hundreds of DRAFT or CANDIDATE test plan versions and we would want to cache these results. These array operations on each test plan version on each rerender (since they are used as rendering conditionals) could grow expensive. One idea is that it could be a custom hook that takes `testPlanVersions` as an arg and returns a `hasCompletedRequiredReportsMap{testPlanVersionId -> bool}` using `useMemo` internally.
Again, just a note for the future.

stalgiag (approved):
I'ved test and confirmed that this fixes both issues that were cited. The code is also clean, clear, and minimally intrusive. Thanks @alflennik !

Issue #793, opened 2023-09-26
It is possible to add invalid AT/Browser combinations to the test queue


mcking65:
Using the add to test queue function, I added and NVDA/Safari test plan run to the test queue in staging.

Related to #625.

![screenshot of confirmation dialog](https://github.com/w3c/aria-at-app/assets/7241781/2518ac3b-8842-4769-9dd0-a96544db9cb5)

![screenshot of test queue](https://github.com/w3c/aria-at-app/assets/7241781/fa428185-c1e4-45ae-9b51-bc9517e877a9)

PR #813, opened 2023-10-17
Validate At / Browser combinations in the Test Queue

Paul-Clue:
see #793

This Pull Request fixes the issue of admins being able to add invalid At and Browser combination test plan runs to the test queue.

While adding a test plan run to the test queue, the user was able to select the desired assistive technology and then still had the option of choosing any browser type along with it. This was a problem because not all Ats work with all browsers.

The fix makes it so that only valid assistive technologies and browser combinations could be selected in the UI.


alflennik (approved):
Great work, code looks clean, tests pass, and it works perfectly.

stalgiag:
Minimally intrusive, works well, good job!

stalgiag (approved):
minimally intrusive, works well, good job!

PR #815, opened 2023-10-18
Use pathToFileURL to support windows drive letter paths

gnarf:
Received this error trying to set the project up on windows:

```
Error found: Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'f:'
```

The `import()` line here was using `f:\\code\\aria-at\\aria-at-app` etc as a path (which is correct on windows) but requires this `url.pathToFileURL()` to be used for the `import(url)` statement or the `f:` gets processed as a url scheme for the import()

### DIFF
--- a/server/scripts/import-tests/index.js
+++ b/server/scripts/import-tests/index.js
@@ -3,6 +3,7 @@
 const path = require('path');
 const { Client } = require('pg');
 const fse = require('fs-extra');
+const { pathToFileURL } = require('url');
 const spawn = require('cross-spawn');
 const { At } = require('../../models');
 const {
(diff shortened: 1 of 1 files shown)


alflennik (approved):
Thanks for fixing this issue on Windows!! I did a check for other cases in that app that might suffer from the same issue, but it seems to be a one-off, this is the only place in the repo where we do a dynamic import. I also successfully ran the import on Mac so Mac support is unaffected, and the Linux-based CI ran the import as well, successfully. Thanks for contributing Corey!

ccanash:
Hi @mcking65 , this was included in dec 5th production release

Issue #789, opened 2023-09-23
Incorrect text in H1 and title tag of test plan review pages


mcking65:
The H1 and title tag of test plan review pages is using the name of the test plan directory instead of the title of the test plan. This creates inconsistency between the names of the links that opens the pages and the title of the target of the link. This inconsistency is a failure in accessibility requirements. It also creates confusion because it is inconsistent with the title of the test plan version history page.

The format of the H1 and title tag element should be :

>Review Test Plan for TEST_PLAN_TITLE

Examples:

The title of the review page for [Checkbox Example (Mixed-State)](https://aria-at-staging.w3.org/test-review/b3d0576a2901ea7f100f49a994b64edbecf81cff/checkbox-tri-state) is:

>Test plan review for pattern: checkbox-tri-state

But it should be:

>Review Test Plan for Checkbox Example (Mixed-State)

The title of the review page for [Color Viewer Slider](https://aria-at-staging.w3.org/test-review/1c6ef2fbef5fc056c622c802bebedaa14f2c8d40/horizontal-slider) is:

>Test plan review for pattern: horizontal-slider

But it should be:

>Review Test Plan for Color Viewer Slider

PR #816, opened 2023-10-19
Fix H1 title for test plan review page

Paul-Clue:
see issue #789 

This pull request changes the H1 value in the test plan review page from the name of the test plan directory to the actual title of the the test plan.

### DIFF
--- a/server/apps/test-review.js
+++ b/server/apps/test-review.js
@@ -347,7 +347,6 @@ app.get('/:commit/:pattern', async (req, res) => {
     const pattern = req.params.pattern;
     const testPlan = await getTestPlanById(pattern);
     const testPlanTitle = testPlan.title;
-    //  console.log("New Log", testPlan.title);
 
     const { commitDate, commitMessage } = await getCommitInfo(commit);
     const { tests, scripts: setupScripts } = await generateTests(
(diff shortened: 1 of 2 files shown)


server/apps/test-review.js
```diff
@@ -342,6 +345,9 @@ const generateTests = async (pattern, commit, commitDate) => {
 app.get('/:commit/:pattern', async (req, res) => {
     const commit = req.params.commit;
     const pattern = req.params.pattern;
+    const testPlan = await getTestPlanById(pattern);
+    const testPlanTitle = testPlan.title;
+    //  console.log("New Log", testPlan.title);
```
alflennik, line 350: Could you remove this line?

alflennik (approved):
Approved, assuming you remove that one console log. I verified it is working before / after, and I ran the tests successfully.

Code changes look good. I actually checked the code pretty carefully for this one, because I noticed that you added the first database-related call in this file. I wanted to make sure there isn't an important reason that database calls haven't yet been used. I found nothing that gives me doubt, so I do think the way you did this is the simplest and most straightforward approach.

stalgiag (approved):
I am also going to approve. Title renders correctly and the DB query is necessary.

I do think that `getTestPlanId` (and by extension `getTestPlans`) could stand to be improved so that we can specify columns for retrieval. In this instance, we only need one property but we end up with the result of `SELECT * FROM` which is actually a deep object. This is outside the bounds of this PR though.

Good job!

server/apps/test-review.js
Paul-Clue: Ok. Done

PR #827, opened 2023-10-26
Force IPv4 for internal HTTP requests

gnarf:
The backend localhost server only listens on IPv4, force the http agent from axios to use that when communicating with it's own backend via http.

Without this change, a computer that has the IPv6 address for localhost "first" in the hosts file will experience a hard crash when testing the bot workflows


stalgiag (commented):
Merged already so that is probably a clue but I approve. Thank you!

Issue #795, opened 2023-09-26
Change title of Test Reports page to "Assistive Technology Interoperability Reports"


mcking65:
Currently the reports page is titled "Test Reports". While that title is appropriate from the point of view of only the aria-at site information architecture, it doesn't help anyone outside the project or site understand what the pages in the reports section of the site contain. Imagine seeing that title in google search results; it tells you almost nothing.

Since the reports are our most important deliverable, I propose we title the reports pages to communicate their purpose and value.

Proposal for H1 and title tag:

>Assistive Technology Interoperability Reports

Proposal for link text in site navigation:

>AT Interop Reports

Or, if it fits:

>AT Interoperability Reports

mcking65:
@ccanash 

It appears this is in sandbox and the work on it seems complete. Is there something more that needs to be coded?

PR #828, opened 2023-10-26
Change "Test Reports" text, nav breakpoint change from 'lg' to 'xl'

stalgiag:
#### Text changes
The text changes in this PR are as follows:
"Test Reports" -> "AT Interop Reports" in metadata titles
"Test Reports" -> "AT Interoperability Reports" in the nav link label.
"Test Reports" -> "Assistive Technology Interoperability Reports" in headings.

#### Note on nav styles
Previously with "Test Reports", while logged in a line break would happen in the nav items with viewports between `992px-1070px` wide. At the lower end, the nav switches to a mobile view.

Now with "AT Interop Reports", the nav items have linebreaks with viewports between `992px-1114px`.

Given that the linebreaks were an issue with "Test Reports" and that all possible copy changes here would've expanded the linebreak viewing space, I decided it best to change the nav breakpoint to the "xl" bootstrap breakpoint.

see #795

### DIFF
--- a/client/components/App/App.css
+++ b/client/components/App/App.css
@@ -61,7 +61,7 @@ a.nav-link {
 .nav-link[aria-current='true']:after {
     background: #0b60ab;
 }
-@media all and (min-width: 992px) {
+@media all and (min-width: 1200px) {
     .navbar-expand-lg .navbar-collapse ul {
         display: flex;
         flex-basis: auto;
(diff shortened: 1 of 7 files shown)


howard-e (approved):
LGTM

alflennik (approved):
I verified the fixes and confirmed that there are no more errant references to "test reports" remaining, and wow, the word report is used a LOT in the codebase. Thanks for noticing that the nav bar needed to be tweaked, it's looking great.

PR #492, opened 2023-02-13
Upgrade vagrant environment

evmiguel:
This PR
- Upgrades the Vagrant environment
- Upgrades Node using Ansible
- Fixes bugs on fresh install of an ARIA-AT App instance. The bugs were:
  - Redirects going to backend instead of frontend
  - GitHub auth redirects not going to /test-queue

To mimic an upgrade in Vagrant:
- Start with a branch that uses Node 14: `git checkout vagrant-main-node-14`
- Launch a VM: `vagrant up`
- Verify that the app is working at http://aria-at-app.loc/
- `git checkout upgrade-vagrant-environment`
- Provision the VM again: `vagrant reload --provision`
- Verify that this task shows:
```
TASK [nodejs : Show status of nodejs installation] *****************************
ok: [vagrant] => {
    "msg": "v14.21.2"
}
```

and this task shows:
```
TASK [nodejs : Show status of new nodejs installation] *************************
ok: [vagrant] => {
    "msg": "v18.14.0"
}
```
- Verify that the app is working at http://aria-at-app.loc/ when provision completes

PR #831, opened 2023-11-01
Start Automation Collector via github action

gnarf:
## Github Workflow Automation Configuration
* The `jwt-signing-key.pem` file should be located in the project root folder.
  `ansible-vault view --vault-password-file ansible-vault-password.txt files/jwt-signing-key.pem.enc > ../jwt-signing-key.pem`
* The `AUTOMATION_CALLBACK_FQDN` environment variable in the environment configuration file should be a **fully qualified domain name** that is accessible from the github workflow server pointing at the running instance of aria-at-app.  For local development testing of these features, a forwarding proxy server like `ngrok` is recommended: `npx ngrok http 3000 --host-header=rewrite` will setup a server forwarding to your local 3000 development port.  You can then use the domain it gives you when launching the app:
  `AUTOMATION_CALLBACK_FQDN=128935b17294.ngrok.app yarn dev`


server/models/services/GithubWorkflowService.js
stalgiag, line 1: This might fit in better in `server/services`. Think most of the files in this folder are specifically for services that interact with the DB/Sequelize.

server/models/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,161 @@
...
+const WORKFLOW_REPO = 'bocoup/aria-at-gh-actions-helper';
+
+// Generated from the GitHub.com UI
+let privateKey = null;
+try {
+    privateKey = fs.readFileSync(
+        path.join(
+            __dirname,
+            '../../../config',
+            'aria-at-collection-scheduler.2023-09-27.private-key.pem'
```
stalgiag, line 30: I can envision this being a little annoying but it might be the best solve; haven't thought through it fully.

gnarf: same - I put it in as a "not sure how best to do this" moment - I'll add a comment that another approach might be better

config/dev.env
```diff
@@ -21,3 +21,4 @@ SESSION_SECRET='aria_at_report_dev'
 AUTOMATION_SCHEDULER_URL=http://localhost:8866
 AUTOMATION_SCHEDULER_SECRET=3Bq10QMDcca2IuxrTr23!eAsDCoILhHrrlBda7-3-WE9?El2S?R5REWXTyE3B72!
 AUTOMATION_SCHEDULER_PORT=8866
+AUTOMATION_CALLBACK_FQDN=
```
jugglinmike, line 24: Can we use `API_SERVER` for this?

server/models/services/CollectionJobService.js
```diff
@@ -329,19 +330,23 @@ const scheduleCollectionJob = async (
...
-            testPlanVersionGitSha: gitSha,
-            testIds: testIds ?? tests.map(t => t.id),
-            testPlanName: directory,
-            jobId
-        },
-        axiosConfig
-    );
+    let status = 'QUEUED';
+
+    if (!process.env.AUTOMATION_CALLBACK_FQDN) {
```
jugglinmike, line 335: I'd rather use a more semantic condition to make the code a little more self-documenting.

```suggestion
    if (process.env.ENVIRONMENT !== 'production') {
```

server/models/services/CollectionJobService.js
```diff
@@ -350,6 +355,12 @@ const scheduleCollectionJob = async (
         COLLECTION_JOB_ATTRIBUTES,
         options
     );
+
+    if (process.env.AUTOMATION_CALLBACK_FQDN) {
```
jugglinmike, line 359: To reduce the potential for discrepancies between development and production, I'd prefer to increase parity between the code paths. Here, that would mean a consistent sequence of job creation followed by collection scheduling.

server/models/services/CollectionJobService.js
```diff
@@ -350,6 +355,12 @@ const scheduleCollectionJob = async (
...
+        await createGithubWorkflow({ job, directory, gitSha });
```
jugglinmike, line 360: Can we catch exceptions and cancel the collection job?

server/models/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,157 @@
...
+
+        throw new Error(message);
+    }
+
+    return (await response.json()).token;
+};
+
+const createGithubWorkflow = async ({
+    job,
+    directory = 'alert',
```
jugglinmike, line 101: A default here doesn't seem like a convenience that would benefit us in production. I'd rather fail in the absence of a value.

server/models/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,157 @@
...
+    gitSha = 'master'
```
jugglinmike, line 102: Since this accepts branch names, `gitRef` would be more appropriate.

server/models/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,157 @@
...
+        exp: calculateExpiresAt(),
+        iss: GITHUB_APP_ID
+    };
+    const jsonWebToken = await createJWT(payload, privateKey, ALGORITHM);
+    const accessToken = await fetchInstallationAccessToken(
+        jsonWebToken,
+        GITHUB_APP_INSTALLATION_ID
+    );
+    const inputs = {
+        nonce: job.id,
```
jugglinmike, line 115: Does the system use this?

server/models/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,157 @@
...
+        callback_url: `https://${process.env.AUTOMATION_CALLBACK_FQDN}/api/jobs/${job.id}/result`,
+        status_url: `https://${process.env.AUTOMATION_CALLBACK_FQDN}/api/jobs/${job.id}/update`,
+        // TODO: This should probably just be configured instead of us passing it to the scheduler every time
+        callback_header: `x-automation-secret:${process.env.AUTOMATION_SCHEDULER_SECRET}`,
```
jugglinmike, line 119: It seems to me that passing it every time alleviates us from the need to synchronize a setting between two systems (ARIA-AT App and GitHub Actions). Is that right? What are the drawbacks?

server/models/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,157 @@
...
+                return 'Unable to initiate workflow';
+            }
+        })();
+
+        throw new Error(message);
+    }
+
+    // TODO: Listen for GitHub Webhook event describing the new workflow, and
+    // return its ID.
+    return true;
```
jugglinmike, line 154: I believe this is no longer the plan. If so, we can remove the comment and the return value.

server/models/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,157 @@
...
+
+// Generated from the GitHub.com UI
+let privateKey = null;
+try {
+    privateKey = fs.readFileSync(
+        path.join(__dirname, '../../../jwt-signing-key.pem')
+    );
+} catch (e) {
+    // Empty catch block, if the key is null we will hear about it later.
+}
```
jugglinmike, line 31: In production scenarios, multiple days may pass between the deployment of this code and the use of the feature. If the person deploying has failed to provide this critical file, we want them to be made aware of that immediately.

jugglinmike (changes requested):
Regarding handling secrets: this project already has a solution for that. I'll spare you the details, but if you're curious, you can see how it's hooked up in [the commit I just pushed to this branch](https://github.com/w3c/aria-at-app/pull/831/commits/e69e06d59865e60f09c50dde31c47706fffadfcb).

server/models/services/CollectionJobService.js
gnarf: I had initially thought of this approach, but that setup would make testing in development branch harder, which is why I used an environment variable that could act as a toggle regardless of production/dev/testing "node envrionment" - I am happy to make it a more obvious one like ENABLE_AUTOMATION_JOB_CREATE or something similar (not sold on the name, feel free to paint the shed)

config/dev.env
gnarf: Makes sense to me!

server/models/services/CollectionJobService.js
gnarf: I agree. I'll make sure to reduce the switching around of path/order on mock test/real automation

server/models/services/CollectionJobService.js
gnarf: I.E. if the create job fails set collection job to canceled immediately? If I understood correctly, this seems smart! Happy to do

server/models/services/GithubWorkflowService.js
gnarf: not actively, it does send it back but it fairly "unneeded" in the current setup

server/models/services/GithubWorkflowService.js
gnarf: I think my goal with putting this here instead of synced between the two servers was some day to have this `x-automation-secret` be a secret key per job instead of globally.  As it works now, (for MVP) we could define this key as a permanent secret in the github action repo and remove it from here entirely (assuming we use the same key across all environments we want to be able to actively test the github workflow against)

I doubt there are actors trying to sabotage our test results, but by exchanging the key, and having it be individual per job should greatly reduce the ability for someone to corrupt/fake/hack the data collection process.

server/models/services/GithubWorkflowService.js
gnarf: My plan: re-throw on production/ENABLE_AUTOMATION only - ignore the error about missing file in dev environments that aren't using it to create github jobs?

server/models/services/CollectionJobService.js
gnarf: AUTOMATION_JOB_SCHEDULER=production|mock

jugglinmike: I'm generally averse to complexities in the definition of the deployment environment. Specifically in this case, that means creating a two-dimensional space rather than the current enumeration. Any complexity complicates maintenance (since contributors have to perform some mental combinatorics to determine if/when their work needs to be modulated), and the introduction of an independent variable often creates states which aren't needed. All of these things also pose a hazard during deployment since they introduce new opportunities for misconfiguration.

[We were considering a similar change earlier this year, but we managed to find a solution that minimized the configuration space.](https://github.com/w3c/aria-at-app/pull/492#issuecomment-1447395777) I don't know if that remediation is appropriate here, but it does motivate me to ask: what kind of difficulties arise from using the existing `ENVIRONMENT` environment variable for this purpose?

server/models/services/CollectionJobService.js
jugglinmike: Exactly!

server/models/services/GithubWorkflowService.js
jugglinmike: Let's 86 it, then

server/models/services/GithubWorkflowService.js
jugglinmike: It sounds like we are in agreement, after all! Let's remove the "TODO" comment because we don't actually desire the configuration-based approach it suggests.

server/models/services/GithubWorkflowService.js
jugglinmike: I'd like to shrink this testing seam (really, *all* testing seems) as much as possible. Doing so makes it easier to reason about each component's behavior. In this case, since the project uses `require`, we can allow this exception to propagate implicitly if we defer the loading of the `GithubWorkflowService.js` module in to the existing seam in `CollectionJobService.js`.

gnarf:
Squashed (to make the rebase cleaner) then rebased on automation-mvp

server/models/services/CollectionJobService.js
gnarf: I tend to agree with the points and the linked issue. My reluctance is more about not wanting to have to set my local server up in "production" to be able to test full cycle communications - I think 95% of the time local development environment will want to mock these jobs, but for the situations in which we do want to work directly on local (like working on this PR or future changes to these services), switching my local dev environment to "production" mode has other possible unknown (to me atm) behavior switches.  If you don't think there are any reasons to fear this step for local testing of GitHub agent job creation then I totally agree with the assessment.

config/dev.env
gnarf: Was also realizing while looking at the other configuration settings, on the `dev` config, API_SERVER is set to `localhost:8000` - this really needs to be an FQDN that the github action can POST to, and may or may not be equal to the `API_SERVER` value

jugglinmike:
Thanks, @gnarf! Given that the `APP_SERVER` value cannot be used as the automation callback, I'll support the introduction of the `AUTOMATION_CALLBACK_FQDN` environment variable. There are two steps we can take in this patch to reduce the risk: document the environment variables with comments in `config/dev.env` (adding a call-out to those descriptions in `deploy/README.md` for discoverability), and throw at startup if this variable is defined when `ENVIRONMENT` is anything other than `dev` or `vagrant`. What do you think?

deploy/README.md
```diff
@@ -109,3 +109,8 @@ From the `deploy` folder:
 2. Load the backup that was created
   `psql -d aria_at_report -f <environment>_dump_<timestamp>.sql`
 
+## Github Workflow Automation Configuration
+* The `jwt-signing-key.pem` file should be located in the project root folder.
+  `ansible-vault view --vault-password-file ansible-vault-password.txt files/jwt-signing-key.pem.enc > ../jwt-signing-key.pem`
+* The `AUTOMATION_CALLBACK_FQDN` environment variable in the environment configuration file should be a **fully qualified domain name** that is accessible from the github workflow server pointing at the running instance of aria-at-app.  For local development testing of these features, a forwarding proxy server like `ngrok` is recommended: `npx ngrok http 3000 --host-header=rewrite` will setup a server forwarding to your local 3000 development port.  You can then use the domain it gives you when launching the app:
+  `AUTOMATION_CALLBACK_FQDN=128935b17294.ngrok.app yarn dev`
```
jugglinmike, line 116:
```suggestion
  `AUTOMATION_CALLBACK_FQDN=128935b17294.ngrok.app yarn dev`
```

server/app.js
```diff
@@ -43,8 +43,20 @@ listener.route('/aria-at/:branch*').get(
...
+    process.env.ENVIRONMENT === 'production' ||
+    process.env.AUTOMATION_CALLBACK_FQDN
+) {
+    require('./services/GithubWorkflowService')
+        .setup()
+        .catch(error => {
+            console.error(
+                'Failed to initialize github workflow service:',
+                error
+            );
```
jugglinmike, line 57:
```suggestion
            error.message = `Failed to initialize github workflow service: ${error.message}`;
            throw error;
```

Or remove the `catch` invocation altogether. I just want the process to quit because this is a critical error.

jugglinmike (changes requested):

jugglinmike:
It looks like between file renaming and branch rebasing, GitHub has incorrectly designated some of my comments as "outdated". In any event, these still stand:

- https://github.com/w3c/aria-at-app/pull/831#discussion_r1395021807
- https://github.com/w3c/aria-at-app/pull/831#discussion_r1395022431
- https://github.com/w3c/aria-at-app/pull/831#discussion_r1395022738
- https://github.com/w3c/aria-at-app/pull/831#discussion_r1395022298

server/models/services/GithubWorkflowService.js
gnarf: moved this

server/models/services/CollectionJobService.js
gnarf: Wrapped up a try/catch point for this in the `triggerWorkflow` helper

server/models/services/GithubWorkflowService.js
gnarf: This also correlates to the `testPlanVersion.gitSha`

server/models/services/GithubWorkflowService.js
gnarf: TODO removed

server/models/services/GithubWorkflowService.js
gnarf: there isn't a "value" here; basically success or throw; so left `return true`

server/services/GithubWorkflowService.js
```diff
@@ -0,0 +1,172 @@
...
+
+// > The expiration time of the JWT, after which it can't be used to request an
+// > installation token. The time must be no more than 10 minutes into the
+// > future.
+//
+// https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app
+const calculateExpiresAt = () => Math.round(Date.now() / 1000) + 9 * ONE_MINUTE;
+
+const fetchInstallationAccessToken = async (jsonWebToken, installationID) => {
+    const response = await fetch(
```
gnarf, line 95: We might need to port this to axios.post or upgrade minimum node version to 18

gnarf: (there is a second call to fetch further down in this file also)

jugglinmike:
During my local testing, the ARIA-AT App server encountered an error while processing six results from GitHub Actions. I determined this from the information available in the ngrok command-line interface:

>     HTTP Requests
>     -------------
>     
>     POST /api/jobs/59/update       200 OK
>     POST /api/jobs/59/result       500 Internal Server Error
>     POST /api/jobs/59/result       500 Internal Server Error
>     POST /api/jobs/59/result       500 Internal Server Error
>     POST /api/jobs/59/result       500 Internal Server Error
>     POST /api/jobs/59/result       500 Internal Server Error
>     POST /api/jobs/59/result       500 Internal Server Error
>     POST /api/jobs/59/result       200 OK
>     POST /api/jobs/59/result       200 OK
>     POST /api/jobs/59/result       200 OK

Unfortunately, neither the server nor the harness logged any information about these errors, so I'll need to do further research to learn what went wrong.

I took this as an opportunity to test the recovery mechanism we built into the system. There appears to be a problem with that feature, as well. Clicking the button labeled "Stop Running" correctly disabled that button, but it didn't update the "queued" tests to "cancelled", and so the button labeled "Retry Cancelled Collections" remains disabled.

I'll continue investigating these anomalies tomorrow (I've also begun prototyping patches to both the server and the harness to document errors like these in their runtime logs).

gnarf:
This might be fetch missing if node<18 let me fix that for axios

gnarf:
Hrm, I know @stalgiag and I managed to get it working on her local (that's where we discovered axios.post vs fetch), wondering if maybe there is a migration issue on your local?

stalgiag:
> Hrm, I know @stalgiag and I managed to get it working on her local (that's where we discovered axios.post vs fetch), wondering if maybe there is a migration issue on your local?

Yeah I was able to get it working on my local with one test plan but these look like problems with specific test results so this might be test dependent. That 500 response should have an error message in it. Would be good to log those on both the workflow side and the app server side.

jugglinmike:
Today, Stalgia and I began honing in on a solution. [Our work-in-progress is available here.](https://github.com/w3c/aria-at-app/compare/automation-mvp...bocoup:aria-at-app:automation-mvp-fix-test-references)

server/services/GithubWorkflowService.js
gnarf: 57a1e690

jugglinmike (approved):
Way to go, folks!

jugglinmike:
It looks like [my latest commit to the automation-mvp branch](https://github.com/w3c/aria-at-app/commit/6269a96156399bdc7d898a9357605f0b239a5c38) broken a few tests. I'm going to see about fixing that before merging this branch.

Issue #800, opened 2023-09-27
Advance to recommended button is not consistently available when it seems base requirements are met


mcking65:
In staging, I advanced the meter plan to candidate. Required reports are complete. There are 0 open issues. The advance to recommended button is not available.

![screenshot of data management page](https://github.com/w3c/aria-at-app/assets/7241781/2fbac7d7-df48-4e4a-9538-9ddb5206aabf)

I am guessing there is a date-based rule associated with the visibility of the button. If my guess is correct, to resolve this bug, please remove it and allow admins to advance sooner than the timeframe specified in the working mode.

While it is not yet apparent in the current version of the working mode, there are scenarios where at its discretion, the CG should be able to advance a new version to recommended sooner than the timeframe specified in the working mode. For example, a new version of a plan contains only a minor but still substantive change and that change impacts only one AT. In that case, 180 days is unnecessary. They agreement may have been made in advance, so 0 days may be needed.

In lieu of preventing advancement on a faster time table, the confirmation dialog could contain the following text:

>Warning! the TEST_PLAN_NAME test plan has been in candidate review for X days, which is Y fewer days than recommended by the ARIA-AT working mode. Confirm this action only if impacted stakeholders have had sufficient opportunity to provide consensus.

In addition, as part of this proposed resolution, make it possible for admins to control the recommended number of days for candidate review by adding a workingMode object to property to /tests/support.json with a property for the length of candidate review in days:

~~~
{
  "workingMode": {"numberCandidateReviewDays": 180}
}
~~~

mcking65:
Potentially related to #798

lolaodelola:
Hi @mcking65,

Just wanted to follow up on this issue and get some clarification/draw your attention to some potential future problems, especially concerning

> In addition, as part of this proposed resolution, make it possible for admins to control the recommended number of days for candidate review by adding a workingMode object to property to /tests/support.json with a property for the length of candidate review in days

This presents a maintenance burden and potential gotcha down the line. Within our current architecture, we don't store config data in the ARIA-AT repo except for the tests. Introducing config data in this repo creates two sources of truth that any engineer engaging with this will need to know about, an additional source of complexity, and the configuration will need to be piped through the entire import system, including the database, and as mentioned it’s a deviation from our current architecture

There is already UI within the app to do this work, does that UI not meet your needs? As we build we want to ensure we're not adding redundancy as it'd be a shame if, considering the work we'd need to do to implement and maintain this, it's unused.

[...]

Issue #758, opened 2023-08-30
Fix required reports data


mcking65:
The report status dialogs are showing 6 reports required during candidate review. Only 3 are required for draft and candidate:

1. jaws/chrome
2. NVDA/Chrome
3. VoiceOver/Safari
3 more are required in recommended phase:

1. JAWS/Firefox
2. NVDA/Firefox
3. VoiceOver/Chrome

howard-e:
This must have been a misunderstanding for the condition when updating to `Recommended`.

1. The thought was those additional 3 were required BEFORE being able to transition to `Recommended`.
2. That means the `Required Reports` button should also be available on the `Recommended Version` cell and it will most likely say `Some Required Reports Missing`, when initially transitioned to.

@mcking65 if 1 was a false assumption and 2 is the expectation, then #764 addresses this. Could you confirm?

mcking65:
> This must have been a misunderstanding for the condition when updating to `Recommended`.
> 
> 1. The thought was those additional 3 were required BEFORE being able to transition to `Recommended`.

No, we can advance to recommended without them.

> 2. That means the `Required Reports` button should also be available on the `Recommended Version` cell and it will most likely say `Some Required Reports Missing`, when initially transitioned to.

Definitely, that is the expectation. There are other conditions where it will appear in the recommended phase column. There is a to-do in issue #648 for me to work with you on the algorithm for that.

> @mcking65 if 1 was a false assumption and 2 is the expectation, then #764 addresses this. Could you confirm?

Confirmed.

howard-e:
> > @mcking65 if 1 was a false assumption and 2 is the expectation, then #764 addresses this. Could you confirm?
> 
> Confirmed.

Thanks! #764 has been added to the sandbox and staging environments

mcking65:
Confirmed that the issue is resolved in staging, at least for draft and candidate test plans.

ccanash:
released. #792 and #791 will be tackled separately.

PR #832, opened 2023-11-01
Allow candidate test plans to advance with override

alflennik:
See issue https://github.com/w3c/aria-at-app/issues/800. Note this does not address the new source of configuration which was requested in that issue.


client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -854,18 +862,10 @@ const DataManagementRow = ({
...
-                    );
-                    const DAYS_TO_PROVIDE_FEEDBACK = 120;
                     const shouldShowAdvanceButton =
                         isAdmin &&
                         completedRequiredReports(latestVersion) &&
-                        issuesCount === 0 &&
-                        (recommendedTestPlanVersions.length ||
-                            (!recommendedTestPlanVersions.length &&
-                                daysBetweenDates > DAYS_TO_PROVIDE_FEEDBACK));
+                        issuesCount === 0;
```
howard-e, line 868: @alflennik has there been any discussion on also allowing the advance to happen even if there is open issues related to the test plan? #800 correctly assumes that the button wasn't available due to the candidate review time constraint. But an additional factor would be currently open issues (also need to consider that an open issue doesn't have be tied to that specific test plan version, with recent updates).

Do we forsee a need to also bypass that?

cc @lolaodelola

client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -854,18 +862,10 @@ const DataManagementRow = ({
...
                         );
 
-                    const daysBetweenDates = checkTimeBetweenDates(
-                        currentDate,
-                        latestVersion.candidatePhaseReachedAt
-                    );
-                    const DAYS_TO_PROVIDE_FEEDBACK = 120;
                     const shouldShowAdvanceButton =
                         isAdmin &&
                         completedRequiredReports(latestVersion) &&
```
howard-e, line 867: Not a change in this PR, but this function prevents the button from appearing for a `CANDIDATE` test plan version, unless the reports required for a recommended test plan version has been met. Has that [requirement since changed](https://github.com/w3c/aria-at-app/issues/758#issuecomment-1699694393) again?

howard-e (commented):
@alflennik this change looks good and I've verified the warning is displayed when expected. It should definitely grant some more freedom during the advancement process!

I do have 2 follow up questions, in the thread below.

client/components/DataManagement/DataManagementRow/index.jsx
alflennik: Interesting point! I tried to follow requested change as close as I could, but I'd be willing to address that too if there's consensus about it being helpful.

client/components/DataManagement/DataManagementRow/index.jsx
alflennik: OK, noted! Let me know if you need me to do anything about this line.

alflennik: We talked it over in a meeting! Thanks for clarifying what you meant Howard. Basically it should be possible to advance to recommended even without all the recommended reports completed. I will update this line to support that.

howard-e (approved):
@alflennik LGTM! Thanks for addressing the feedback

I also acknowledge that my discussion point on relaxing the constraints around the issues count is separate and can be done in a follow up issue (and PR if needed)

alflennik:
@howard-e this PR now incorporates the fix you requested before!

howard-e (approved):
> @howard-e this PR now incorporates the fix you requested before!

Thanks for checking that out

> @alflennik LGTM!

x2 🙂

Paul-Clue (approved):

PR #833, opened 2023-11-06
No longer mark test plan reports with skipped tests as final

Paul-Clue:
This pull request removes the ability for a user to mark test reports as final that have skipped tests.

- The `Mark As Final` button will not show in the test queue for any reports that have skipped tests.
- The system (database and some files) was updated to no-longer consider the possibility of reports being marked as final that have skipped tests.
- A migration was introduced that would send tests that are not completed back to the test queue that have already been finalized.

### DIFF
--- a/client/components/Reports/SummarizeTestPlanReport.jsx
+++ b/client/components/Reports/SummarizeTestPlanReport.jsx
@@ -10,7 +10,6 @@ import {
     faExternalLinkAlt,
     faHome
 } from '@fortawesome/free-solid-svg-icons';
-import { differenceBy } from 'lodash';
 import { convertDateToString } from '../../utils/formatter';
 import DisclaimerInfo from '../DisclaimerInfo';
 import TestPlanResultsTable from '../common/TestPlanResultsTable';
@@ -99,12 +98,6 @@ const SummarizeTestPlanReport = ({ testPlanVersion, testPlanReports }) => {
         at,
         browser
     };
-
-    const skippedTests = differenceBy(
-        testPlanReport.runnableTests,
-        testPlanReport.finalizedTestResults,
-        testOrTestResult => testOrTestResult.test?.id ?? testOrTestResult.id
-    );
 
     return (
         <Container id="main" as="main" tabIndex="-1">
@@ -231,26 +224,6 @@ const SummarizeTestPlanReport = ({ testPlanVersion, testPlanReports }) => {
                     </Fragment>
                 );
             })}
-            {skippedTests.length ? (
-                <Fragment>
-                    <div className="skipped-tests-heading">
-                        <h2 id="skipped-tests" tabIndex="-1">
-                            Skipped Tests
-                        </h2>
-                        <p>
-                            The following tests have been skipped in this test
-                            run:
-                        </p>
-                    </div>
-                    <ol className="skipped-tests">
-                        {skippedTests.map(test => (
-                            <li key={test.id}>
-                                <a href={test.renderedUrl}>{test.title}</a>
-                            </li>
-                        ))}
-                    </ol>
-                </Fragment>
-            ) : null}
         </Container>
     );
 };
--- a/client/components/TestQueueRow/index.jsx
+++ b/client/components/TestQueueRow/index.jsx
@@ -414,9 +414,9 @@ const TestQueueRow = ({
             'completed'
         ].join('-');
 
-    const completedAllTests = !testPlanReport.draftTestPlanRuns.find(
+    const completedAllTests = testPlanReport.draftTestPlanRuns.every(
         testPlanRun =>
-            testPlanRun.testResultsLength !== testPlanReport.runnableTestsLength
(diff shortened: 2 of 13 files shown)


server/scripts/populate-test-data/populateFakeTestResults.js
```diff
@@ -115,34 +147,27 @@ const populateFakeTestResults = async (testPlanRunId, fakeTestResultTypes) => {
 
         index += 1;
     }
+
+    if (testPlanReport.runnableTests.length !== fakeTestResultTypes.length) {
```
howard-e, line 151: Can a comment be included on why this is being added as well? Outside the context of this PR, it may take a bit longer to follow on why, for any future maintenance.

client/components/Reports/SummarizeTestPlanReport.jsx
```diff
@@ -269,26 +263,6 @@ const SummarizeTestPlanReport = ({ testPlanVersion, testPlanReports }) => {
                     </Fragment>
                 );
             })}
-            {skippedTests.length ? (
```
howard-e, line 272: It's great getting rid of these lines

howard-e (commented):
Looks good! I've only left an inline comment on sharing some clarifying comments for the future.

There's also a bit of `.skipped` related classes you could get rid of in `Reports.css` and `TestRun.css` to finish that cleanup.

server/migrations/20231031162340-verifyNoSkippedTests.js
```diff
@@ -0,0 +1,67 @@
...
+            });
+
+            for (const data of Object.values(dataById)) {
+                let isComplete = true;
+
+                data.runLengths.forEach(runLength => {
+                    if (runLength !== data.runnableTestsLength) {
+                        isComplete = false;
+                    }
+                });
```
stalgiag, line 47:
```suggestion
                    const isComplete = data.runLengths.every(runLength => runLength === data.runnableTestsLength);

```
Another instance where it could be useful to use `.every` for an affirmative calculation in one line.

server/resolvers/testPlanReportsResolver.js
```diff
@@ -30,12 +30,7 @@ const testPlanReportsResolver = async (
         attributes: testPlanReportAttributes
     } = retrieveAttributes('testPlanReport', TEST_PLAN_REPORT_ATTRIBUTES, info);
 
-    const { attributes: testPlanRunAttributes } = retrieveAttributes(
-        'draftTestPlanRuns',
-        TEST_PLAN_RUN_ATTRIBUTES,
-        info,
-        true
-    );
+    const testPlanRunAttributes = TEST_PLAN_RUN_ATTRIBUTES;
```
stalgiag, line 33: We could use the imported variable directly instead of copying a reference here.

stalgiag (changes requested):

stalgiag (changes requested):
A few small changes requested in-line but overall looks great! Thanks for doing this work!

server/scripts/populate-test-data/populateFakeTestResults.js
Paul-Clue: Comment added.

client/components/TestQueueRow/index.jsx
```diff
@@ -414,6 +414,11 @@ const TestQueueRow = ({
             'completed'
         ].join('-');
 
+    const completedAllTests = !testPlanReport.draftTestPlanRuns.find(
+        testPlanRun =>
+            testPlanRun.testResultsLength !== testPlanReport.runnableTestsLength
+    );
```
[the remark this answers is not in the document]
Paul-Clue: Change made.

server/resolvers/testPlanReportsResolver.js
Paul-Clue: Hey can you elaborate a little more on this.

stalgiag: Since this is no longer a calculation and it also isn't used anywhere else, you can use the `TEST_PLAN_RUN_ATTRIBUTES` imported variable directly at the bottom where this value is passed in `getTestPlanReports`. No need to create a new variable named `testPlanRunAttributes`.

stalgiag: I mean really, it looks like you could even pass in null instead and not import`TEST_PLAN_RUN_ATTRIBUTES` in this file as that same import is the [default value for that parameter in `getTestPlanReports`.](https://github.com/w3c/aria-at-app/blob/no-skipped-tests/server/models/services/TestPlanReportService.js#L158) This is a personal preference thing though. I like being explicit and doing my first suggestion.

Paul-Clue: Ok. I went with the first suggestion. I now totally understand and agree with the change.

stalgiag (approved):
Looks good to me. Thanks @Paul-Clue !

howard-e (approved):
LGTM!

PR #834, opened 2023-11-07
Fix timeout issues with dataManagement tests

stalgiag:
See title.

Global jest timeout is typically set outside of any hooks and hooks like `afterAll` have a different timeout context. This should fix the intermittently failing tests.


howard-e (approved):
@stalgiag looks good to me. Thanks for taking the time to investigate and solve this!

ccanash:
Hi @mcking65  this was included in yesterday's release

PR #835, opened 2023-11-07
Implement target date modal improvements

alflennik:
Implements fixes listed in this issue: https://github.com/w3c/aria-at-app/issues/785 As we discussed there are lingering questions about one of the notes so we will start a discussion about it.

### DIFF
--- a/client/components/DataManagement/DataManagementRow/index.jsx
+++ b/client/components/DataManagement/DataManagementRow/index.jsx
@@ -9,7 +9,7 @@ import {
 } from '../queries';
 import { LoadingStatus, useTriggerLoad } from '../../common/LoadingStatus';
 import {
-    checkTimeBetweenDates,
+    checkDaysBetweenDates,
     convertDateToString,
     convertStringFormatToAnotherFormat
 } from '../../../utils/formatter';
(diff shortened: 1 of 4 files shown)


howard-e (changes requested):
@alflennik the changes look good to me other than for (5). That doesn't seem to have been addressed as yet.

From https://github.com/w3c/aria-at-app/issues/785#issuecomment-1749887927:
> > Accessibility issue: The dialog is not labeled. It has aria-labelledby, but the idref is invalid. It should reference the heading.
>
> This is still the same. The aria-labelledby value on the div with role dialog is not referring to the heading element inside the dialog.

screenshot showing a role=dialog div with an aria-labelledby id, which doesn't match any known ids

This PR also doesn't address (1) because you've explained that you'd like to have a follow up discussion regarding that condition. I'd be okay approving without it and adding it later if so, as it's a small change. But in the meantime, that discussion can continue in #785.

Also from https://github.com/w3c/aria-at-app/issues/785#issuecomment-1749887927, can support be added for:

> ... other bug ... the dialog does not close when escape is pressed.

stalgiag:
> > ... other than for (5). That doesn't seem to have been addressed as yet.
> 
> @alflennik perhaps this point of feedback shouldn't block the PR after all. #837 seems to be resolving it separately. Could you confirm?
> 
> cc @stalgiag

Yes all instances of this problem should be fixed by that PR

alflennik:
@howard-e the escape key support has been implemented.

howard-e (approved):
LGTM and confirmed that once #837 is merged with this, it does address the `aria-labelledby` issue.

Issue #811, opened 2023-10-06
Report status dialog missing accessible name


mcking65:
The report status dialog is contained in a div declared with the HTML:

`<div role="dialog" aria-modal="true" class="modal show" tabindex="-1" aria-labelledby="basic-modal" style="display: block;">`

The id "basic-modal" does not appear to exist. The element that should have an ID that is referenced by aria-labeledby is:

`<h1 class="modal-title css-yefjat" tabindex="-1">`

mcking65:
@ccanash 

The status is still showing as to-do, but there is a PR and the issue appears fixed in the sandbox.

ccanash:
Hi @mcking65 , I  have updated the status and will confirm it was released to production tomorrow.

PR #837, opened 2023-11-09
Fix aria-labeledby for BasicModal/BasicThemedModal

stalgiag:
see #811 

The issue existed in the underlying component. This adds a unique id to the title of the modal and using that id for `aria-labeledby` in both BasicModal and BasicThemedModal.

This also adds a unit test to ensure the attribute correctly points to the title.


howard-e (approved):

alflennik (approved):
Code looks great, thanks for adding a test as well. I reproduced the issue and verified the PR fixes it.

Issue #836, opened 2023-11-08
The version string elements are empty when test plan version is advanced on Data Management Page


ccanash:
After advancing a test plan version on the Data Management page, the text included for the version string elements in the cells of that row is empty. A refresh solves it which indicates a data persistence problem with the version string values.

Resolving this should also solve a related comment in https://github.com/w3c/aria-at-app/issues/785#issuecomment-1749887927:

> After saving a new target date value, the link for the test plan version is getting corrupted. The text of the link goes away. In particular, this element inside of the link, `<b>V22.03.17</b>`, is empty after saving a new target date value.

mcking65:
Verified fix.

PR #839, opened 2023-11-09
Fix empty version strings on Data Management and Test Queue pages

Paul-Clue:
See issue #836 

This pull request solves the issue of the version string disappearing after advancing a test plan version or changing the target date on the data management page.
Simply adding `versionString` to both mutations in the data management queries solved the issue.


howard-e (approved):

stalgiag (approved):
Fix is confirmed in both Data Management and Test Queue pages. Thank you @Paul-Clue and @howard-e !

PR #840, opened 2023-11-13
WIP: Support v2 Test Format Import

howard-e:


client/components/Reports/getMetrics.js
```diff
@@ -91,35 +91,44 @@ const countUnexpectedBehaviors = ({
...
+    scenarioResult, // Choose one to provide
+    testResult, // Choose one to provide
+    testPlanReport // Choose one to provide
+}) => {
+    const result = { scenarioResult, testResult, testPlanReport };
+    const {
+        assertionsCount: requiredAssertionsCount,
+        assertionsPassedCount: requiredAssertionsPassedCount,
+        assertionsFailedCount: requiredAssertionsFailedCount
+    } = calculateAssertionPriorityCounts(result, 'REQUIRED');
```
howard-e, line 119: For reviewers: A follow up task to replace instances of `REQUIRED` and `OPTIONAL` could follow this PR landing for consistency. I'd rather not inflate this PR further.

server/resolvers/TestResultOperations/saveTestResultCommon.js
```diff
@@ -92,8 +98,30 @@ const saveTestResultCommon = async ({
 
     await updateTestPlanRun(testPlanRun.id, { testResults: newTestResults });
 
-    // TODO: Avoid blocking loads in test runs with a larger amount of tests
-    //       and/or test results
+    if (isSubmit) {
```
howard-e, line 101: Adding this block to address an easily reproducible process breakage where the metrics of a TestPlanReport could fail to be tracked.

Scenario:
1. Tester is assigned to a Test Run on Test Queue
2. They go through the tests for the assigned run in their Test Run page, submitting results along the way
3. Returns to the Test Queue page
4. Reselects their assigned Test Run
5. Selects "Edit Results"
6. Makes a change to the test assertion that's collected or unexpected behaviors
7. Selects "Submit Results" (they never navigated to another test in the Test Run)

The expectation is that the metrics would be recalculated in the instance of a "Success case" going to a "Failure case". But this doesn't happen. It seems the metrics were only being recorded on test transition within the test run. This change makes sure it is also done when those results are submitted.

server/scripts/populate-test-data/pg_dump_2021_05_test_data.sql
```diff
@@ -12,14 +12,18 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
-CREATE OR REPLACE FUNCTION get_test_plan_version_id(testPlanVersionTitle text) RETURNS integer
+CREATE OR REPLACE FUNCTION get_test_plan_version_id(testPlanVersionTitle text, testFormatNumber text) RETURNS integer
```
howard-e, line 15: To eventually make this `.sql` sufficiently test for v2 test cases

howard-e:
I'm unable to reproduce these test failures locally. Will check out why this is.

server/scripts/import-tests/index.js
```diff
@@ -346,58 +433,179 @@ const getTests = ({ builtDirectoryPath, testPlanVersionId, ats, gitSha }) => {
...
+                            assertionPhrase: assertion.assertionPhrase
+                        };
+                    }),
+                    viewers: []
+                };
+
+                tests.push(test);
+            }
+        } else {
+            const common = allCollected[0];
```
stalgiag, line 553: This block reads quite densely with quite a bit of logic inside of the object construction. Not sure if breaking up into smaller functions would help but just wanted to note that it was challenging to read this.

server/tests/integration/testPlanRun.test.js
```diff
@@ -15,17 +15,21 @@ let assertionResultId1;
 let assertionResultId2;
 let assertionResultId3;
 
-const getTestPlanVersionId = async () => {
+const getTestPlanVersionId = async (testFormatVersion = 1) => {
```
stalgiag, line 18: A unit test for a v2 test plan version would be great. I assume it is going to happen with other work but just noting that desire here since I see a pathway opened up.

server/util/getMetrics.js
```diff
@@ -74,35 +78,44 @@ const countUnexpectedBehaviors = ({
     return countScenarioResult(scenarioResult);
 };
```
stalgiag, line 80: Not unique to this PR but curious why these utilities get copied across client and server. If they were pure functions, would they be usable across both since we have the monorepo?

stalgiag (approved):
LGTM! I left a couple of small, optional changes and points of discussion in-line.

Thanks for this work!

config/test.env
```diff
@@ -12,6 +12,8 @@ IMPORT_CONFIG=../config/test.env
 
 IMPORT_ARIA_AT_TESTS_COMMIT_1=5fe7afd82fe51c185b8661276105190a59d47322
 IMPORT_ARIA_AT_TESTS_COMMIT_2=1aa3b74d24d340362e9f511eae33788d55487d12
+# Commit before incoming v2 test format changes
+IMPORT_ARIA_AT_TESTS_COMMIT_3=836fb2a997f5b2844035b8c934f8fda9833cd5b2
```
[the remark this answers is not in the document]
howard-e: Agreed on these being tied to a `package.json` command as it's done semi frequently during testing.

server/scripts/import-tests/index.js
```diff
@@ -124,11 +130,20 @@ const importTestPlanVersions = async () => {
...
+        );
+        const isV2 = fse.existsSync(assertionsCsvPath);
+
         const tests = getTests({
             builtDirectoryPath,
             testPlanVersionId,
             ats,
-            gitSha
+            gitSha,
+            isV2
```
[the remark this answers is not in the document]
howard-e: Right, which is also why I chose to go with storing the number in `metadata` for now. But the `isV2` in code, makes it a little easier to process I think right now as we're only working with the 2 versions. My hope is making the change to support a possible v3 onwards, in the future _shouldn't_ be a huge lift.

Also, if a `v2+n` never comes, that small optimization may not be necessary.

server/scripts/import-tests/index.js
howard-e: Fair point. I can address this feedback.

server/tests/integration/testPlanRun.test.js
howard-e: I'll be adding v2 tests in a separate PR

server/scripts/import-tests/index.js
howard-e: @stalgiag I've pushed this change

stalgiag: Much easier to read. Thank you!

server/resolvers/helpers/retrieveCommands.js
```diff
@@ -0,0 +1,123 @@
+const commands = require('../../resources/commands.json');
```
alflennik, line 1: Would it be possible to rename commands to commandsV1? Otherwise it might give the impression that V1 is more authoritative or "normal" than V2 is, when V2 is actually the new standard going forward.

server/graphql-schema.js
```diff
@@ -396,7 +396,7 @@ const graphqlSchema = gql`
         """
         The AT mode the test was written to expect.
         """
-        atMode: AtMode!
+        atMode: AtMode
```
alflennik, line 399: I feel like having a MODELESS AtMode makes more sense than making this field nullable. Null to me implies more that the mode is missing or not relevant, and it would match the way we're handling modes in the database.

server/models/AtMode.js
```diff
@@ -21,14 +21,17 @@ module.exports = function (sequelize, DataTypes) {
             },
             name: {
                 type: DataTypes.TEXT,
-                // type: DataTypes.ENUM(
-                //     MODE.READING,
-                //     MODE.INTERACTION,
-                //     MODE.MODELESS
-                // ),
                 defaultValue: MODE.MODELESS,
```
alflennik, line 24: I thought I saw references somewhere that implied VoiceOver's quick nav was akin to a mode switcher, like I think I saw some "mode instructions" for VoiceOver that say to turn quick nav off ... but it sounds like we're still treating VoiceOver as modeless? I was wondering if you could bring me up to speed on what we decided about that?

alflennik (commented):
After spending quite a lot of time wondering what adding support for the V2 format would look like (and failing to figure out a clear direction I might add) it's quite interesting to see the move you found to get this done. I do have comments of course but first I should say how relieved and appreciative I am that you were able to figure out a way to accomplish this task, which certainly confounded me.

server/models/AtMode.js
howard-e: > I thought I saw references somewhere that implied VoiceOver's quick nav was akin to a mode switcher

I'm not too sure on that statement.

> I was wondering if you could bring me up to speed on what we decided about that?

VO can have a 'quick nav on' mode, 'quick nav off' mode and any AT could technically support a modeless state (but right now, only VO's v2 tests show any examples of that).

server/graphql-schema.js
howard-e: This was a missed usage opportunity. But returning `MODELESS` may not be logical for some tests. How the tests are in the v2 test format now is that they can have multiple states per test whereas v1 had a defined `atMode` attribute so this would have to be the following instead:
```diff
-atMode: AtMode!
+atModes: [AtMode]!
```

That was already being handled because of how collected.json passes along the assertions data but there is a reason for extending that here, for sure.

server/resolvers/helpers/retrieveCommands.js
howard-e: Yep, I can make that change!

server/migrations/20231114170315-addScreenTextAndInstructionsToAtMode.js
```diff
@@ -0,0 +1,34 @@
...
+            await queryInterface.removeColumn('AtMode', 'screenText', {
+                transaction
+            });
+
+            await queryInterface.removeColumn('AtMode', 'instructions', {
+                transaction
+            });
+        });
+    }
+};
```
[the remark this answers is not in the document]
howard-e: That's described by [v2 test format#settings](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#assertionstatement). And that "seeder" comes from each [at.settings under ats](https://github.com/w3c/aria-at/blob/master/tests/support.json).

server/scripts/import-tests/index.js
```diff
@@ -313,27 +394,219 @@ const getVersionString = async ({ directory, updatedAt }) => {
...
+                    title: collected.info.title,
+                    at: {
+                        key: collected.target.at.key,
+                        name: collected.target.at.name,
+                        settings: collected.target.at.raw.settings
+                    },
+                    atIds: [
+                        ats.find(at => at.name === collected.target.at.name).id
+                    ],
+                    renderableContent: {
```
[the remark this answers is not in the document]
howard-e: > I made a comment in one of your other PRs that I thought it made sense to add a testFormatVersion field to the renderableContent and then allowing the differences in the test format to be faithfully expressed within that section. What do you think?

To me, that then puts work on a resolver (or some other server op) to evaluate some change in the `renderableContent` based on that `testFormatVersion`. I'm not against it, but I am a bit lost, on why add to each `test.renderableContent` rather than referencing the parent `TestPlanVersion.metadata`.

> renderableContent was to faithfully preserve the collected.json file content

A big thing about this change was also maintaining the already existing structure of `renderableContent` to avoid unnecessary hiccups and maintaining that control. But the majority of `renderableContent` is the same as the collected file, and could have been `collected` or `{ ...collected, extras }` really. Outside of excluding `target.ats` which the new collected.json would have, and also making changes to assertions to use [tokenized assertionStatements](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#assertionstatement) where needed, it is exactly the same.

I can make a change to reflect that.

server/scripts/import-tests/index.js
```diff
@@ -15,6 +15,10 @@ const {
     getTestPlans,
```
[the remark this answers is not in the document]
howard-e: Is there a specific section to point out? The majority of the v1 vs v2 complexity is now included in `func:createTestsForFormat` following Stalgia's comment.

Some of the existing functions were renamed and to extend support of what they were originally doing. But I can add additional clarifying comments which attributes have been added.

The reason that binary logic can't happen further up is because of when the v2 evaluation for an example's folder has to happen.

server/scripts/import-tests/index.js
```diff
@@ -267,28 +284,92 @@ const readCsv = ({ sourceDirectoryPath }) => {
...
+        const commandsV2PathString = fse.readFileSync(commandsV2Path, 'utf8');
+        const commandsV2Parsed = JSON.parse(commandsV2PathString);
+
+        // Write commands for v2 format
+        await fse.writeFile(
+            path.resolve(__dirname, '../../resources/commandsV2.json'),
+            JSON.stringify(flattenObject(commandsV2Parsed), null, 4)
+        );
+    } catch (error) {
+        console.error("commands.json for v2 test format doesn't exist");
```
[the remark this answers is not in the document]
howard-e: > Maybe we can keep this message, rewording it slightly to be like "commands.json for v2 test format may not exist"

Agreed

> throw the error to make sure it's shown to developers

That means this block would also crash out when there's a manually imported commit before that v2 test format specific file existed.

server/scripts/import-tests/index.js
```diff
@@ -267,28 +284,92 @@ const readCsv = ({ sourceDirectoryPath }) => {
     };
 };
 
-const updateCommandsJson = async () => {
+const flattenObject = (obj, parentKey = '') => {
+    const flattened = {};
+
+    for (const key in obj) {
+        if (typeof obj[key] === 'object') {
```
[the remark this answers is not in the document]
howard-e: JS up to its ways again! That also means several instances of this pattern in aria-at should be updated.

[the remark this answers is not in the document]
howard-e: Addressed in https://github.com/w3c/aria-at-app/pull/840/commits/8dd8119e744e1dfc2a57b83a3c979ad5dd04a326

server/scripts/import-tests/index.js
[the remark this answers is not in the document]
howard-e: Addressed in https://github.com/w3c/aria-at-app/pull/840/commits/74a6f1abe26916c71d13f5fe98ca0c67dcbd6af2

server/resolvers/helpers/retrieveCommands.js
howard-e: So after looking into this, it does inflate the PR a bit. Like my comment in https://github.com/w3c/aria-at-app/pull/840#discussion_r1396413513, I'd rather to make this change in a follow up PR.

server/graphql-schema.js
howard-e: After an internal discussion and discussing some ways to best handle this, it makes more sense to just remove `atMode` from the `type Test` in graphql-schema.js. The field isn't being referenced on any pages and it's only been referenced when copy the value during the "test update" processes. When mode for a test is needed, it is already correctly referenced by the aria-at-harness from `renderableContent`.

Like my comment in https://github.com/w3c/aria-at-app/pull/840#discussion_r1396413513. I'd rather make this change in a follow up PR.

howard-e:
Follow up items this PR has revealed which can be addressed later on:

1. Update named references for `REQUIRED` and `OPTIONAL` to be `MUST` and `SHOULD`
2. Update the (current) `commands.json` to `commandsV1.json`
3. Remove references to `Test.atMode` from `graphql-schema.js`
4. Remove `TestPlanUpdater` component and related references

server/scripts/import-tests/index.js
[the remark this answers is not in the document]
alflennik: Ok, thanks for addressing that part of it! I'm cool with not throwing the error. But how about, instead, adding a second console.error with the actual error?

server/scripts/import-tests/index.js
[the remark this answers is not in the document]
alflennik: Cool thanks

server/graphql-schema.js
alflennik: I'm definitely happy with this outcome, nice idea Howard!

alflennik (approved):
Fantastic work getting this implemented Howard! Intense changes but great stuff.

Issue #743, opened 2023-08-15
Support V2 test format when manually running a test


mcking65:
#### Overview

Change the content shown in the test runner when a test plan written in the V2 format is opened.

Note:
* Unlike test plan viewing and reporting functions, It is not essential the system be able to support running both V1 and V2 formatted test plans. We could stop running test plans that are not in the V2 format on a planned transition date.
* Implementing the solution for this issue does not require w3c/aria-at#969 to be complete. This issue is a higher priority, although it might be more efficient to work on both in the same PR.
* A few changes specified below are not dependent on the V2 format; they are minor wording improvements that are simple to make while working on this issue.

In general, the changes needed in the runner are exact reflections of a subset of changes specified for the test plan preview generated during test plan build. Those changes are specified in w3c/aria-at#977.

##### Change heading 1

Streamline wording of the first level 1 heading. It is currently:

>Testing task:1. Navigate forwards to a collapsed disclosure button in reading mode

Replace the string "Testing task:1." with the string "Test 1:"

>Test 1: Navigate forwards to a collapsed disclosure button in reading mode

Note that the [V2 format](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2) contains a column for `presentationNumber` in tests.csv. That number specifies the sequence of test presentation.

##### Remove redundant heading 1 and subsequent paragraph

Following the AT and browser version information, there is another `h1` followed by a paragraph that describes the task. Remove the `h1` and the paragraph. Both are unnecessary.

Example:

>Testing task: Navigate forwards to a collapsed disclosure button in reading mode
>
>With JAWS in reading mode, describe how JAWS behaves when performing task "With the reading cursor on the 'Navigate forwards from here' link, navigate to the 'About' disclosure button."

##### Change instructions

Change the text of the level 2 heading from "Test Instructions" to "Instructions". Implement the instructions format as described in w3c/aria-at#977. Be sure to include the instructions for command settings where appropriate.

##### Remove success criteria

Remove the level 2 heading with text "Success Criteria" and the paragraph and list that follow it.

Example:

>Success Criteria
>
>To pass this test, JAWS needs to meet all the following assertions when each specified command is executed:
>
>1. Role 'navigation landmark' is conveyed 
>2. Name 'Mythical University' is conveyed 
>3. Role list is conveyed 
>4. Number of list items, '3', is conveyed 
>5. Role 'button' is conveyed 
>6. Name 'About' is conveyed 
[...]

PR #843, opened 2023-11-15
Update Test Runner for V2 Test Format

howard-e:
See #743. Based on #840.

This PR also updates `<InstructionsRenderer>` which was being used on the Candidate Test Plan Run page.

### DIFF
--- a/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
@@ -52,7 +52,12 @@ const NumberedList = styled.ol`
     }
 `;
 
-const InstructionsRenderer = ({ testResult, testPageUrl, at }) => {
+const InstructionsRenderer = ({
+    testResult,
+    testPageUrl,
+    at,
+    testFormatVersion = 1
+}) => {
     const { test = {} } = testResult;
     const { renderableContent } = test;
     const [testRunExport, setTestRunExport] = useState();
@@ -122,9 +127,7 @@ const InstructionsRenderer = ({ testResult, testPageUrl, at }) => {
     }, [testRunExport]);
 
     let allInstructions;
-    const isV2 =
-        !!renderableContent.target.at?.raw
-            ?.defaultConfigurationInstructionsHTML;
+    const isV2 = testFormatVersion === 2;
 
     if (isV2) {
         const commandSettingSpecified = renderableContent.commands.some(
@@ -185,7 +188,8 @@ InstructionsRenderer.propTypes = {
     testPageUrl: PropTypes.string,
     at: PropTypes.shape({
         name: PropTypes.string.isRequired
-    })
+    }),
+    testFormatVersion: PropTypes.number
 };
 
 export default InstructionsRenderer;
--- a/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
@@ -134,17 +134,23 @@ const InstructionsRenderer = ({
             ({ settings }) => settings && settings !== 'defaultMode'
         );
 
-        allInstructions = [
+        const defaultInstructions =
             renderableContent.target.at.raw
-                .defaultConfigurationInstructionsHTML,
-            `${supportJson.testPlanStrings.openExampleInstruction} ${renderableContent.target.setupScript.scriptDescription}.`,
-            `${renderableContent.instructions.instructions} ${
-                supportJson.testPlanStrings.commandListPreface
-            }${
-                commandSettingSpecified
-                    ? ` ${supportJson.testPlanStrings.commandListSettingsPreface}`
-                    : ''
-            }`
+                .defaultConfigurationInstructionsHTML;
+        const setupScriptDescription = `${supportJson.testPlanStrings.openExampleInstruction} ${renderableContent.target.setupScript.scriptDescription}`;
+        const testInstructions = renderableContent.instructions.instructions;
+        const settingsInstructions = `${
+            supportJson.testPlanStrings.commandListPreface
+        }${
+            commandSettingSpecified
+                ? ` ${supportJson.testPlanStrings.commandListSettingsPreface}`
+                : ''
+        }`;
+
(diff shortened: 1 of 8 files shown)


client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -125,104 +117,62 @@ const InstructionsRenderer = ({ testResult, testPageUrl, at }) => {
...
-    const allInstructions = [
-        ...instructionsContent.instructions.instructions.instructions,
-        ...instructionsContent.instructions.instructions.strongInstructions,
-        instructionsContent.instructions.instructions.commands.description
-    ];
-
-    const commands =
-        instructionsContent.instructions.instructions.commands.commands;
+    let allInstructions;
+    const isV2 =
```
howard-e, line 125: Updating to make use of `TestPlanVersion.metadata.testFormatVersion` instead

client/components/TestRun/index.jsx
```diff
@@ -1011,6 +1011,10 @@ const TestRun = () => {
                                               )
                                     }
                                     testPageUrl={testPlanVersion.testPageUrl}
+                                    testFormatVersion={
+                                        testPlanVersion.metadata
+                                            .testFormatVersion
```
alflennik, line 1016: Big ask here: can the testFormatVersion be included inside the renderableContent?

It's a pretty sizable change, but here's my reasoning about why it might be worth it. 

When contributors come across this variable, called testFormatVersion, the implication is that there can be multiple ways tests are formatted. It's right there in the name. So what's going to happen is that they are going to doubt whether they can rely on the documented concepts like "scenarios" and "assertions" and "tests" and "testResults."

As you know, in reality all these concepts are completely unaffected by the testFormatVersion. But it seems like they could be. That's the disconnect that I want to try to mitigate.

If the testFormatVersion was only accessible from within the renderableContent, A.K.A. renderableContent.testFormatVersion, suddenly it becomes immediately apparent where the safe zone ends and where the danger zone starts. It becomes obvious that if you are attempting to extract some kind of data from the renderableContent, you should make sure to account for all the versions that the data can be formatted.

There reason it is a big ask is that this field is not currently included in the renderableContent and I think it would need to be added in a migration. But furthermore, to do it right, it seems like going forward the field would need to be added into the build process of the ARIA-AT repo for both v1 and v2 build processes.

That's a lot, definitely reach out if I can clarify anything I said above.

alflennik (commented):
Generally the code looks great! I did leave one pretty lengthy comment as you can see, about the location where we are persisting the testFormatVersion variable. I'm hypothesizing that it would be a large scale change requiring a lot of effort (I might be wrong) so that's why I went into more detail about why I think it's worth it.

Note I haven't yet been able to test this locally but I'm working on it.

client/components/TestRenderer/utils.js
```diff
@@ -0,0 +1,104 @@
...
+    > li {
+        display: list-item;
+        list-style: circle;
+    }
+`;
+
+const NumberedList = styled.ol`
+    > li {
+    }
+`;
```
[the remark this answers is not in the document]
howard-e: Nope! Got caught copying and pasting 🙂

client/components/TestRun/index.jsx
howard-e: Answered by https://github.com/w3c/aria-at-app/pull/840#discussion_r1402645001, which is also where that change would happen.

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -125,104 +122,60 @@ const InstructionsRenderer = ({ testResult, testPageUrl, at }) => {
...
+                .defaultConfigurationInstructionsHTML,
+            `${supportJson.testPlanStrings.openExampleInstruction} ${renderableContent.target.setupScript.scriptDescription}.`,
+            `${renderableContent.instructions.instructions} ${
+                supportJson.testPlanStrings.commandListPreface
+            }${
+                commandSettingSpecified
+                    ? ` ${supportJson.testPlanStrings.commandListSettingsPreface}`
+                    : ''
+            }`
+        ].map(e => unescape(e));
```
stalgiag, line 148: I have a hard time seeing where an array item starts and stops here. This is mainly due to the fact that our project uses the default printwidth in prettier but maybe this could benefit from a couple of intermediate variables just to make the array easier to read.

client/components/TestRenderer/utils.js
```diff
@@ -0,0 +1,104 @@
...
+        if (typeof value === 'string') {
+            if (value === '.')
+                content = (
+                    <>
+                        {content}
+                        {value}
+                    </>
+                );
+            else
+                content = content = (
```
stalgiag, line 35:
```suggestion
                content = (
```

client/components/TestRenderer/utils.js
```diff
@@ -0,0 +1,104 @@
...
+                                </li>
+                            );
+                        })}
+                    </NumberedList>
+                </div>
+            );
+        }
+    });
+    return settingsContent;
+};
```
stalgiag, line 102: Glad to have these in a separate file!

stalgiag (approved):

stalgiag (approved):
I am still learning how the different parts of this app work with aria-at but I believethat I was able to get this up and view the changes without any issues.

Code feedback is minimal. Mostly everything looks good. All suggestions are optional so I am approving.

client/components/TestRenderer/index.jsx
```diff
@@ -216,6 +212,7 @@ const TestRenderer = ({
     at,
     testResult = {},
     testPageUrl,
+    testFormatVersion = 1,
```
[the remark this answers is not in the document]
howard-e: Done in https://github.com/w3c/aria-at-app/pull/843/commits/c29eef05fec95e6a84ffce3eea7f8552924fe64d

client/components/TestRenderer/utils.js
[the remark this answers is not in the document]
howard-e: Done in https://github.com/w3c/aria-at-app/pull/843/commits/eab5ec889979da7c1a7644c48c9f7e62822be250

alflennik (approved):

Issue #842, opened 2023-11-15
Data management page crashes when Adding Test Plans to the Test Queue  and Opening Assistive Technology select to pick any option


ccanash:
On the Data Management page, the app crashes when doing the following:

1. Navigate to /data-management
2. Expand Add Test Plans to the Test Queue
3. Open Assistive Technology select and pick any option

ccanash:
Hi @mcking65  this fix was included in dec 5th production release

PR #844, opened 2023-11-16
Fix Data management page crash when Adding Test Plans to the Test Queue and Opening Assistive Technology select to pick any option

stalgiag:
see #842 

This first commit in this PR fixes the bug. The other two commits remove the query and state for `browsers` in both `DataManagement` and `TestQueue`. Now that the `ManageTestQueue` component only uses browsers that are linked to an AT, the base `browsers` array is no longer needed as a prop. Since this was the only need for this array in the two parent components, the fetching and storing of that array is removed from the parents as well.

Note that the current design (both before and after this PR) does not allow for selecting a browser **before** selecting an AT. This seems fine from a usability standpoint but wanted to note it here.

### DIFF
--- a/client/components/DataManagement/index.jsx
+++ b/client/components/DataManagement/index.jsx
@@ -30,7 +30,6 @@ const DataManagement = () => {
 
     const [pageReady, setPageReady] = useState(false);
     const [ats, setAts] = useState([]);
-    const [browsers, setBrowsers] = useState([]);
     const [testPlans, setTestPlans] = useState([]);
     const [testPlanVersions, setTestPlanVersions] = useState([]);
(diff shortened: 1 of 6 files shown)


howard-e (approved):
@stalgiag thanks, LGTM!

> Note that the current design (both before and after this PR) does not allow for selecting a browser before selecting an AT. This seems fine from a usability standpoint but wanted to note it here.

This seems fine to me for now. At least, until there's a need to directly manipulate Browser related data from the app itself (in a similar way ATs are through the app)

alflennik (approved):
Code looks good, and I was able to verify that the crash is fixed, thanks Stalgia!

Issue #744, opened 2023-08-15
Support V2 test format when displaying reports


mcking65:
Assuming that #733 and #737 are complete, The [V2 test format](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2) drives only a few more changes to reports.

#### Change command heading text to include settings

In the implementation of #733, the command result tables are preceded by a heading that has text:

>COMMAND_NAME Results: X passed, Y failed

Example:

>Down Arrow then Down Arrow Results: 4 passed, 2 failed

As specified for the headings that precede assertion tables in w3c/aria-at#977, if settings were specified for a command, include the settings in parentheses. If there are multiple settings, separate them with the string " and ".

Example:

>Down Arrow then Down Arrow (browse mode on) Results: 4 passed, 2 failed

#### Command results table content changes

Update the table caption to match heading content.

In the assertion column, render the phrase form of the assertion specified in the `assertionPhrase` column of `assertions.csv`.

Change row sorting to:
1. priority descending: MUST then SHOULD then MAY
2. Within each priority group, sort with failures first then passes.
3. Within fail/pass groups, sort in assertion order specified for the test in the assertions column of `tests.csv`.
    
#### Changes to heading text on report pages

In the implementation of #733 on [report pages](https://aria-at.w3.org/report/1890/targets/156#result-YTg2ZeyIxMiI6MzYxfQTFiMG) where the set of command results tables for a test is preceded by a level 2 heading with text:

>TEST_NAME (X passed, y failed)

Example:

>Trigger an alert (2 passed, 2 failed)

Add the string "Test TEST_NUMBER: " to the beginning of the heading text. For example:

>Test 1: Trigger an alert (2 passed, 2 failed)

#### Mockup

Here is a zip of an HTML file containing a mockup for VoiceOver results of 1 test:
[TestResultsTableMockupForIssue744.zip](https://github.com/w3c/aria-at-app/files/12355770/TestResultsTableMockupForIssue744.zip)

mcking65:
I will be adding implementing option B from w3c/aria-at#999 into this issue as discussed in CG meeting.

lolaodelola:
@ccanash (& relevant internal engineer) for internal review, this has been updated to the following format:

> Option B
> Add a new level 3 heading:
> 
> Results for each command
> 
> Then Change the level of the heading for each command to 4, keeping the same text as today:
> 
> B Results: 2 passed, 0 failed
> 
[...]

PR #845, opened 2023-11-16
Update Reports for V2 Test Format

howard-e:
See #744. Based on #840.

This PR does the following:
> Add the string "Test TEST_NUMBER: " to the beginning of the heading text.

The other items seem to already be addressed.

### DIFF
--- a/client/components/common/TestPlanResultsTable/index.jsx
+++ b/client/components/common/TestPlanResultsTable/index.jsx
@@ -21,7 +21,7 @@ const TestPlanResultsTable = ({
     optionalHeader = null,
     commandHeadingLevel = 3
 }) => {
-    const Heading = `h${commandHeadingLevel}`;
+    const CommandHeading = `h${commandHeadingLevel}`;
 
     return (
         <>
@@ -65,11 +65,11 @@ const TestPlanResultsTable = ({
 
                 return (
                     <React.Fragment key={scenarioResult.id}>
-                        <Heading>
+                        <CommandHeading>
                             {commandsString}&nbsp;Results:&nbsp;
                             {passedAssertions.length} passed,&nbsp;
                             {failedAssertions.length} failed
-                        </Heading>
+                        </CommandHeading>
                         <p className="test-plan-results-response-p">
                             {test.at?.name} Response:
                         </p>
--- a/client/components/common/TestPlanResultsTable/index.jsx
+++ b/client/components/common/TestPlanResultsTable/index.jsx
@@ -18,12 +18,14 @@ const TestPlanResultsTable = ({
     test,
     testResult,
     tableClassName = '',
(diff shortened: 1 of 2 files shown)


client/components/common/TestPlanResultsTable/index.jsx
```diff
@@ -14,9 +14,16 @@ const renderAssertionRow = (assertionResult, priorityString) => {
...
-const TestPlanResultsTable = ({ test, testResult, tableClassName = '' }) => {
+const TestPlanResultsTable = ({
+    test,
+    testResult,
+    tableClassName = '',
+    isReportsPage = false
+}) => {
     return (
         <>
+            {isReportsPage ? <h3>Results for each command</h3> : null}
```
alflennik, line 25: If this headline is not always needed maybe it's a sign it doesn't belong in this component and can be handled by the parent component.

client/components/common/TestPlanResultsTable/index.jsx
```diff
@@ -56,11 +63,19 @@ const TestPlanResultsTable = ({ test, testResult, tableClassName = '' }) => {
 
                 return (
                     <React.Fragment key={scenarioResult.id}>
-                        <h3>
-                            {commandsString}&nbsp;Results:&nbsp;
-                            {passedAssertions.length} passed,&nbsp;
-                            {failedAssertions.length} failed
-                        </h3>
+                        {isReportsPage ? (
```
alflennik, line 66: Instead of changing the behavior based on an `isReportsPage` prop, which is difficult to understand without delving in here, maybe we can add a `headingLevel={3}` prop. It would work like this ...

```js
const Header = `h${headingLevel}`

return (
  // ...
  <Header>
    Header content
  </Header>
)
```

alflennik (commented):
I left a couple of comments but generally looks good. I'm currently working out a way to run this code so stand by for any feedback that comes from when I do that.

client/components/common/TestPlanResultsTable/index.jsx
howard-e: Good point, addressed in https://github.com/w3c/aria-at-app/pull/845/commits/9eb58751f7aabdd88fe8aaf53d1afe21d67d7b7f.

client/components/common/TestPlanResultsTable/index.jsx
howard-e: Thanks for that, been seeing this sprinkled around! We could make general component for this pattern now. Also updated in https://github.com/w3c/aria-at-app/pull/845/commits/9eb58751f7aabdd88fe8aaf53d1afe21d67d7b7f.

stalgiag (approved):
Able to test locally with no issues. The code looks good to me. Thanks @howard-e !

Issue #788, opened 2023-09-23
add test plan status and app navigation banner to test plan review page


mcking65:
On the test plan review page, example screenshot below, it is important that readers can determine the current working mode status of the test plan and can readily discern that this page is part of the aria-at app. They need to be able to get from this page to the version history for the test plan as well as other pages in the app. These requirements are particularly important for this page because it is a page that may be shared in contexts outside the app, e.g., issues or email.

Specific requirements are provided below. Note that they use language that is a slightly more verbose than in some other pages. The language is intended to help give people who have little or no familiarity with the ARIA-AT working mode some idea of the current status and history of this test plan. Like the reports, this page is one of the pages in the app that may be shared with people who are not directly involved with the project and are looking at this information with a singular purpose, e.g., an AT developer assigned to a bug, an AX specialist asked to learn about expected screen reader behaviors for a specific widget.

![Screenshot of color viewer slider test plan review page](https://github.com/w3c/aria-at-app/assets/7241781/8a23c773-c404-40f3-b6b2-ed630e75475a)

#### Add aria-at banner

Add the site-wide banner that includes top-level navigation to the top of the page.

#### Add link to version history page

Under the H2 and before the commit summary, add a link named "View all versions of this test plan".

Present this link as the first item in an unordered list. Subsequent list items are described below.

#### Add current test plan working mode progress and status

Add elements to the above mentioned list that show the current and previously completed working mode phases for this plan using the following text. 

* If current phase is R&D Complete, add list item with string "This version is a proposed  draft; ARIA-AT review processes have not yet started."
* If draft review has started:
     * Replace prior list item text with "R&D complete on COMPLETION_DATE for this version."
    * Add list item with text "ARIA-AT draft review process started on START_DATE for this version."
* If candidate review has started, add a list item with text "ARIA-AT candidate review process started on START_DATE for this version."
* If the version is recommended, add element with string "ARIA-AT Community Group approved this version as recommended on RECOMMENDED_DATE."
* If version is deprecated, add element with string "This version was deprecated on DEPRECATION_DATE. Most recent version is VERSION_STRING_LINK." where VERSION_STRING_LINK is a link to the test plan review page for the most recent active version.

#### Emphasize deprecated status in heading

If the version is deprecated, add "Deprecated " to the start of the H2 before the word "Version" and visually style the word and/or heading to draw attention to the fact that this version of the plan is currently deprecated.

#### Example

[...]

PR #846, opened 2023-11-20
Implement Test Review in React

alflennik:
Replaces the existing Test Review page, which is server-rendered with Handlebars and basically a direct port from a page on the ARIA-AT repo, with a client-rendered page in React, using the existing GraphQL endpoints to populate its data. Using existing React infrastructure allows us to resolve some limitations like being unable to put the normal nav bar on the page (see https://github.com/w3c/aria-at-app/issues/788, which is fixed by this PR) and it also allows for a more compact implementation that's more similar to other pages which display test content.

Regarding #788, while bringing the page into React, its structure changed enough that it wouldn't make sense to implement the requested changes exactly as requested, but I do think all the requested information has been added in an equivalent form.


client/components/TestReview/index.jsx
```diff
@@ -0,0 +1,270 @@
...
+                                    const isLast =
+                                        index === specifications.length - 1;
+                                    return (
+                                        <Fragment key={title}>
+                                            <a
+                                                href={link}
+                                                rel="noreferrer"
+                                                target="_blank"
+                                            >
+                                                {title}
```
howard-e, line 234: This produces link text such as `example`, `aria-live`, etc which is different from what we have now being `APG example: <pattern>.html`, `ARIA specification: aria-live`, etc.

I think this is fine for now given how the app imports those tests, but just calling out that this will be extended in the future to [support clearly defined link text for v2 test format tests](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#referencescsv).

There is no request for a change here, just a note.

client/components/TestReview/index.jsx
```diff
@@ -0,0 +1,270 @@
...
+                    {`${testPlanVersion.versionString} `}
+                    <Link
+                        to={`/data-management/${testPlanVersion.testPlan.directory}`}
+                    >
+                        (View all versions of this test plan)
+                    </Link>
+                </li>
+                <li>
+                    <strong>Version History:&nbsp;</strong>
+                    <Ul>
```
howard-e, line 125: This doesn't cover the following condition from #788:

> If current phase is R&D Complete, add list item with string "This version is a proposed draft; ARIA-AT review processes have not yet started."

client/components/TestReview/index.jsx
```diff
@@ -0,0 +1,270 @@
...
+            <h1>
+                {`Tests for ${testPlanVersion.title} ` +
+                    `${testPlanVersion.versionString}` +
+                    `${testPlanVersion.deprecatedAt ? ' (Deprecated)' : ''}`}
+            </h1>
+            <h2>Introduction</h2>
+            <p>
+                {`This page contains the full content of all ${testCount} ` +
+                    `tests the ARIA-AT team has authored for ` +
+                    `${testPlanVersion.title} as of version ` +
```
howard-e, line 105:
```suggestion
                    `${testPlanVersion.title}, version ` +
```

This read to me like newly added tests, in some future `VYY.MM.DD` would and could also show up for the pattern, on the 'current' version's page.

client/components/TestReview/index.jsx
```diff
@@ -0,0 +1,270 @@
...
+                <title>{`Tests for ${testPlanVersion.title} ${testPlanVersion.versionString} | ARIA-AT`}</title>
+            </Helmet>
+            <h1>
+                {`Tests for ${testPlanVersion.title} ` +
+                    `${testPlanVersion.versionString}` +
+                    `${testPlanVersion.deprecatedAt ? ' (Deprecated)' : ''}`}
+            </h1>
+            <h2>Introduction</h2>
+            <p>
+                {`This page contains the full content of all ${testCount} ` +
```
howard-e, line 103:
```suggestion
                {`This page contains ${testCount} ` +
```

howard-e (changes requested):
@alflennik looks good! There was a lot of "cleanup" here in terms of references and code and it seems you've gotten to it all.

I've left some inline comments on minor (I think) changes.

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -202,8 +193,10 @@ const InstructionsRenderer = ({ testResult, testPageUrl, at }) => {
 
     const assertionsContent = parseListContent(assertions);
 
+    const Heading = `h${headingLevel}`;
```
stalgiag, line 196: I didn't know that you could create HTML elements with strings in JSX like this. Very neat.

client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -553,7 +553,7 @@ const DataManagementRow = ({
                         <VersionString
                             role="listitem"
                             iconColor="#2BA51C"
-                            linkHref={`/test-review/${latestVersion.gitSha}/${latestVersion.testPlan.directory}`}
+                            linkHref={`/test-review/${latestVersion.id}`}
```
stalgiag, line 556: It seems like the app isn't so heavily used or linked to that this will matter but just putting a note here to make sure that we all agree that breaking previous sha-based routing is okay.

client/components/Reports/SummarizeTestPlanReport.jsx
```diff
@@ -151,11 +151,7 @@ const SummarizeTestPlanReport = ({ testPlanVersion, testPlanReports }) => {
             <ul>
                 <li>
                     Generated from&nbsp;
-                    <a
-                        href={`/test-review/${testPlanVersion.gitSha}/${testPlanVersion.testPlan.directory}`}
-                        target="_blank"
-                        rel="noreferrer"
-                    >
+                    <a href={`/test-review/${testPlanVersion.id}`}>
```
stalgiag, line 154: Nice. I definitely agree with not opening the new tab.

client/components/DataManagement/DataManagementRow/index.jsx
stalgiag: There is something nice about having the `testPlan.directory`in the URL. It feels more human-readable. Not requesting a change but more a point of discussion.

client/components/TestReview/index.jsx
```diff
@@ -0,0 +1,270 @@
...
+                                    return (
+                                        <Fragment key={title}>
+                                            <a
+                                                href={link}
+                                                rel="noreferrer"
+                                                target="_blank"
+                                            >
+                                                {title}
+                                            </a>
+                                            {isLast ? '' : ', '}
```
stalgiag, line 236: I understand that this is more visually compact by just making them into sibling <a>'s but the tradeoff is that the comma confuses VoiceOver (other ATs untested). I think it is worth the added scroll height to have a <ul> -> <li> -> <a> structure without commas here.

client/components/TestReview/index.jsx
```diff
@@ -0,0 +1,270 @@
...
+                            </li>
+                        )}
+                    </Ul>
+                </li>
+                <li>
+                    <strong>Commit:&nbsp;</strong>
+                    {testPlanVersion.gitMessage}
+                </li>
+            </ul>
+            <h2>List of Tests</h2>
```
stalgiag, line 177: This component is pretty lengthy. I feel like the filtered tests could become their own component fairly easy. This would shorten this component and it seems likely that it could be reused elsewhere at some point since this section has a lot of standalone use value.

client/components/common/FilterButtons/index.jsx
```diff
@@ -57,8 +57,9 @@ const FilterButtons = ({
 };
 
 FilterButtons.propTypes = {
+    filterLabel: PropTypes.string.isRequired,
+    filterAriaLabel: PropTypes.string,
```
stalgiag, line 61: Nice updates to this component!

stalgiag (approved):
Looks good overall. I appreciate the attention to detail with a change that touches so many parts of the application. I compared this page to the previous page and there is a significant change in the text content present. There are also minor changes in the experience of navigating with a screenreader. I don't have the history with the project to know how proactive changes like this are received but thought it worth noting.

I am going to approve as all of my inline suggestions are optional.

stalgiag (approved):
Edit: My code review tool double posted for some reason. Apologies!

client/components/TestReview/index.jsx
alflennik: Ah right, I actually noticed a contradiction in between the spec and the example so I decided to go with the text the example gave, which is "R&D completed on Apr 14, 2022 for this version." Do you think that's ok?

client/components/TestReview/index.jsx
alflennik: Sure this is fine!

client/components/TestReview/index.jsx
alflennik: Ok sure

client/components/DataManagement/DataManagementRow/index.jsx
alflennik: I do feel like we still have enough leeway to change the URLs, I really doubt there are any bookmarks out there, but I appreciate you mentioning that!

For the URLs, my main motivation was to avoid using the whole git sha which made the URL feel unwieldy, but another option could be `/test-review/checkbox-two-state/v23.08.23`. This would require adding graphql support to query by versionString so it's a bit more work, but if you like it better maybe it's worth it?

That said, we generally use the TestPlanVersion ids elsewhere so this URL is pretty in keeping with our other URLs.

client/components/TestReview/index.jsx
alflennik: Sure, good idea

client/components/DataManagement/DataManagementRow/index.jsx
stalgiag: That would definitely be the most human-readable URL but even just using the id with the test plan directory would add some info without adding a new query, like `/test-review/checkbox-two-state/5`. I tend to be biased towards putting as much human-readable information in the URL as possible but I definitely do not feel strongly about this. All of the options, including keeping as-is, work for me.

client/components/TestReview/index.jsx
alflennik: I think this is a change that probably makes more sense to implement when this content is needed somewhere else, because I think for now the connective tissue needed like prop types might undercut some of the space saving benefits, and the component interface might be negatively affected by the need to support filtering out irrelevant ATs.

Issue #736, opened 2023-08-04
Support pass/fail verdicts in algorithms that compare reports to determine which tests have equivalent results


mcking65:
This issue is one element of the plan documented in w3c/aria-at#961 to migrate from 3 assertion verdicts (good/no output/incorrect output) to two verdicts (pass/fail).

ccanash:
Hi @mcking65  this was included in dec 5th release. Can it be closed?

PR #847, opened 2023-11-21
Passed/failed for assertion results in TestRenderer and migrate to pass/fail for report comparison algorithms

stalgiag:
see #736 

This PR does the following:
- Changes input elements for test runs to a fieldset with checkbox inputs.
- Updates test rendered component to use correct input change callbacks on updated assertion data structure
- Updates result comparison algorithm so that it no longer takes `failedReason` into account when calculating conflicts

For reviewers, the `client/resources` changes are not relevant to this PR and have been reviewed elsewhere.

### DIFF
--- a/client/components/TestRenderer/AssertionsFieldset/index.jsx
+++ b/client/components/TestRenderer/AssertionsFieldset/index.jsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Fieldset } from '..';
+import styled from '@emotion/styled';
+
+const Label = styled.label`
+    display: block;
+`;
+
+const AssertionsFieldset = ({ assertions, commandIndex }) => {
+    return (
+        <Fieldset>
+            <legend>Assertions</legend>
+            {assertions.map((assertion, assertionIndex) => {
+                const { description, passed, click } = assertion;
+
+                return (
+                    <Label key={`AssertionKey_${assertionIndex}`}>
+                        <input
+                            type="checkbox"
+                            id={`pass-${commandIndex}-${assertionIndex}`}
+                            name={`assertion-${commandIndex}-${assertionIndex}`}
+                            defaultChecked={passed}
+                            onClick={click}
+                        />
+                        {description[0]}
+                    </Label>
+                );
+            })}
+        </Fieldset>
+    );
+};
+
+AssertionsFieldset.propTypes = {
+    assertions: PropTypes.array.isRequired,
+    commandIndex: PropTypes.number.isRequired
+};
+
+export default AssertionsFieldset;
--- a/client/components/TestRenderer/AssertionsFieldset/index.jsx
+++ b/client/components/TestRenderer/AssertionsFieldset/index.jsx
@@ -1,16 +1,31 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { Fieldset } from '..';
 import styled from '@emotion/styled';
+import supportJson from '../../../resources/support.json';
 
 const Label = styled.label`
     display: block;
+
+    input {
+        margin-right: 0.25rem;
+    }
 `;
 
-const AssertionsFieldset = ({ assertions, commandIndex }) => {
+const AssertionsFieldset = ({ assertions, commandIndex, assertionsHeader }) => {
+    // Handle case where build process didn't include assertionResponseQuestion
+    const normalizedHeader = useMemo(() => {
+        return assertionsHeader?.descriptionHeader?.replace(
+            'undefined',
+            supportJson.testPlanStrings.assertionResponseQuestion
+        );
+    }, [assertionsHeader]);
+
     return (
         <Fieldset>
-            <legend>Assertions</legend>
+            <legend id={`command-${commandIndex}-assertions-heading`}>
+                {normalizedHeader}
+            </legend>
             {assertions.map((assertion, assertionIndex) => {
                 const { description, passed, click } = assertion;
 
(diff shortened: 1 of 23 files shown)


client/components/TestRenderer/AssertionsFieldset/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+    display: block;
+
+    input {
+        margin-right: 0.25rem;
+    }
+`;
+
+const AssertionsFieldset = ({ assertions, commandIndex, assertionsHeader }) => {
+    // Handle case where build process didn't include assertionResponseQuestion
+    const normalizedHeader = useMemo(() => {
```
stalgiag, line 17: The value from `support.json` wasn't being correctly included with test commands in `assertionsHeader.descriptionHeader`. I am not sure but my sense is that this isn't from changes in this PR but instead comes from the sequence of updates to the test data in current development. If this isn't the case, let me know. Either way this calculation handles the issue and it can become a tiny tech debt task after all of the v2 changes have landed.

howard-e: The renderer in this app has to be built on top of the individual [collected json files](https://github.com/w3c/aria-at/blob/e617ea452142f92446792ca19d299d9fa99d740a/tests/resources/aria-at-harness.mjs#L116:L130). Architecturally, that means the .collected.json for any of these tests have to include the relevant data. Looking into this, it's clear there is some extending that can happen here to support that.

howard-e: I've submitted https://github.com/w3c/aria-at/pull/1005 as a potential fix. Let me know your thoughts.

There's other potential tech debt I've noticed, that may come up during future v2 format support work, that I think we can handle as they arise.

server/resolvers/TestResultOperations/saveTestResultCommon.js
```diff
@@ -131,12 +131,8 @@ const assertTestResultIsValid = newTestResult => {
...
     const checkAssertionResult = assertionResult => {
         if (
-            !(
-                (assertionResult.passed === true &&
-                    !assertionResult.failedReason) ||
-                (assertionResult.passed === false &&
-                    !!assertionResult.failedReason)
-            )
+            assertionResult.passed === null ||
+            assertionResult.passed === undefined
```
howard-e, line 135:
```suggestion
            !assertionResult.passed
```

client/components/TestRenderer/AssertionsFieldset/index.jsx
howard-e: Looking again at how this is used, I'm also not against just explicitly depending on the `support.json` in this use case. It should always be up to date because of the import script.

howard-e (commented):
These changes look good to me. Although I imagine in the future, we'll want to clean up references to `failedReason`.

Also I think these [references to failedReason being updated in the harness' state](https://github.com/w3c/aria-at-app/blob/77419d790218e27ea7f4bfa48fe024d66d59cd83/client/components/TestRenderer/index.jsx#L267:L273) may need to be updated to be safe? I imagine it's something like the following based on your other changes:
```diff
                 if (assertionResult.passed)
                     commands[i].assertions[j].result = 'pass';
-                else if (assertionResult.failedReason === 'NO_OUTPUT')
-                    commands[i].assertions[j].result = 'failMissing';
-                else if (assertionResult.failedReason === 'INCORRECT_OUTPUT')
-                    commands[i].assertions[j].result = 'failIncorrect';
-                else commands[i].assertions[j].result = 'notSet';
+                else commands[i].assertions[j].result = 'fail';
```

server/resolvers/TestResultOperations/saveTestResultCommon.js
stalgiag: This looked good to me so I committed it but then the tests were failing. This made me realize that I can't do this because `assertionResult.passed === false` is a valid assertion result. Very easy to get tripped up on that 😅

stalgiag:
> These changes look good to me. Although I imagine in the future, we'll want to clean up references to `failedReason`.
> 
> Also I think these [references to failedReason being updated in the harness' state](https://github.com/w3c/aria-at-app/blob/77419d790218e27ea7f4bfa48fe024d66d59cd83/client/components/TestRenderer/index.jsx#L267:L273) may need to be updated to be safe? I imagine it's something like the following based on your other changes:
> 
> ```diff
>                  if (assertionResult.passed)
>                      commands[i].assertions[j].result = 'pass';
> -                else if (assertionResult.failedReason === 'NO_OUTPUT')
> -                    commands[i].assertions[j].result = 'failMissing';
> -                else if (assertionResult.failedReason === 'INCORRECT_OUTPUT')
> -                    commands[i].assertions[j].result = 'failIncorrect';
> -                else commands[i].assertions[j].result = 'notSet';
> +                else commands[i].assertions[j].result = 'fail';
> ```

Thanks for this! I forgot that this value is not just excluded from v2, it is being deprecated from use across all test versions. This allowed to more destructively remove the references. I only left them in two locations:
1. GraphQL schema, just as documentation to explain the DB
2. `populateTestData` functions. Perhaps these will need to be removed. I was unsure of how much we want to rewrite  our base local testing state.

This also means that there is some additional clean up that can happen upstream in the resources files in aria-at. I'll handle that next week.

client/components/TestRenderer/AssertionsFieldset/index.jsx
stalgiag: Let's leave as-is for now and I'll check out #1005 for sweeping this up later. Definitely would be nice to avoid the weirdness of looking for the undefined value here.

stalgiag: Circling back around - It doesn't seem that `collectedText.supportJSON` is coming through with the changes in #1005. I can look more into this next week if we think it is important to handle now.

PR #851, opened 2023-12-04
Trap backwards navigation from initial focus on header in focus-trapped modals

stalgiag:
see #805. 

Previously, if one were to SHIFT+TAB immediately upon opening a focus-trapped `BasicModal`, one would inadvertently escape the focus trapping. This was due to the fact that the underlying Bootstrap modal component places focus on an a header upon first open but that header is not part of the set of wrapping focusable elements. Fixed in this branch and add a unit test to prevent this problem in the future.

tested on Chrome 119.

Note: there is a slightly mangled history. Apologies for that. I will squash before merging.

### DIFF
--- a/client/components/common/BasicModal/index.jsx
+++ b/client/components/common/BasicModal/index.jsx
@@ -48,7 +48,13 @@ const BasicModal = ({
     }, []);
 
     return (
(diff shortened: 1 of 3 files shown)


Paul-Clue (approved):
Works perfectly!

PR #852, opened 2023-12-05
Do not render zeros in place of omitted content

jugglinmike:
By relying on the "truthiness" of Number values to short-circuit boolean expressions, some rendering logic produced the text "0" instead of omitting any markup at all.

Replace the expressions whose value could be coerced to booleans with expressions which directly evaluate to "fallback" values that do not produce markup (i.e. `false` or an empty array).


howard-e (approved):

howard-e:
Thanks for catching that!

jugglinmike:
Sure thing!

PR #854, opened 2023-12-05
Correct heuristic for identifying "finished" runs

jugglinmike:
Prior to this commit, the user interface reported that Test Plan Runs were "finished" when they included zero test results. This state, though difficult to observe in development environments, will be apparent whenever there is a perceivable delay while scheduling collection jobs (including when collection jobs are never scheduled due to some unexpected error). Test Admins were unable to manage Test Plan Runs from this state.

Update the heuristic to identify Test Plan runs in this state as "unfinished."


stalgiag (approved):
Good catch and thanks for the fix!

jugglinmike:
Thanks for the review!

PR #857, opened 2023-12-06
Store "No output was detected." for empty string output during automated tests

stalgiag:
This PR updates the automation controller to store "No output was detected." for any output submitted as an empty string by an automated run. This brings the behavior inline with human submitted test results that have the "No output" checkbox checked.

### DIFF
--- a/server/controllers/AutomationController.js
+++ b/server/controllers/AutomationController.js
@@ -256,7 +256,11 @@ const updateJobResults = async (req, res) => {
 // Human test runners are able to use a checkbox to indicate no output was detected.
 // This checkbox stores 'No output was detected.' as the output value for that scenarioResult.
 const convertEmptyStringsToNoOutputMessages = outputs =>
-    outputs.map(output => (output === '' ? 'No output was detected.' : output));
+    outputs.map(output =>
+        output === null || output.trim() === ''
+            ? 'No output was detected.'
+            : output
+    );
 
 module.exports = {
     cancelJob,
--- a/server/controllers/AutomationController.js
+++ b/server/controllers/AutomationController.js
@@ -239,10 +239,12 @@ const updateJobResults = async (req, res) => {
         })
     ]);
 
+    const processedResponses = convertEmptyStringsToNoOutputMessages(responses);
+
     await updateOrCreateTestResultWithResponses({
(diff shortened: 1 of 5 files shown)


server/controllers/AutomationController.js
```diff
@@ -251,6 +253,11 @@ const updateJobResults = async (req, res) => {
     res.json({ success: true });
 };
 
+// Human test runners are able to use a checkbox to indicate no output was detected.
+// This checkbox stores 'No output was detected.' as the output value for that scenarioResult.
+const convertEmptyStringsToNoOutputMessages = outputs =>
+    outputs.map(output => (output === '' ? 'No output was detected.' : output));
```
gnarf, line 259: A little bit worried about `null` or `' '`, not enough to push for handling it now, but, just wondering what "right" answer for those two options are.

gnarf (approved):
I left a comment that doesn't need to be addressed for this to land, but we might want to make sure we know what we want in those extra cases.

server/controllers/AutomationController.js
stalgiag: Oh good point. I was really thinking about what the value would be with a human test run. I think we should handle null and only white space cases also. This will put the responsibility of distinguishing these cases during debugging on the sending side exclusively but it will look better from the app which seems important right now.

stalgiag: Updated!

stalgiag:
failing tests should be addressed by #858

stalgiag:
> Given that different parts of the code base rely on this particular value (and given my understanding that it is difficult to express that dependency in code), I'd like to define `'No output was detected.'` as a constant, and annotate the value both here and [in the front-end code](https://github.com/w3c/aria-at-app/blob/457caea7b3e3f559923a6e64546cf6fc9e018633/client/components/TestRenderer/OutputTextArea/constants.js#L1).
> 
> To be clear, it's not so much that I think anyone is going to change one of these willy-nilly. I just hope that we eventually move toward representing this case more formally in the data model, and when that happens, I want folks working on it to recognize the seemingly-unrelated places that will need updating.

Agreed and addressed!

jugglinmike (approved):
Very nice!

PR #860, opened 2023-12-07
Show tests as queued even before the job starts running, BotRunTestStatusList

stalgiag:
see title.

Previously the BotRunTestStatusList would not show tests as "queued" until the test runner was running the test. This changes makes it so that tests show as queued immediately upon initialization.


jugglinmike (approved):
Works like a charm

PR #863, opened 2023-12-07
Support 0-level assertion priorities on TestRun page

howard-e:
See https://github.com/w3c/aria-at/issues/1008.

**How To Run:**
```
yarn db-import-tests:dev;
yarn run dev-debug;
```

Note 1: This work is dependent on https://github.com/w3c/aria-at/pull/1012 landing and any updated `client/resources` files here should be reviewed there.

Note 2: This is using a sample example data made available by [w3c/aria-at branch collection-0-priority-support-test](https://github.com/w3c/aria-at/tree/collection-0-priority-support-test). You can view the specifics of the change in the description of https://github.com/w3c/aria-at/pull/1014.

___

Set `ariaAtDefaultBranch` in `import-tests/index.js` to `collection-0-priority-support-test` for testing purposes.

**NOTE: REVERT TEST BRANCH BEING IMPORTED IN `import-tests/index.js` BEFORE MERGING**


server/scripts/import-tests/index.js
```diff
@@ -53,7 +53,8 @@ Default use:
 const client = new Client();
 
 const ariaAtRepo = 'https://github.com/w3c/aria-at.git';
-const ariaAtDefaultBranch = 'master';
+// const ariaAtDefaultBranch = 'master';
```
howard-e, line 56: This needs to be reverted before merging.

client/components/common/queries.js
```diff
@@ -0,0 +1,183 @@
+import { gql } from '@apollo/client';
```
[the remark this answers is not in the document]
alflennik: I'm down with this. Most fields don't impact query performance so it's totally fine to just query all of them every time. For fields that do impact performance, like tests on TestPlanVersions and testResults on TestPlanRuns, I think it should be divided into two fragments to make accidentally tanking your performance less likely to happen. Something like `TEST_PLAN_RUN_FIELDS` and `TEST_PLAN_RUN_TEST_RESULTS_FIELDS`.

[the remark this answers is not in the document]
alflennik: I think fields like renderableContent might be preferable to leave off, since they're extremely verbose and "specialty" for specific use cases. If it's there where it's not needed it will definitely make debugging responses harder.

server/graphql-schema.js
```diff
@@ -689,6 +695,11 @@ const graphqlSchema = gql`
         results require this field to be filled in.
         """
         passed: Boolean
+        """
+        Indicates if this assertion result needs to be ignored because it is included with
+        a 0-level priority
+        """
+        exclude: Boolean
```
alflennik, line 702: If some assertions are not relevant, I don't think we should require users of the API to be responsible for detecting that and requiring them to pass in a special assertionResult with an "exclude" flag set to true. It feels like an example where complexity is escaping the abstraction. API consumers should only have to provide assertionResults which are relevant to the testResult.

server/graphql-schema.js
```diff
@@ -492,6 +492,12 @@ const graphqlSchema = gql`
         A human-readable version of the command, such as "Control+Alt+Down"
         """
         text: String!
+        """
+        The AT mode this command may be getting ran in, such as "quickNavOn".
+        The same command can be ran during the same test, but in a different
+        mode
+        """
+        settings: String
```
alflennik, line 500: It is possible to make this an enum?

server/scripts/import-tests/index.js
alflennik: Nice trick!

alflennik (changes requested):
Great work! I like what you did with the GraphQL fragments, you were able to delete a surprising amount of code, and that's definitely a good thing.

I feel quite strongly that we should not add a concept of inert assertionResults that API consumers must disable with an exclude flag. If an assertionResult is not relevant I feel like API consumers shouldn't need to provide anything or worry about it at all.

As for an alternate way to do it, one suggestion is that we add the ability to filter test assertions by mode.

It might be more involved to implement up front but I think it will be worth it make sure our abstractions are robust enough to shoulder the weight of all this complexity.

howard-e:
> I feel quite strongly that we should not add a concept of inert assertionResults that API consumers must disable with an exclude flag. If an assertionResult is not relevant I feel like API consumers shouldn't need to provide anything or worry about it at all.

@alflennik agreed! I had the same concern when initially making this PR and was hoping to resolve it in a follow up task because of my assumption that further changes to the harness and TestRenderer would be required, which would conflict with already floating changes in that bit of work that would have been too mentally hard to track. Taking another look at it now (and with fresh thoughts 🙂), turns out that wasn't needed. https://github.com/w3c/aria-at-app/pull/863/commits/fe222f796e90e89d55bcf90c609bff3f566a0e22 should address this.

server/graphql-schema.js
howard-e: Addressed in https://github.com/w3c/aria-at-app/pull/863/commits/fe222f796e90e89d55bcf90c609bff3f566a0e22

server/graphql-schema.js
howard-e: The values here won't be static. If the `support.json` is updated with a new AT + settings (or just a new setting in general for a currently imported AT), and the auto-importer triggers, it will fail to query the relevant content if this is an enum of the `settings` values we know of now.

server/graphql-schema.js
alflennik: Looks good thanks for addressing that!!

server/graphql-schema.js
alflennik: What do you think about calling this "atOperatingMode"? I have a couple of motivations for the suggestion. One is that since the word "settings" is plural it makes me suspicious that I'm dealing with something like a comma separated list. Another is that settings is quite a broad term, and are not obviously dealing specifically with the AT, for example it could be referring to operating system settings, browser settings, keyboard settings like holding the shift key, any kind of settings, and I think that's one of the things which is confusing me. I suggested "operatingMode" because I feel like "atMode" is a bit too restrictive a term.

Also, can you improve the documentation for this field? I think we should exhaustively list out all the values which are currently used, one by one. Also if you have a link to a specification listing all possible options, putting that here would be hugely helpful as well.

alflennik:
Thanks for addressing my big piece of feedback! Also I followed up on the conversation about the settings field.

server/graphql-schema.js
howard-e: **On settings > atOperatingMode request**

> I suggested "[atO]peratingMode" because I feel like "atMode" is a bit too restrictive a term.

Agreed, I can make that change.

> ... since the word "settings" is plural it makes me suspicious that I'm dealing with something like a comma separated list. 

Note that eventually, [this may happen](https://github.com/w3c/aria-at/issues/1002). That's based on the spec request of [AT_KEY-commands.csv > settings](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#settings).

**On documentation update request**

> Also, can you improve the documentation for this field? I think we should exhaustively list out all the values which are currently used, one by one.

Listing them out will require a change [any time they're updated](https://github.com/w3c/aria-at/commit/399b44ece13542345425c6593227d76dfd9631d2), which would be of an unknown frequency. That would make that bit of inline documentation irrelevant for a bit of time and I'm doubtful of how best that will be kept updated.

> Also if you have a link to a specification listing all possible options, putting that here would be hugely helpful as well.

I don't but let's suggest updating the [v2 Test Format Definition](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2) with a section on that and link there. Otherwise, the only spec for that is to check out [each key in ats[].settings in `support.json`](https://github.com/w3c/aria-at/blob/master/tests/support.json), which isn't too easy.

client/components/common/queries.js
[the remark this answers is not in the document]
howard-e: I can see that benefit, as well as some others, especially in the mock testing area and such. For clarity though, the fields replaced by a fragment has the exact same fields as were listed before, so no additional _(or unnecessary)_ fields just yet. But I do think there's a benefit to this approach, and an even greater benefit if that structure is set up **properly**.

So I'd rather to remove this proposed enhancement bit from these unrelated code changes and follow up in a separate issue so this enhancement can get it's own focused space.

alflennik (approved):
Thanks for addressing all the changes I requested! Also, big thanks for adding that branch with assertionExceptions populated, that was invaluable to make it easier to test this.

server/scripts/import-tests/index.js
```diff
@@ -465,6 +466,33 @@ const getTests = ({
...
+                                    let priority = '';
+                                    if (assertionPriority === 0)
+                                        priority = 'EXCLUDE';
+                                    if (assertionPriority === 1)
+                                        priority = 'MUST';
+                                    if (assertionPriority === 2)
+                                        priority = 'SHOULD';
+                                    if (assertionPriority === 3)
+                                        priority = 'MAY';
+
```
stalgiag, line 487: It would be nice to have a utility variable that maps these ints to these strings (which could also be enums). This would help centralize and lift the significance of these relationships which would be helpful for someone like me that was just editing `assertionExceptions`` on my local aria-at without fully understanding what the numbers meant.

howard-e: Good point. There's the `server/resolvers/helpers/convertAssertionPriority.js` file, which could be moved to a more appropriate location, and be expanded to something like:

```javascript
const convertAssertionPriority = priority => {
    if (priority === 0) return 'EXCLUDE';
    if (priority === 'REQUIRED' || priority === 1) return 'MUST';
    if (priority === 'OPTIONAL' || priority === 2) return 'SHOULD';
    if (priority === 3) return 'MAY';
};

module.exports = convertAssertionPriority;
```

Would this support what you're thinking of?

client/components/TestRun/index.jsx
```diff
@@ -656,10 +659,14 @@ const TestRun = () => {
                 id,
                 output: output,
                 unexpectedBehaviors: unexpectedBehaviors,
-                assertionResults: assertionResults.map(({ id, passed }) => ({
-                    id,
-                    passed
-                }))
+                assertionResults: assertionResults
+                    // All assertions are always being passed from the TestRenderer results, but in the instance where
+                    // there is a 0-priority exception, and id won't be provided and that cannot be saved
```
stalgiag, line 664: Is this taking advantage of a side effect or is this the strategy used for communicating that an assertion is excluded? It seems easy to misunderstand or forget. I would expect them to be passed prefiltered but I might be missing something.

howard-e: It's been some time now, so I'll have to look back into why exactly it was done that way. But from what I can remember, the `TestRenderer` still technically requires the index of that excluded assertion to properly determine which assertions are which, but at some point. It definitely had to be removed before saving and those excluded ones had no ids. That was a part of fulfilling [this request](https://github.com/w3c/aria-at-app/pull/863#pullrequestreview-1778516739) with https://github.com/w3c/aria-at-app/pull/863/commits/5c4ec6a5568b685e1c7f044f7556912c32b811a6.

> Is this taking advantage of a side effect or is this the strategy used for communicating that an assertion is excluded?

So seems to be a bit of both.

This point in the code looks like the safest place to filter them out, as the single point before being passed to the server/db. The comment could be much clearer as well.

server/scripts/import-tests/index.js
stalgiag: That would work. Thank you!

client/components/TestRun/index.jsx
stalgiag: Oh okay. The background is helpful. I'll leave it to you as to whether you want to change the comment or address it any other way.

stalgiag (approved):
I am approving after our discussion in the two comment threads. I know that some temporary branch settings need to be reset before merging so I will pass this back to you with an approval to handle those changes.

client/resources/aria-at-test-io-format.mjs
```diff
@@ -778,7 +783,31 @@ class BehaviorInput {
...
+              cs.presentationNumber === c.presentationNumber &&
+              cs.settings === c.settings
+          );
+          if (!foundCommandInfo || !foundCommandInfo.assertionExceptions) return cs;
+
+          // Only works for v2
+          let assertionExceptions = json.output_assertions.map(each => each.assertionId);
+          foundCommandInfo.assertionExceptions.split(' ').forEach(each => {
+            let [priority, assertionId] = each.split(':');
+            const index = assertionExceptions.findIndex(each => each === assertionId);
```
jugglinmike, line 799: `Array.prototype.indexOf` would be a bit more readable here

client/resources/aria-at-test-io-format.mjs
```diff
@@ -778,7 +783,31 @@ class BehaviorInput {
...
+
+            priority = Number(priority);
+            assertionExceptions[index] = priority;
```
jugglinmike, line 802: Under the assumption that "fewer mutable bindings is better", these statements could be combined to allow swapping the earlier `let` for a `const`.

client/resources/aria-at-test-io-format.mjs
```diff
@@ -778,7 +783,31 @@ class BehaviorInput {
...
+          });
+          // Preserve default priority or update with exception
+          assertionExceptions = assertionExceptions.map((each, index) =>
+            isNaN(each) ? json.output_assertions[index].priority : each
```
jugglinmike, line 806: Aren't the members of `assertions` string values? Returning `each` from a filter predicate suggests otherwise.

server/resolvers/TestPlanRunOperations/createTestResultSkeleton.js
```diff
@@ -30,16 +30,26 @@ const createTestResultSkeleton = ({
...
+                        .filter(assertion => {
+                            return !assertion.assertionExceptions?.some(
+                                e =>
+                                    scenario.commands.find(
+                                        c =>
+                                            c.id === e.commandId &&
+                                            c.atOperatingMode === e.settings
+                                    ) && e.priority === 'EXCLUDE'
+                            );
+                        })
```
jugglinmike, line 44: The comment doesn't seem to acknowledge the first half of the filter expression, making me wonder whether I understand what's going on here. Would it be accurate to say, "Filter out assertionResults for the current scenario which were marked with a 0-priority exception"?

Stylistically, I think that the closure nesting is reaching a point where it interferes with readability. While the in-line documentation you've suggested helps, breaking this logic out into a standalone function would afford us more room to describe what's going on, reduce the cognitive load (by formalizing the inputs), and limit distraction for folks seeking to understand `createTestResultSkeleton` at a high level.

```js
/**
 * Determine whether a given assertion belongs to a given scenario and includes
 * at least one exception with a given priority.
 *
 * @param {string} priority
 * @param {Scenario} scenario
 * @param {Assertion} assertion
 */
const hasExceptionWithPriority = (priority, scenario, assertion) => {
  return assertion.assertionExceptions?.some(
      e =>
          scenario.commands.find(
              c =>
                  c.id === e.commandId &&
                  c.atOperatingMode === e.settings
          ) && e.priority === priority
  );
};
```

Then invoking like this:

    .filter(assertion => !hasExceptionWithPriority('EXCLUDE', scenario, assertion))

I'm only sharing something concrete as a means to demonstrate my point; I'm not too concerned with the specific factoring. For instance, a more purpose-built function would simplify the callsite even further (though it might be difficult to name it).

jugglinmike (approved):
Looks good, @howard-e; this works as expected, so nothing blocking from me.

> Note 1: This work is dependent on https://github.com/w3c/aria-at/pull/1012 landing and any updated `client/resources` files here should be reviewed there.

That patch was recently merged, so I've left my comments here.

>     // Filter out assertionResults which were marked with a 0-priority exception
>     .filter(assertion => {
>         return !assertion.assertionExceptions?.some(
>             e =>
>                 scenario.commands.find(
>                     c =>
>                         c.id === e.commandId &&
>                         c.atOperatingMode === e.settings
>                 ) && e.priority === 'EXCLUDE'
>         );
>     })

The comment doesn't seem to acknowledge the first half of the filter expression, making me wonder whether I understand what's going on here. Would it be accurate to say, "Filter out assertionResults for the current scenario which were marked with a 0-priority exception"?

Stylistically, I think that the closure nesting is reaching a point where it interferes with readability. While the in-line documentation you've suggested helps, breaking this logic out into a standalone function would afford us more room to describe what's going on, reduce the cognitive load (by formalizing the inputs), and limit distraction for folks seeking to understand `createTestResultSkeleton` at a high level.

```js
/**
 * Determine whether a given assertion belongs to a given scenario and includes
 * at least one exception with a given priority.
 *
 * @param {string} priority
 * @param {Scenario} scenario
 * @param {Assertion} assertion
 */
const hasExceptionWithPriority = (priority, scenario, assertion) => {
  return assertion.assertionExceptions?.some(
      e =>
          scenario.commands.find(
              c =>
                  c.id === e.commandId &&
                  c.atOperatingMode === e.settings
          ) && e.priority === priority
  );
};
```

Then invoking like this:

    .filter(assertion => !hasExceptionWithPriority('EXCLUDE', scenario, assertion))

I'm only sharing something concrete as a means to demonstrate my point; I'm not too concerned with the specific factoring. For instance, a more purpose-built function would simplify the callsite even further (though it might be difficult to name it).

server/scripts/import-tests/index.js
howard-e: Done in https://github.com/w3c/aria-at-app/pull/863/commits/3d1fbcb61e4e42ed3cc56802fe90827f7a001d8a

server/resolvers/TestPlanRunOperations/createTestResultSkeleton.js
howard-e: +1 on this. I was able to take your suggestion and only apply minor changes, thanks! Done in https://github.com/w3c/aria-at-app/pull/863/commits/bba8ad4edb425d890effddf85dc9afa4aa535261

howard-e (commented):
@jugglinmike I appreciate your feedback on the aria-at related components. Sad that the PR was merged before but I can create a follow up PR there and request your review afterwards

PR #869, opened 2023-12-11
Correct bug in dropdown rendering

jugglinmike:
Limit the insertion of horizontal padding to the specific case which requires it.


stalgiag (approved):
I noticed that the indent was looking off in the dropdown during one of your demos. This change addresses that effectively. Thank you!

jugglinmike:
Great, thanks for the review!

Issue #877, opened 2023-12-14
Deprecated test plans remain in test queue and in-progress results are lost when a newer test plan version deprecates an older test plan version


mcking65:
Advancing a new version of a test plan to draft review caused the test queue to contain test runs for a deprecated test plan version and cause the loss of previously completed work on manual test runs.

The procedure that manifest this bug was:
1. Command button V23.12.06 was in draft review and had three test runs in the test queue.
2. Results were present for all the runs.
3. We discovered that a change was needed to the settings specified for two VoiceOver commands.
4. We made V23.12.13 of the plan with the changes to the VoiceOver commands and merged it to master so it showed on the data management page as R&D Complete.
5.  We advanced V23.12.13 of the command button test plan to draft review.

Expected results:
1. Since V23.12.06 is deprecated, its runs no longer show in the test queue.
2. To avoid unnecessary loss of manual test work, equivalent runs of V23.12.13 are added to the test queue, preserving as many test results as possible.

Actual results:
1. Runs of V23.12.06, which became deprecated,  remain in the test queue.
2. Runs of V23.12.13 that we subsequently added to the test queue were empty, so testers have to start over even for tests that were not changed.

### Requirements for data preservation during test plan deprecation

To prevent unnecessary loss or repetition of manual test run work, the system needs to determine which test results can be preserved when a version of a test plan is deprecated as the result of a test plan advancing into draft review, into candidate review, or into recommended.

the system should:
1. Identify AT/browser preservation candidates: For each AT/browser combination supported by the test plan, identify the AT/browser combinations for which the deprecated plan has results (either draft or final) and the new plan version does not have results. If the new version has results for a given combination, it is not necessary to attempt to copy any results from the prior test plan version to the new test plan version for that AT/browser combination. This comparison can ignore AT and browser versions, e.g., if the new plan version and the deprecated plan version both have results for JAWS/Chrome, the version of JAWS and Chrome are un important. There is no need to copy results for JAWS/Chrome in that scenario.
2. Identify preservable results: For each AT/browser preservation candidate, identify the test/assertion/command/settings that are equivalent in the new and deprecated plans.
3. If an AT/Browser preservation candidate has a final report generated from the deprecated test plan and if all results are preservable, copy the final report from the deprecated test plan to the new test plan.
4. If an AT/Browser preservation candidate has a final report generated from the deprecated test plan and if some but not all results are preservable, then:
    1. copy results for the equivalent tests/assertions/commands into new incomplete test plan runs.
[...]

PR #880, opened 2023-12-18
Update `hashTests` function to support the migration of test results across test plan versions using the v2 test format

howard-e:
This PR updates `aria.js` to account for properties to be ignored during a v2 test format version update so applicable results can be transferred over during phase update process.

This should address the scenario described in #877 and produce the expected results along with the additional requirements:

> The procedure that manifest this bug was:
> 
> 1. Command button V23.12.06 was in draft review and had three test runs in the test queue.
> 2. Results were present for all the runs.
> 3. We discovered that a change was needed to the settings specified for two VoiceOver commands.
> 4. We made V23.12.13 of the plan with the changes to the VoiceOver commands and merged it to master so it showed on the data management page as R&D Complete.
> 5. We advanced V23.12.13 of the command button test plan to draft review.
> 
> Expected results:
> 
> 1. Since V23.12.06 is deprecated, its runs no longer show in the test queue.
> 2. To avoid unnecessary loss of manual test work, equivalent runs of V23.12.13 are added to the test queue, preserving as many test results as possible.
> 
> Actual results:
> 
> 1. Runs of V23.12.06, which became deprecated,  remain in the test queue.
> 2. Runs of V23.12.13 that we subsequently added to the test queue were empty, so testers have to start over even for tests that were not changed.


server/util/aria.test.js
```diff
@@ -79,3 +83,97 @@ describe('Verify test hashes are matching as expected', () => {
...
+        const v2TestHash_2_JAWS = hashTest(commandButton_V20231213_V2[0]);
+
+        const v2TestHash_1_NVDA = hashTest(commandButton_V20231206_V2[1]);
+        const v2TestHash_2_NVDA = hashTest(commandButton_V20231213_V2[1]);
+
+        expect(v2TestHash_1_JAWS).toEqual(v2TestHash_2_JAWS);
+        expect(v2TestHash_1_NVDA).toEqual(v2TestHash_2_NVDA);
+    });
+
+    /* eslint-disable */
```
alflennik, line 143: Would it be possible to explain in a comment why disabling linting is needed?

server/util/aria.js
```diff
@@ -8,7 +8,14 @@ const evaluateAtNameKey = atName => {
 };
 
 const testWithNoIds = test => ({
```
alflennik, line 10: It seems like it might be time to update this variable name since it's now responsible for filtering a lot of information which is not really IDs.

server/util/aria.test.js
```diff
@@ -79,3 +83,97 @@ describe('Verify test hashes are matching as expected', () => {
         expect(testsHashA).not.toEqual(testsHashB);
     });
 });
+
+describe('Verify test hashes are matching as expected - (v2 test format)', () => {
```
alflennik, line 87: Huge thanks for expanding these tests so much! Hashes are such delicate flowers that it really eases my mind a lot to know we have some extensive testing in place to make sure they keep working.

alflennik (approved):
I am approving! There's a couple of minor comments but it all feels optional. I verified the tests are passing (big thank you for buffing up the test coverage), and also I rebuilt the app from scratch to verify that it was still able to compare test versions without introducing duplicates.

server/util/aria.test.js
howard-e: I think I included that in error when pushing this PR. I think I was ignoring the warnings to make the code changes here easier to read for myself.

But there is a point to be made here, that the default `printWidth` wrapping value made that block of code a bit hard to read (to me, and this does apply elsewhere, especially in frontend components). There was a discussion sometime back on this, which should be had to see what we think on updating that value.

howard-e: Removed the `eslint-disable` in https://github.com/w3c/aria-at-app/pull/880/commits/eeb2fb8f1e11176d3c4e27a40c05d2e2f1ce1ed4

server/util/aria.js
howard-e: Good catch! Done in https://github.com/w3c/aria-at-app/pull/880/commits/eeb2fb8f1e11176d3c4e27a40c05d2e2f1ce1ed4

server/util/aria.test.js
alflennik: Yeah I'd love to update the tabWidth and printWidth 🤠

server/migrations/utils/index.js
```diff
@@ -0,0 +1,245 @@
...
+                    await queryInterface.sequelize.query(
+                        `UPDATE "TestPlanVersion" SET "hashedTests" = ? WHERE id = ?`,
+                        {
+                            replacements: [hashedTests, testPlanVersion.id],
+                            transaction
+                        }
+                    );
+                } else {
+                    // Remove any accidentally created TestPlanVersion rows created prior to migration
+                    await queryInterface.sequelize.query(
```
howard-e, line 69: For reviewers, found an issue during checks against production where several TestPlanVersions were created as duplicates (yet had non-matching `hashedTests`, because of some failed sync between pushes I suppose).

Also broke this out into a separate utility because several migrations may have to make use this in the future as well. Immediately seeing a need for #863 to use this as well.

stalgiag (approved):
Thanks for the updates and accompanying expanded test coverage. I was able to confirm that this was functioning as expected by rebuilding the DB and running the unit tests. The new code also has greater clarity than what it replaces. Approving!

alflennik (approved):
I'm happy to approve a second time on this one. I like the idea of making it easy to recalculate the hashes, it does seem pretty likely it might become necessary down the line. I took a look at the broader changes to the tests as well and didn't see anything that rang any alarm bells. I see the tests are passing. Nice work!

PR #881, opened 2023-12-18
Add NVDA bot user using a seeder

stalgiag:
Add NVDA Bot as a User to the DB using a sequelize seeder.

Previously the NVDA Bot `User` was being added to the DB upon the creation of the first `CollectionJob`.

Note that the user has a predetermined id (9999). This was done in order to account for the way that the `User` table is populated currently. With the current approach to DB population, [seeders run before](https://github.com/w3c/aria-at-app/blob/main/.github/workflows/runtest.yml#L37-L42) the `populate-test-data-` script that [stores a DB dump](https://github.com/w3c/aria-at-app/blob/main/server/scripts/populate-test-data/pg_dump_2021_05_test_data.sql#L103-L104). This means that if an ID isn't specified in the seeder then the id of 1 will be automatically assigned causing a collision during the restoration of the dump. The NVDA bot user was not added directly to the DB dump because presumably the ID 3 is already taken and if we have to assign a "magically high number" then best to retain the down functionality of a Sequelize seeder.

### DIFF
--- a/server/models/services/CollectionJobService.js
+++ b/server/models/services/CollectionJobService.js
@@ -18,11 +18,6 @@ const {
     createTestPlanRun,
     removeTestPlanRun
 } = require('./TestPlanRunService');
-const {
-    createUser,
-    getUserByUsername,
-    addUserToRole
-} = require('./UserService');
 const responseCollectionUser = require('../../util/responseCollectionUser');
 const { getTestPlanReportById } = require('./TestPlanReportService');
 const { HttpQueryError } = require('apollo-server-core');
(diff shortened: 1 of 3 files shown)


server/seeders/20231218191524-addNVDABot.js
```diff
@@ -0,0 +1,33 @@
...
+const responseCollectionUser = require('../util/responseCollectionUser');
+
+module.exports = {
+    async up(queryInterface) {
+        const user = await queryInterface.bulkInsert(
+            'User',
+            [
+                {
+                    id: responseCollectionUser.id, // Specified ID for NVDA Bot
+                    username: 'NVDA Bot',
```
jugglinmike, line 12:
```suggestion
                    username: responseCollectionUser.username,
```

jugglinmike (changes requested):

server/seeders/20231218191524-addNVDABot.js
stalgiag: Good thought! Applied it

server/models/services/CollectionJobService.js
```diff
@@ -106,26 +101,8 @@ const createCollectionJob = async (
     options
 ) => {
     if (!testPlanRun) {
-        let user = await getUserByUsername(responseCollectionUser.username);
-        if (!user) {
-            const roles = [{ name: User.TESTER }];
-            user = await createUser(
```
jugglinmike, line 112: It would have been easy to persist this obviated logic, though I can only imagine the kind of confusion it might have caused in some future exceptional circumstance. Thanks for being thorough and removing it!

jugglinmike (approved):
The code looks good to me, and it does the right thing in the development environment! I'll ask you to merge at your convenience since the target branch is a little hot right now.

PR #882, opened 2023-12-18
Remove last remaining "Success Criteria" section

alflennik:
Removes the success criteria section from the Candidate Review page as well as the Test Review page. I tackled this in the simplest way I could (the naive way?) so I would encourage you to not be shy if you can think of a better way to do this.


stalgiag (approved):
Great surgical change that I was able to verify in both locations. Think the hard-coded "Assertions" header makes sense in this instance. Approved!

howard-e (approved):
Looks good! I think additional changes may come to this after #883 but this if fine at the moment. Making the header `Assertions` for now seems like a good move to me.

Issue #799, opened 2023-09-27
Cannot view reports for draft test plans


mcking65:
Issues #649 and #719 include requirements that are intended to enable users to access reports from test plans that are in draft review. These reports are intentionally not available from the reports page because we do not want them widely visible, however we still need to be able to review them before advancing a plan to candidate review.

After a report from a test plan that is in draft review is marked final, it should be available from either the report status dialog or the test plan versions page. Those report links are not provided as specified in issues #649 and #719.

In issue #649, the description of the Report Status column is:
>* If the report is complete, a link to the report with text "Report Completed on DATE" 
>* If report is missing, the string "Missing", and if user is admin, button "Add to Test Queue" 
>* If report is not complete and not missing and not in test queue, the string "None", and if user is admin, "Add to Test Queue" button 
>* If report is in the test queue, the string "X% complete by Y testers with Z conflicts" 
>* Note: Percent complete = NUMBER_COMPLETED_TESTS_BY_ALL_TESTERS / (NUMBER_ASSIGNED_TESTERS * NUMBER_TESTS_IN_PLAN) * 100 

Currently, if a test plan is in draft review and a report is complete, the report status column is showing the value that it should show if the test plan run for that report were still in the test queue and not yet marked final. For example, it shows:

>100% complete by 2 testers with 0 conflicts

Here is a screen shot of the report status dialog for meter that I get after finalizing all required reports. This is the dialog when there are 0 runs of the meter test plan in the test queue.

![screenshot of report status dialog](https://github.com/w3c/aria-at-app/assets/7241781/50c40469-40a2-43a5-b728-e559763e70ce)

Similarly, as specified in #719, the [test plan versions page for meter](https://aria-at-staging.w3.org/data-management/meter) does not include a link to view reports under the heading with text:

>V22.08.05 Draft on Sep 22, 2023

![screenshot of test plan versions page for meter](https://github.com/w3c/aria-at-app/assets/7241781/241bd714-0707-48a0-ada4-ff72b466128d)

mcking65:
Re-opening; I can't find the links to reports for test plans in the draft review phase. 

Looking at prod, the toggle button test plan V24.03.12 is in draft review and some of the required reports are complete. Here is the data management page.

![screenshot of data management page](https://github.com/w3c/aria-at-app/assets/7241781/028becf8-2884-4259-9b60-95493bf20c24)

[...]

PR #885, opened 2023-12-20
Add links to draft reports

alflennik:
See https://github.com/w3c/aria-at-app/issues/799. Although we currently have the ability to show reports for test plans currently in a draft state, currently there are no links to those reports, making them inaccessible. This PR makes draft reports accessible on both the test plan version page and within the report status dialog.


howard-e (approved):
Looks good and verified the report links are showing in the expected locations. Thanks!

PR #886, opened 2023-12-21
Support automation reporting test results by presentationNumber for v2 tests

stalgiag:
In the app, v1 tests have the `rawTestId` value in the `rowNumber` [property as seen here](https://github.com/w3c/aria-at-app/blob/8215cb96af64aba1cf84b31e86d480668bf8a37e/server/scripts/import-tests/index.js#L485). Whereas, v2 tests use the `presentationNumber` in the `rowNumber` property [as seen here.](https://github.com/w3c/aria-at-app/blob/56205a32f904964dbb36c943d4e3fdb9d2522ce9/server/scripts/import-tests/index.js#L533).

For this reason, the app hasn't supported receiving results from v2 tests. After [these updates to the harness](https://github.com/w3c/aria-at-automation-harness/pull/35), the harness now sends the `presentationNumber`. This change to the app will enable look up of the test by this value.

---
  - https://app.asana.com/0/0/1206226143670047

### DIFF
--- a/server/controllers/AutomationController.js
+++ b/server/controllers/AutomationController.js
@@ -135,7 +135,7 @@ const getApprovedFinalizedTestResults = async testPlanRun => {
 };
 
 const updateOrCreateTestResultWithResponses = async ({
-    testCsvRow,
+    testRowIdentifier,
     testPlanRun,
     responses,
     atVersionId,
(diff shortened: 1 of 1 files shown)


gnarf (approved):
This all makes sense to me!

jugglinmike:
The change appears valid, but we can't run it without changes to the target branch. Since [that branch is itself under review](https://github.com/w3c/aria-at-app/pull/879), I'm going to merge this so we can validate the automation work as a whole.

PR #887, opened 2024-01-02
Update `ansible` script and instructions on deploying to the updated staging and production environments

howard-e:
The host server of the staging and production environments has been changed. This updates the relevant ansible task and instructions to account for this change.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -59,6 +59,7 @@ To deploy this project to  server:
     Host *.w3.internal
       ProxyJump bocoupinfra@ssh-aws.w3.org
     ```
+    - The RSA key fingerprint for `ssh-aws.w3.org` is `SHA256:Nlyly0XFZebw0/k8yCGUXA+Y03W7WB7CPnXiZZb7XE8`.
   - Run `ssh root@bestla.w3.internal` (aria-at-staging.w3.org's server) and `ssh root@fenrir.w3.internal` (aria-at.w3.org's server) to verify that you can connect to the servers.
     - The RSA key fingerprint for `bestla.w3.internal` is `SHA256:F16aX2Wx4e39jbHhqEkeH8iRgY41C3WgxvAgvh7PQZ0`.
     - The RSA key fingerprint for `fenrir.w3.internal` is `SHA256:cF6u/K00P2ELEVbIazVVqqMz5q+Sbh4+Jog/VmXZomg`.
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -60,6 +60,8 @@ To deploy this project to  server:
       ProxyJump bocoupinfra@ssh-aws.w3.org
     ```
   - Run `ssh root@bestla.w3.internal` (aria-at-staging.w3.org's server) and `ssh root@fenrir.w3.internal` (aria-at.w3.org's server) to verify that you can connect to the servers.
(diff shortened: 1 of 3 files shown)


deploy/README.md
```diff
@@ -54,10 +54,15 @@ To deploy this project to  server:
   - Place it in the ~/.ssh directory.
   - For security, set permissions on the key file, which is required by the OS: `chmod 600 ~/.ssh/aria-at-bocoup`.
   - Add it to your keychain with the following command: `ssh-add ~/.ssh/aria-at-bocoup`.
-  - Run `ssh root@aria-at-staging.w3.org` and `ssh root@aria-at.w3.org` to verify that you can connect to the servers.
+  - Add the following `Host` client configuration option to `~/.ssh/config`:
+    ```
+    Host *.w3.internal
+      ProxyJump bocoupinfra@ssh-aws.w3.org
+    ```
+  - Run `ssh root@bestla.w3.internal` (aria-at-staging.w3.org's server) and `ssh root@fenrir.w3.internal` (aria-at.w3.org's server) to verify that you can connect to the servers.
```
jugglinmike, line 62: Can you include the RSA key fingerprint for each of these servers?

howard-e: @jugglinmike done in https://github.com/w3c/aria-at-app/pull/887/commits/2aad34841b5fdd1f3b43bd1d3d37dfb74bf3660d.

jugglinmike: Thanks! Could you also share the one for ssh-aws.w3c.org? (Sorry, my original request was ambiguous about *which* servers, though I admittedly didn't recognize the need for that server's fingerprint.)

jugglinmike: That's surprising to me--I only asked because I needed it. After all, isn't that what the ProxyJump configuration is all about?

jugglinmike (approved):
I was able to safely connect using these instructions! I think this is good to merge, but I'm going to hold off on doing so because I have an open question about Howard's process which might inform further changes.

deploy/README.md
howard-e: @jugglinmike I thought the same as well, but based on the instructions provided by W3C. We needed to only directly interact with bestla.w3.internal and fenrir.w3.internal (for this particular migration).

That isn't to say a future need may not arise but for now, I was able to perform expected deployment actions and checks without directly interacting with ssh-aws.w3.org.

jugglinmike: It almost sounds like we don't need the SSH configuration, but I'm no longer able to connect if I remove it:

```
$ ssh root@fenrir.w3.internal
ssh: Could not resolve hostname fenrir.w3.internal: Name or service not known
$ ssh root@bestla.w3.internal
ssh: Could not resolve hostname bestla.w3.internal: Name or service not known
```

Does `ssh-keygen -H -F ssh-aws.w3.org` come up empty for you?

jugglinmike: Oh, I guess I know the answer to that question. Maybe more to the point: can you connect if you remove the SSH configuration?

howard-e: > Oh, I guess I know the answer to that question. Maybe more to the point: can you connect if you remove the SSH configuration?

Nope!

jugglinmike: Okay, so I think we need the SSH configuration and the fingerprint.

Issue #810, opened 2023-10-05
Display only major browser version number when prompting user to confirm the version at the start of a test run session


ccanash:
This issue is related to the version confirmation dialog that appears when a tester activates a link on the test queue page to open a test plan in the test runner.

It seems that the app is saving the major version number but detecting the full version number. The information presented in the browser version field in the dialog is confusing because .0.0 is appended to the major version. That gives users the impression that the full version is expected and saved. But that is not the case.

If the app is only saving the major version number, it should not display .0.0 in the version number field. This is probably one reason for the incorrect alert.

_Originally posted by @mcking65 in https://github.com/w3c/aria-at-app/issues/797#issuecomment-1736555503_

mcking65:
@evmiguel  wrote:

>The app saves the full browser version in the database, which is why .0.0 is being shown in the version number field.

For Chrome, for example, 121.0.0 is not a valid full version number. There is no such version as far as I can tell. So, if the full version is saved, why is a number like 121.0.0 shown as the saved number? The number should be something like 121.0.6167.185.

>This is the current behavior implemented, but were you looking to have the major version saved instead of the full version?

In this issue I proposed saving and showing only the major version number. However, if we are already saving the full version number, then an alternative is to continue to save the full version number and then show the full version number in the app. 

Continuing to show an invalid, non-existent version is confusing.

It remains to be determined how important the browser version is to the ARIA-AT work. So, perhaps the the original intent I expressed in this issue is not the best course of action. In the absence of clear evidence that the full version is not important, it is probably safer to keep storing it if it does not create additional work to do so.

evmiguel:
@mcking65 Since we were saving invalid versions, we decided to show the major browser version, but save minor versions and patches in the database if they were present.

PR #889, opened 2024-01-04
Show major browser version on first load of Test Run

evmiguel:
This PR addresses #810, which aims to only show the major browser version when confirming the user's browser version.

---
  - https://app.asana.com/0/0/1206550059310407

### DIFF
--- a/client/components/common/AtAndBrowserDetailsModal/index.jsx
+++ b/client/components/common/AtAndBrowserDetailsModal/index.jsx
@@ -198,17 +198,17 @@ const AtAndBrowserDetailsModal = ({
         // Discard trailing zeros in found minor and patches.
         // This also accounts for trailing numbers after the patch.
         let fullVersion = '';
-        let [majorVersion, minorVersion, patch, ...rest] =
+        let [majorVersion, minorVersion = '0', patchVersion = '0', ...rest] =
             updatedBrowserVersion.split('.');
 
-        if (minorVersion === '0' && (patch === '0' || patch === undefined)) {
+        if (minorVersion === '0' && patchVersion === '0') {
             fullVersion = `${majorVersion}`;
-        } else if (patch === '0' || patch === undefined) {
+        } else if (patchVersion === '0') {
             fullVersion = `${majorVersion}.${minorVersion}`;
         } else if (rest === undefined || rest.length === 0) {
-            fullVersion = `${majorVersion}.${minorVersion}.${patch}`;
+            fullVersion = `${majorVersion}.${minorVersion}.${patchVersion}`;
         } else {
-            fullVersion = `${majorVersion}.${minorVersion}.${patch}.${rest.join(
+            fullVersion = `${majorVersion}.${minorVersion}.${patchVersion}.${rest.join(
                 '.'
             )}`;
         }
--- a/client/components/common/AtAndBrowserDetailsModal/index.jsx
+++ b/client/components/common/AtAndBrowserDetailsModal/index.jsx
@@ -54,6 +54,8 @@ const AtAndBrowserDetailsModal = ({
     handleAction = () => {},
     handleClose = () => {}
 }) => {
+    // Only take into account the major version
+    // of the provided browser version
     if (browserVersion.includes('.')) {
         browserVersion = browserVersion.split('.')[0];
     }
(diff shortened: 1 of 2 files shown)


client/components/common/AtAndBrowserDetailsModal/index.jsx
```diff
@@ -90,15 +95,13 @@ const AtAndBrowserDetailsModal = ({
         }
 
         const foundBrowserVersion =
-            uaBrowser === browserName &&
-            uaMajor !== '0' &&
-            `${uaMajor}.${uaMinor}.${uaPatch}`;
+            uaBrowser === browserName && uaMajor !== '0' && `${uaMajor}`;
```
howard-e, line 98:
```suggestion
            uaBrowser === browserName && uaMajor !== '0' && uaMajor;
```

client/components/common/AtAndBrowserDetailsModal/index.jsx
```diff
@@ -90,15 +95,13 @@ const AtAndBrowserDetailsModal = ({
...
 
         if (
             // don't force browserVersion update with admin (unless first run)
             (!isAdmin || (isAdmin && firstLoad)) &&
             // check that saved browserVersion is the same as detected
-            !browserVersion.includes(`${uaMajor}.${uaMinor}.${uaPatch}`) &&
+            !browserVersion.includes(`${uaMajor}`) &&
```
howard-e, line 104: Leaving as the string template does make it more intentional but the hook does return `ua...` values as string if you'd like to simplify this.

The function would also implicitly convert a numeric value to string as well. But this is a nit!

```suggestion
            !browserVersion.includes(uaMajor) &&
```

howard-e (commented):
Thanks @evmiguel! Everything looks good and works as expected.

I think you could also include a change to the displayed browser version at the top of the Test Run page as well. See screenshot.

screenshot showing full version number information for browser at top of test run page, as Chrome 120.0.0

client/components/common/AtAndBrowserDetailsModal/index.jsx
```diff
@@ -194,7 +194,26 @@ const AtAndBrowserDetailsModal = ({
                 );
         }
 
-        handleAction(updatedAtVersion, updatedBrowserVersion, updateMessage);
+        // Save full version even though we don't display it.
+        // Discard trailing zeros in found minor and patches.
+        // This also accounts for trailing numbers after the patch.
+        let fullVersion = '';
+        let [majorVersion, minorVersion, patch, ...rest] =
```
howard-e, line 201:
```suggestion
        let [majorVersion, minorVersion = '0', patch = '0', ...rest] =
```

Also, could you use `patchVersion` for consistency?

client/components/common/AtAndBrowserDetailsModal/index.jsx
```diff
@@ -194,7 +194,26 @@ const AtAndBrowserDetailsModal = ({
...
+            updatedBrowserVersion.split('.');
+
+        if (minorVersion === '0' && (patch === '0' || patch === undefined)) {
```
howard-e, line 204:
```suggestion
        if (minorVersion === '0' && patch === '0') {
```

The `minorVersion` being `undefined` wasn't being handled which was causing cases of `{MAJOR}.undefined` being saved to the database.

client/components/common/AtAndBrowserDetailsModal/index.jsx
```diff
@@ -194,7 +194,26 @@ const AtAndBrowserDetailsModal = ({
...
+            fullVersion = `${majorVersion}`;
+        } else if (patch === '0' || patch === undefined) {
```
howard-e, line 206:
```suggestion
        } else if (patch === '0') {
```

howard-e (commented):
@evmiguel looks good, thanks for taking a look at this! I've left some inline comments and suggestions before approving.

howard-e (approved):
@evmiguel looks good! Thanks for addressing my feedback

alflennik:
Hi @evmiguel thanks for sorting out the question I raised about how to tackle this issue! 

So I created a new entry in the test queue using chrome and it detected and displayed a version of 122, which is perfect. However when I looked at the database it did not store the minor or patch parts of the version number. Is that expected? See below for a screenshot of the database contents.

Table of browser versions in the database which include `99.0.4844.84` and `122` for chrome

If we plan to store the full version number, which I think we do, I think this might be a bug worth fixing. If we're only planning on storing the major number I think we should update our automatically seeded data to only include the major number.

The full version number, by the way, was 122.0.0, I got that number like this:

Code reading console.log(`${uaMajor}.${uaMinor}.${uaPatch}`)

alflennik:
Ah ok, sounds good then!

alflennik (approved):

Issue #891, opened 2024-01-08
Remove TestPlanUpdater component and related references


Paul-Clue:
The `TestPlanUpdater` component is no longer being used, so It should be removed from the app.

PR #893, opened 2024-01-10
Remove test plan report update modal

Paul-Clue:
see issue #891

This pull request removes the `UpdateTestPlanReportModal` from the app.


howard-e (approved):
Thanks for cleaning that up!

alflennik (approved):
Thanks for removing this, that was a lot of dead code that could definitely trip new contributors up and get them confused. Not too much to test here, but I did verify that the test queue page is still working as normal.

Issue #797, opened 2023-09-27
Inaccurate screen reader alert appears in actions column of test queue after assigning a tester


mcking65:
To reproduce:
1. Open test queue and go to any test plan.
2. In the testers column, open Assign testers menu and choose a tester.
3. After the assignment is made and the queue page is updated, read the actions column.

Result: After the buttons in the actions column, JAWS and NVDA see the message:

>You have been removed from this test run.

It comes from this element in the cell:

~~~
<div class="at-alert" role="status" aria-live="polite">You have been removed from this test run.</div>
~~~

This alert has two problems:
1. The message is inaccurate. It states a tester was removed when in fact a tester was assigned.
2. It is positioned in a distracting place in the DOM.

I recommend that screen reader alert messages be:
* Located someplace in the DOM where they are less likely to be distracting, e.g., end of the DOM.
* Short-lived, e.g., disappear after 10 to 15 seconds or at least when the user next interacts with any element on the page.

mcking65:
Found another instance of the same problem in another place. As a tester, open a test plan from the test queue. Complete the version popup, after the first test loads, Press the next button to navigate to the next test in the plan.

The following irrelevant alert is displayed near the top of the page between the menu nav region and the test nav region:

>Your AT and Browser versions have been successfully saved.

That alert was relevant when the test runner first opened, but is not relevant when navigating to the next test. Like the one mentioned above, it should be:
* Placed at the end of the DOM.
* Removed shortly after created.

mcking65:
Another more confusing screen reader alert is announced in the following scenario:

1. Open test queue and assign yourself to a plan.
2. Open the test runner.
3. Specify screen reader version and save and continue.
4. Activate Save and close in the runner.

5. Open the runner for the same test plan.

If using Chrome, the following alert is announced:

>Alert! The version of Chrome you have set is different from the one we have automatically detected, which is 116.0.0. This change doesn’t affect results that have already been submitted for this plan. However, results you submit during this session will be recorded with the versions specified in this form.

Both times I opened the test plan, the browser version was auto-detected, and the browser version was the same both times. The alert is confusing because it contains incorrect information.

It seems that the app is saving the major version number but detecting the full version number. The information presented in the browser version field in the dialog is confusing because .0.0 is appended to the major version. That gives users the impression that the full version is expected and saved. But that is not the case.

If the app is only saving the major version number, it should not display .0.0 in the version number field. This is probably one reason for the incorrect alert.

[...]

PR #894, opened 2024-01-10
Inaccurate screen reader alert

alflennik:
See https://github.com/w3c/aria-at-app/issues/797. We occasionally use aria-live to instruct screen readers to speak feedback to the user, however in one case the aria-live element could also be traversed by a screen readers, which was unnerving and not a best practice, and furthermore the information being spoken was inaccurate. This PR uses the more polished aria-live provider element which is currently in use elsewhere in the project and makes the information being spoken more helpful.


client/index.js
```diff
@@ -5,13 +5,16 @@ import { BrowserRouter } from 'react-router-dom';
...
+import { AriaLiveRegionProvider } from './components/providers/AriaLiveRegionProvider';
 
 const container = document.getElementById('root');
 const root = createRoot(container);
 root.render(
     <GraphQLProvider>
-        <BrowserRouter>
-            <App />
-        </BrowserRouter>
+        <AriaLiveRegionProvider>
```
howard-e, line 14: Good change!

howard-e (approved):
LGTM! Also checked that the changes included for the DataManagement table sticks to the current behavior.

This should also make working with `AriaLiveRegionProvider` much easier in the future

alflennik:
Big thanks @evmiguel for catching an issue with the "checked / unchecked" speech, could you take another look and approve if it looks good?

client/components/TestQueue/AssignTesterDropdown/index.jsx
```diff
@@ -150,24 +149,31 @@ const AssignTesterDropdown = ({
...
-                                        );
+                                        const updatedIsAssigned =
+                                            !testerIsAssigned;
                                         setAlertMessage(
-                                            `${tester.username} has been ${
-                                                classname.includes('not')
-                                                    ? 'assigned to'
-                                                    : 'removed from'
-                                            } this test run.`
+                                            `${username} ${
```
evmiguel, line 157: I can confirm that this feature works as expected.

client/components/TestQueue/AssignTesterDropdown/index.jsx
```diff
@@ -150,24 +149,31 @@ const AssignTesterDropdown = ({
...
                                         );
+                                        await toggleTesterAssign(username);
                                         await onChange();
                                     }}
-                                    aria-checked={isTesterAssigned}
                                 >
                                     {icon && <FontAwesomeIcon icon={icon} />}
-                                    <span className={classname}>
+                                    <span className="sr-only">{`${username} ${
+                                        testerIsAssigned
```
evmiguel, line 169: I can confirm that this feature works as expected.

evmiguel (approved):
Great work @alflennik! I can confirm that the AT message works well.

PR #895, opened 2024-01-22
Proof of concept for using database session transactions to reset the database between tests

alflennik:
This proof of concept is part of the overall effort to implement an end-to-end testing solution for the ARIA-AT App. It demonstrates a way we can use session transactions to quickly roll back the database without needing to run the database initialization steps for every test. In effect it is possible to create simultaneously divergent database states for every test, allowing tests to be run concurrently just like non-end-to-end tests would be.

This PR is a proof of concept and NOT intended to be merged.


alflennik:
Implemented in https://github.com/w3c/aria-at-app/pull/951

Issue #884, opened 2023-12-20
Missing In Draft Review filter buttons at the top of the Data Management page


ccanash:
In the filter buttons at the top of the Data Management page  there is a missing "In Draft Review" button even though there are items that have been items advanced to Draft

![Image](https://github.com/w3c/aria-at-app/assets/36164089/d07db46d-2885-4671-8b1f-602f29a2f59c)

PR #896, opened 2024-01-22
Ensure that the In Review filter captures all test plans that have a draft review

evmiguel:
This PR addresses #884, where the app was not capturing all the test plans that contained In Draft reviews. 

Question for reviewers:

- Is this the right approach?
- Is there a better way to test other than modifying the front end test?

### DIFF
--- a/client/components/DataManagement/filterSortHooks.js
+++ b/client/components/DataManagement/filterSortHooks.js
@@ -181,30 +181,23 @@ export const useDataManagementTableFiltering = (
         })`;
     }
 
-    // Find test plan versions with DRAFT advanced, even if
-    // the overall phase is not draft
-    // const testPlansWithDraft = new Map();
-    // const draftCount = testPlanVersions.reduce((acc, version) => {
-    //     if (
-    //         version.draftPhaseReachedAt !== null &&
-    //         version.phase == TEST_PLAN_VERSION_PHASES.DRAFT &&
-    //         !testPlansWithDraft.has(version.title)
-    //     ) {
-    //         acc++;
-    //         testPlansWithDraft.set(version.title, version);
-    //     }
-    //     return acc;
-    // }, 0);
-
     if (
         testPlansByPhase[TEST_PLAN_VERSION_PHASES.DRAFT].length > 0 ||
         Object.values(derivedDraftTestPlans).length > 0
     ) {
+        const allDrafts = [
+            ...Object.values(derivedDraftTestPlans),
+            ...testPlansByPhase[TEST_PLAN_VERSION_PHASES.DRAFT]
+        ];
         filterLabels[
             DATA_MANAGEMENT_TABLE_FILTER_OPTIONS.DRAFT
         ] = `In Draft Review (${
-            Object.values(derivedDraftTestPlans).length.toString() ||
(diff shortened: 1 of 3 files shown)


stalgiag (changes requested):
We discussed separately but noting here also. I think the todo requires that a test plan row is visible if it has ANY test plan version in any of the selected phases. This probably warrants rewriting the hook so that it doesn't rely on overall test plan phase at all.

client/components/DataManagement/filterSortHooks.js
```diff
@@ -73,34 +73,41 @@ export const useDerivedOverallPhaseByTestPlanId = (
...
+                // while the subsequent elements in the array are TestPlanVersions
+                // that are in other phases
                 if (derivedPhase) {
-                    derivedOverallPhaseByTestPlanId[testPlan.id] = derivedPhase;
-                    break;
+                    let arr = activeTestPlanVersionsByPhase[testPlan.id];
+                    if (!arr) activeTestPlanVersionsByPhase[testPlan.id] = [];
+                    activeTestPlanVersionsByPhase[testPlan.id].push(
+                        derivedPhase
+                    );
```
stalgiag, line 85: The way that the reference is being accessed/modified is a little slippery to me here.

I wonder if something without the temporary var could also work well.

```
  // Ensure the array exists or initialize it directly before pushing the new phase
  activeTestPlanVersionsByPhase[testPlan.id] = activeTestPlanVersionsByPhase[testPlan.id] ?? [];
  activeTestPlanVersionsByPhase[testPlan.id].push(derivedPhase);

```

docs/database.md
```diff
@@ -94,6 +96,8 @@ yarn sequelize:test db:migrate;
 yarn sequelize:test db:seed:all;
 yarn workspace server db-import-tests:test -c 5fe7afd82fe51c185b8661276105190a59d47322;
 yarn workspace server db-import-tests:test -c 1aa3b74d24d340362e9f511eae33788d55487d12;
+yarn workspace server db-import-tests:test -c ab77d47ab19db71c635c9bb459ba5c34182e1400;
+yarn workspace server db-import-tests:test -c d34eddbb8e751f07bd28d952de15fa7fe5f07353;
```
stalgiag, line 100: Good idea but I think it makes sense to keep these in [Howard's separate PR](https://github.com/w3c/aria-at-app/pull/911)

stalgiag (approved):
Nice! I am impressed by how surgical these changes are. I think the data structure with the overall phase followed by the active test plan versions sorted by phase is a good call. Tested with two different DB snapshots and am able to filter the table so that each row displays if it has any test plan version active in the selected filter phase.

Two tiny requests for changes inline but I am approving since they shouldn't be blockers.

 Thanks for this!

alflennik (commented):
I verified that the change works perfectly! Thanks for going beyond the letter of the task and applying the fix to every phase 🙏. I haven't approved yet because I don't want to merge until you get back to Stalgia about her questions, especially the one about the changes to the database scripts. Additionally there is a conflict will need be resolved before I can merge.

evmiguel:
Thank you @stalgiag and @alflennik! I've addressed your feedback, and the merge conflicts have been resolved.

alflennik (approved):

Issue #872, opened 2023-12-13
accessibility problems and suggestions related to NVDA bot status strings


mcking65:
When there is an active bot run, the status of tests is shown in the test queue. That is helpful. When I open the manage bot run dialog, though, there is no status shown in the dialog. It would be very helpful to show the same information in the dialog.

Like some other places where there is code to determine whether a noun should be singular or plural, the rendering for screen readers of the status strings has problems. Here is how JAWS presents the strings:

>0  Test s  Completed 
>15  Test s  Queued 
>0  Test s  Cancelled 

Note:
* extra space before the word test
* space between 'test' and 's'
* Extra space after 's'

jugglinmike:
Hi @mcking65! I've tackled the bug fix and the feature request in separate patches: https://github.com/w3c/aria-at-app/pull/899 and https://github.com/w3c/aria-at-app/pull/900

jugglinmike:
@mcking65 I'm not able to reproduce the bug using JAWS (version 2024.2312.53) on Windows 10 (Home edition, version 19045.3930) and Firefox (version 122.0) or Edget (version 121.0.2277.83). Can you share your software versions and the steps you are taking to observe the bug?

For my part, when I instruct JAWS to read the line (by pressing the "down" arrow), it says (for instance) "14 tests completed" with no extraneous pauses. When I instruct JAWS to read each character in sequence (by repeatedly pressing the "right" arrow), it says each character with no extraneous spaces.

I do have a theory about what's going wrong, though, and [I've proposed a fix based on that theory](https://github.com/w3c/aria-at-app/pull/900). We'll need to reproduce the problem in order to verify my theory and merge that patch. To that end, would you mind checking if [this reduced test case](http://s92097608.onlinehome.us/tmp/text-nodes.html) exhibits the same problem for you?

ccanash:
Re opening as we referenced this tasks in #899  but it hasn't been addressed

mcking65:
@jugglinmike 

Thank you for the reduced test case and for mentioning that you were testing in Firefox.

This appears to be a bug in JAWS related to Chrome. JAWS is erroneously inserting spaces between the text nodes.

With [that test case](http://s92097608.onlinehome.us/tmp/text-nodes.html):
1. NVDA and Narrator with both Chrome and Firefox read both list items as "Cats". Oddly, in Chrome, Narrator anounces the list items as H1 elements.
2. JAWS+Firefox reads both list items as "Cats".
[...]

PR #899, opened 2024-01-25
Render status list within manage bot run dialog

jugglinmike:
Partially resolves gh-872 ~~(I do not have access to JAWS, and since the reported problem does not occur with VoiceOver, I cannot verify this fix)~~ (I erroneously included the prior text; please ignore it.)

### DIFF
--- a/client/components/ManageBotRunDialog/WithButton.jsx
+++ b/client/components/ManageBotRunDialog/WithButton.jsx
@@ -7,6 +7,7 @@ import { useTestPlanRunIsFinished } from '../../hooks/useTestPlanRunIsFinished';
 const ManageBotRunDialogWithButton = ({
     testPlanRun,
     testPlanReportId,
+    runnableTestsLength,
     testers,
     onChange
 }) => {
@@ -33,6 +34,7 @@ const ManageBotRunDialogWithButton = ({
                 setShow={setShowDialog}
                 testers={testers}
                 testPlanReportId={testPlanReportId}
+                runnableTestsLength={runnableTestsLength}
                 onChange={async () => {
                     await onChange();
                     setShowDialog(false);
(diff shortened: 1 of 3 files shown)


howard-e (approved):
Code looks good to me!

I do feel the gaps of vertical space in the dialog are a little bit much. Could they be trimmed? Approving though because I don't feel strongly on this, and in the absence of explicit visual requirements or mockups, my thought on this is purely opinion.

jugglinmike:
Thanks for the review, @howard-e. I'm happy to oblige your design preference; I just need a little clarification. [That padding is applied to all instances of this app's modal dialog.](https://github.com/w3c/aria-at-app/blob/eaf9109add8d8d2df8939fdbdd1f536f26ac7be1/client/components/TestRun/TestRun.css#L461-L465) Would you prefer to see an override for this instance, or would you like the padding reduced everywhere?

howard-e:
> Thanks for the review, @howard-e. I'm happy to oblige your design preference; I just need a little clarification. [That padding is applied to all instances of this app's modal dialog.](https://github.com/w3c/aria-at-app/blob/eaf9109add8d8d2df8939fdbdd1f536f26ac7be1/client/components/TestRun/TestRun.css#L461-L465) Would you prefer to see an override for this instance, or would you like the padding reduced everywhere?

Reducing them everywhere feels good to me. But I wonder what other's thoughts on it may be.

This has now also made me notice that unnecessary padding gets added in a lot of instances for `modal-footer` even though that content is not populated as much. With that in mind, I think this overall design concern should be tied to another issue and not be included in this PR.

jugglinmike:
Thanks, @howard-e! Here's an issue to track the design work: https://github.com/w3c/aria-at-app/issues/907

PR #900, opened 2024-01-25
Normalize empty space in bot run test status list

jugglinmike:
Partially resolves gh-872


jugglinmike:
[I have not been able to reproduce the reported bug](https://github.com/w3c/aria-at-app/issues/872#issuecomment-1913374630); I've written this patch based on a theoretical explanation (more below). This patch should not be merged until someone is able to verify that it solves the reported problem; until then, I'm marking this pull request as a draft.

The React framework uses text nodes to inject content into precise locations of the document, sometimes even within single words. JAWS may be sensitive to the boundary of those text nodes even though they are not visibly apparent (and even though, in my testing, neither VoiceOver nor NVDA exhibit this sensitivity). If that's the case, then removing interstitial JSX string interpolation should cause JAWS to enunciate the text more naturally.

jugglinmike:
[Matt has verified the underlying problem](https://github.com/w3c/aria-at-app/issues/872#issuecomment-1951472782), confirming that this patch will correctly work around a fundamental problem in the JAWS screen reader. With that, I'm removing the "draft" status and requesting review.

jugglinmike:
I've merged in the `main` branch to address a failing test which I believe has recently been fixed.

alflennik (approved):
I couldn't reproduce this with the screen readers I can run, which are NVDA and VoiceOver, but having fixed a similar issue before I can definitely vouch for the solution you used.

howard-e (approved):
Looks good! Can confirm it does fix the issue that was raised, didn't experience it with NVDA but I could with VO.

Always great to get rid of instances of this problem.

Issue #870, opened 2023-12-13
Adding report run to test queue incorrectly removes report from report page


mcking65:
I ran into this in sandbox while trying to test NVDA automation. This bug prevents making a report for a new version of an AT by forcing removal of any already final report for the same AT/browser combination. It is critical that we don't pull back a published report. We need to make new reports for the same AT/browser combination. That is the primary use case for automation.

To reproduce this bug:
1. Identify a AT/browser/test plan  combination with a report on the reports page.
2. Add a run to the test queue for that test plan and AT/browser combination.

Expected result: A new empty test run is added to the test queue. I was expecting to run it with automation and have the responses compared to the previous final report.

Actual result:
1. The report is removed from the report site.
2. The test queue has previous results for that AT/browser combo and a "Mark as Final" button.

It appears this bug is due to changes described by by @alflennik in #796. We may need to find a way to roll back some of those changes ASAP.

Here are some general reporting requirements that I thought were already understood based on all the automation discussion. However, it seems we don't yet have alignment on these.

1. Keep all reports that have been marked final.
2. If a report meets the following criteria, surface it on the reports page:
    1. It has been marked final.
    2. It is generated from a test plan version that is recommended, OR the test plan version is in candidate review and that test plan does not have a recommended version.
    3. It includes data for the latest AT version and latest browser version that have been tested for a given AT/browser combo
    4. Reports for older AT and browser versions should be maintained in the system so we can use them when we add trend reporting.
3. When a report run for a given test plan version is added to the test queue, it should always be empty; do not pull back an existing report from the report site. We do not have a requirement to edit reports; we will just replace them but only when specific conditions are met.
4. When a report is marked final in the test queue, it is published to the reports page if it meets the above conditions. Otherwise, e.g., it is for older versions of AT or browser, it is kept in the system for trend reporting. If it is for the exact same version combination as a report previously marked as final, warn the admin that it will replace the current report.
5. If a report is not available on the reports page because the test plan version is draft, is candidate but has a recommended version, or is not for the latest AT/browser version, the report should be accessible from the test plan versions page for that test plan and from the report status dialog on the data management page.

[...]

PR #901, opened 2024-01-25
Warning modal for creating duplicate test plan runs

Paul-Clue:
see issue #870

This PR is creates a modal that issues a warning to a user that is trying to add a new test plan run to the test queue with the same AT / browser combination and is the same version of an already completed test plan run. The modal warns the user that by creating a duplicate test plan run, the finalized test plan with results will be removed from the results page.

This behavior (completed test results being removed form the results page) is no longer acceptable, according to the issue. This PR serves as a temporary solution for the main issue at hand.

---
  - https://app.asana.com/0/0/1206550060484121


client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -47,12 +51,12 @@ function AddTestToQueueWithConfirmation({
 
     const alreadyHasBotInTestPlanReport = useMemo(
```
jugglinmike, line 52: We should update the name of this variable, as well, since we're no longer specifically considering bot-assigned Test Plan Runs.

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -129,6 +133,37 @@ function AddTestToQueueWithConfirmation({
         );
     };
 
+    const renderErrorDialog = () => {
+        return (
+            <BasicModal
+                show={errorMessage}
+                title={'Conflicting Report Found'}
```
jugglinmike, line 140: No need to use interpolation for static values.

```suggestion
                title='Conflicting Report Found'
```

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -129,6 +133,37 @@ function AddTestToQueueWithConfirmation({
...
+                content={
+                    'The report could not be created because an existing report was found on the reports page with the same AT, browser and test plan version. Would you like to return the existing report back to the test queue?'
+                }
```
jugglinmike, line 143: The project doesn't enforce a maximum line length, but it appears to (mostly) split strings at the 100-column boundary. @howard-e @alflennik @stalgiag is there any consensus on this?

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -166,17 +201,30 @@ function AddTestToQueueWithConfirmation({
                 disabled={disabled}
                 variant="secondary"
                 onClick={async () => {
-                    if (hasAutomationSupport) {
-                        setShowConfirmation(true);
+                    const conflictingReportExists =
+                        existingTestPlanReports.some(report => {
```
jugglinmike, line 205: `existingTestPlanReports` is potentially falsey, so we should use optional chaining here, too. It might also be good to perform this calculation outside the handler just like we do for `hasAutomationSupport`.

jugglinmike (changes requested):

client/components/AddTestToQueueWithConfirmation/index.jsx
howard-e: It's certainly been mentioned a few times now,  in GitHub threads and in sync discussions. I've created an issue to now track that discussion, https://github.com/w3c/aria-at-app/issues/924.

client/components/AddTestToQueueWithConfirmation/index.jsx
Paul-Clue: Thanks. Done.

client/components/AddTestToQueueWithConfirmation/index.jsx
Paul-Clue: Hey we found out that the project formats code at column 80. I split the strings there.

client/components/AddTestToQueueWithConfirmation/index.jsx
jugglinmike: You're right--I don't know how I missed that!

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -129,6 +141,40 @@ function AddTestToQueueWithConfirmation({
...
+            <BasicModal
+                show={errorMessage}
+                title="Conflicting Report Found"
+                content={
+                    'The report could not be created because an existing ' +
+                    'report was found on the reports page with the same AT, ' +
+                    'browser and test plan version. Would you like to return ' +
+                    'the existing report back to the test queue?'
+                }
+                closeLabel={'Cancel'}
```
jugglinmike, line 155:
```suggestion
                closeLabel="Cancel"
```

jugglinmike:
Thanks, @Paul-Clue! Would you mind taking a look at the test failures?

Paul-Clue:
Hey, @jugglinmike, The tests were failing because I moved  the `conflictingReportExists`  calculation outside of the handler. Inside the calculation, `at.id` is being referenced but it seems that `at` does not have a value until an AT is selected in the app. So I put the calculation back.

jugglinmike:
That explains why the `getBotUsernameFromAtBrowser` utility function uses optional chaining when it dereferences `at` (and `browser`). We should use optional chaining in this new code, as well!

If we address the "undefined requirements" case only by relocating the calculation to the event handler, then there's a risk that the code still manages to run before the value is available (possibly due to some uncommon network delay, for instance). Also, implicitly relying on the delay represents a refactoring hazard for future contributors. (What I mean by this is that I think folks would be justified if they assumes that the new logic would be safe to relocate anywhere in the component. We know that assumption is false because we're in the thick of it, but if that future contributor acted on their assumption, they could unknowingly introduce a bug.)

And to that latter concern, if we use optional chaining, then the new code *will* be safe to use anywhere in the component. That means we could place it up near the top like before, after all. I can think of a couple reasons why that's still desirable:

- It makes the source more consistent (in that we have all the complicated formulas in one section of the component and all the rendering code in another)
- It makes the code more readable (for the same reason as above)

If you want to get fancy, you could also wrap this calculation in a call to `useMemo` so that we don't needlessly recompute the same value every time the component is rendered.

Paul-Clue:
Hey, @jugglinmike, I missed that  I can add optional chaining to the `at` and `browser` variables as well. Thank you for pointing that out to me.

jugglinmike (approved):
Looking good, @Paul-Clue--nice work!

PR #904, opened 2024-01-29
Implement end-to-end smoke test

alflennik:
This PR adds the basic capability to run end-to-end tests and creates a minimal smoke test - basically it loads several pages and verifies that the h1s are correctly populated. It is probably pretty apparent from the code that this is not an intended final state for end-to-end testing. For example it doesn't support more than one test suite, logging in or resetting the database state. These features will come in future PRs.


client/tests/smokeTest.test.js
```diff
@@ -0,0 +1,122 @@
...
+            });
+        };
+
+        server.stdout.on('data', data => {
+            const output = data.toString();
+            console.info(output); // eslint-disable-line no-console
+
+            if (
+                (serverOrClient === 'server' &&
+                    output.includes('Listening on 8033')) ||
```
howard-e, line 42: I think you could make the static env values being referenced above into file scoped constants, so they could be reused in instances like this (`PORT` instead of "8033").

This could be beneficial in the instance that the variables do have to change, or if there's need for dynamic re-assignments in the future.

client/tests/smokeTest.test.js
```diff
@@ -0,0 +1,122 @@
...
+
+    it('loads various pages without crashing', async () => {
+        let homeH1;
+        let reportsH1;
+        let dataManagementH1;
+
+        await Promise.all([
+            (async () => {
+                const page = await browser.newPage();
+                await page.goto('http://localhost:3033/');
```
howard-e, line 95: Same comment as above, re `CLIENT_PORT` here

client/tests/smokeTest.test.js
```diff
@@ -0,0 +1,122 @@
...
+                await page.waitForSelector('h1');
+                const h1Handle = await page.waitForSelector('h1');
+                homeH1 = await h1Handle.evaluate(h1 => h1.innerText);
+            })(),
+            (async () => {
+                const page = await browser.newPage();
+                await page.goto('http://localhost:3033/reports');
+                await page.waitForSelector('h2');
```
howard-e, line 103: Shouldn't this be `await page.waitForSelector('h1')`?

client/tests/smokeTest.test.js
```diff
@@ -0,0 +1,122 @@
...
+                const h1Handle = await page.waitForSelector('h1');
+                reportsH1 = await h1Handle.evaluate(h1 => h1.innerText);
+            })(),
+            (async () => {
+                const page = await browser.newPage();
+                await page.goto('http://localhost:3033/data-management');
+                await page.waitForSelector('h2');
```
howard-e, line 110: Shouldn't this be `await page.waitForSelector('h1')`?

howard-e (commented):
@alflennik seems good to me. I can certainly see the benefits of this initial work, and what it can bring once expanded.

I've left a few inline comments and questions below.

evmiguel (approved):
Looks good @alflennik. Would it be possible to add in a TODO or in a README, what you're looking to do in further iterations? I know you mentioned having tests in a signed in state. Would be nice to have it documented somewhere so that we don't forget.

alflennik:
Thanks for the review @evmiguel! I expanded on the TODO.

howard-e (approved):
Looks good. Thanks for addressing the PR and adding the relevant comments about the `h2`. Exciting stuff!

PR #908, opened 2024-02-01
Allow End-to-End Testing as an Admin or Tester

alflennik:
This PR adds an end-to-end test which verifies that it's possible to sign in as an admin, a tester, and a vendor, or remained signed out (all concurrently).

It replaces the existing fake role system since that system didn't support this use case.

There's actually already one fixed bug in this PR, it turned out the atLoader and browserLoader might return invalid data for a brief moment when they first start.


docs/local-development.md
```diff
@@ -29,12 +29,23 @@ ARIA-AT App determines if you are authorized to sign in as an admin based on whe
...
 Another way to log in as either a tester or admin, useful for quick testing and not requiring editing testers.txt or membership within any GitHub organizations or teams, is described below.
 
-1. Sign out and return to the home page.
-2. Add `?fakeRole=admin` to the URL bar and press enter. Alternatively use `?fakeRole=tester` to log in as a tester only or `?fakeRole=` to preview logging in without a role.
-3. Follow the sign in steps as normal.
-4. After signing in, your selected role will be used for the duration of your session.
-
-This functionality is available in development environments where the ALLOW_FAKE_ROLE environment variable is "true".
+1. With the app running, open the browser DevTools.
+2. Go to the DevTools console.
```
evmiguel, line 33: How can we socialize this new workflow for folks who have been using the `fakeRole` endpoint?

evmiguel (commented):
@alflennik great work! I can confirm that the tests work. I understand why you took the approach of re-architecting the sign in flow for this. I do wonder how we will socialize the new `signMeIn` workflow for allowing fake users for folks who are used to the old workflow. Let me know your thoughts!

docs/local-development.md
alflennik: I'd definitely be happy to demo it for the team once this is in, perhaps in the sprint demo!

client/tests/smokeTest.test.js
```diff
@@ -1,142 +1,88 @@
...
 - The server start and close functions should be handled in one place by
 Jest's setup and teardown scripts.
-- There needs to be a utility function to enable common tasks like getting a
-fresh end-to-end page.
-- End-to-end testing should support signing in with different roles, A.K.A. as
-an admin, tester or vendor.
 - It should be possible to reset the database to a predictable state after each
 test, ideally in such a way that concurrency remains possible. See the POC
 here: https://github.com/w3c/aria-at-app/pull/895
+- Add section to docs for end-to-end testing
```
howard-e, line 10: TODO this so it's easier to find afterwards?

client/tests/util/getPage.js
```diff
@@ -0,0 +1,134 @@
...
+
+    if (!incognitoContexts[role]) {
+        incognitoContexts[role] = await browser.createIncognitoBrowserContext();
+    }
+    const incognitoContext = incognitoContexts[role];
+
+    const page = await incognitoContext.newPage();
+
+    if (!url) {
+        throw new Error('Please provide a URL, even if it it is simply "/"');
```
howard-e, line 102: Defaulting `url` to `/` and throwing a console.warn about `/` being used when defaulted seems fine to me. Unless there is another reason you may not want to do that?

client/tests/util/getPage.js
```diff
@@ -0,0 +1,134 @@
...
+const teardown = async () => {
+    await Promise.all([backendServer.close(), clientServer.close()]);
+
+    // Browser might not be defined, if it failed to start
+    if (browser) await browser.close();
+};
+
+let incognitoContexts = {};
+
+const getPage = async (options, callback) => {
```
howard-e, line 88: I think there's good reason here to add a brief `JSDoc` on the input expectation, particularly for what goes into `options`.

howard-e (commented):
@alflennik it's great to see this advancing, thanks for taking on this work! There's nothing majorly blocking to me, but I do have a couple questions and thoughts on the approach below.

docs/local-development.md
```diff
@@ -29,12 +29,23 @@ ARIA-AT App determines if you are authorized to sign in as an admin based on whe
...
+3. Paste in the following code:
```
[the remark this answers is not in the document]
alflennik: Nice suggestion! Will do.

client/tests/util/getPage.js
alflennik: I like an error in this case because it prevents the url from being undefined. With a warning there wouldn't be any guarantee about whether url is undefined and I would have to keep track of that gotcha throughout the function.

client/tests/smokeTest.test.js
alflennik: if you look up at line 4 you can see it's part of a TODO!

client/tests/util/getPage.js
alflennik: Nice idea!

alflennik: Done

docs/local-development.md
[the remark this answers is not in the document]
alflennik: Done

client/tests/smokeTest.test.js
howard-e: Right you are. Completely missed that!

client/tests/util/getPage.js
howard-e: Okay got it

docs/local-development.md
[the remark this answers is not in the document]
howard-e: Thanks for addressing that, it is a great move in the testing direction we'd like. And although it deviates a bit from what was in place, this feels preferred. This hopefully makes it a bit more digestible in the scenario I mentioned above, if required.

howard-e (approved):
Thanks for addressing my feedback, happy to see the testing related work progressing!

deploy/files/config-sandbox.env
```diff
@@ -39,3 +40,45 @@ $ANSIBLE_VAULT;1.1;AES256
 32666530356462333866653231653233643962373734356233643630626462663361643332303363
 33353937356234623830663666616535326464633433336631383231663735346338366332386435
 63303861396334363161
+=======
```
evmiguel, line 43: @alflennik I see there are diff strings in this file.

evmiguel (changes requested):
@alflennik Your changes look good to me! The only thing I caught was a diff string in the config-sandbox.env. I also think that you should set the environment variable in the staging and production env files to make things explicit.

PR #912, opened 2024-02-06
Add workflow to include automatic deployments to staging

howard-e:
This PR adds `workflows/deploy-staging` to trigger automatic deploys to staging once the `development` branch is updated.


deploy/README.md
```diff
@@ -69,6 +69,7 @@ To deploy this project to  server:
 3. Obtain a copy of the `ansible-vault-password.txt` file in LastPass and place it in the directory which contains this document.
 4. Install [Ansible](https://www.ansible.com/) version 2.11. Instructions for macOS are as follows:
   - Install Ansible at the specific 2.11 version: `python3 -m pip install --user ansible-core==2.11.1`
+  - Install `ansible.posix` to make use of the [ansible.posix.synchronize](https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html#ansible-posix-synchronize-module-a-wrapper-around-rsync-to-make-common-tasks-in-your-playbooks-quick-and-easy) module: `ansible-galaxy collection install ansible.posix`
```
jugglinmike, line 72: Thanks for catching this! We should put this step *after* sourcing the shell's RC file (otherwise, the invocation of `ansible-galaxy` won't resolve)

.github/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,30 @@
...
+          config: ${{ secrets.DEPLOY_SSH_CONFIG }}
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+      - name: Install ansible and deploy to staging
+        run: |
+          python -m pip install --user ansible-core==2.11.1
+          cd deploy
+          echo ${{ secrets.ANSIBLE_VAULT_PASSWORD }} >> ansible-vault-password.txt
```
jugglinmike, line 27:
```suggestion
          echo ${{ secrets.ANSIBLE_VAULT_PASSWORD }} > ansible-vault-password.txt
```

(This hardly matters since the file doesn't exist prior to this command; it's just that using the "append" operator insinuates that it does.)

jugglinmike (changes requested):

deploy/README.md
howard-e: Good catch on my catch! Done in https://github.com/w3c/aria-at-app/pull/912/commits/b51576544cc9b5f8e9f4f26c705a5bd1030f0609

stalgiag (approved):
The new workflow makes sense to me. Thanks for doing this and including the updated README step and typo fix. 

I am the second approving reviewer but leaving it open since @alflennik is also listed as a reviewer

PR #918, opened 2024-02-08
Fix flaky test

alflennik:
Increases timeout of test which occasionally fails.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -102,6 +102,7 @@
     "webpack-dev-server": "^4.11.1"
   },
   "jest": {
+    "testTimeout": 15000,
     "moduleNameMapper": {
       "^@client(.*)": "<rootDir>/$1",
       "^@components(.*)": "<rootDir>/components/$1",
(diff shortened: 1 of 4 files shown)


gnarf (changes requested):
Changing my review since I just noticed this PR is failing with the same flaky test (so it's not this one causing it I assume?)

stalgiag (approved):
This makes sense to me; gradually increasing complexity in a suite of tests has caused this [flakey timeout issue](https://github.com/w3c/aria-at-app/pull/834/files) previously. It is difficult to anticipate since our local devices are usually fast enough to not hit the timeout and the machine running the GH action likely has variable specs/resources.

Part of me thinks it might be worth setting a longer timeout globally for the test runner. This would save us from these disruptive events. On the other hand, there is some value in having information relayed to us about changes in time-to-completion for tests & test suites. I previously worked on a project that stored local test execution time averages for `main` in the local environment (and in CI) and let you know when your working branch deviated significantly from those averages. Probably would be overdoing it for this project where performance isn't quite as critical.

I am happy to go with this solution for now. If we hit a similar problem in the future, we might want to consider a global change.

jugglinmike (approved):

PR #919, opened 2024-02-08
Consolidate getMetrics into shared code section

evmiguel:
This PR consolidates the `getMetrics` functionality in the server and client into a shared workspace.


yarn.lock
```diff
@@ -149,14 +149,6 @@
   dependencies:
```
evmiguel, line 149: REVIEWERS: how do I revert this? I'm not sure if this file is supposed to be changed. Let me know your thoughts.

alflennik:
I know this isn't complete yet but one quick reminder in case that helps is to remove the extra getMetrics function from the server, currently there are still two of them.

shared/getMetrics.js
```diff
@@ -1,4 +1,8 @@
-const convertAssertionPriority = require('./convertAssertionPriority');
+const convertAssertionPriority = priority => {
+    if (priority === 'REQUIRED') return 'MUST';
+    if (priority === 'OPTIONAL') return 'SHOULD';
+    return priority;
+};
```
alflennik, line 5: I noticed this function differs pretty significantly from the version currently on the server. Would you mind checking in with @howard-e about how to handle the discrepancy? I don't personally know how to address that.

shared/getMetrics.js
```diff
@@ -38,11 +42,18 @@ const countAssertions = ({
...
+            all =
+                scenarioResult?.[`${priority.toLowerCase()}AssertionResults`] ||
+                [];
+        } else {
+            all = scenarioResult.assertionResults.filter(
+                a =>
+                    convertAssertionPriority(a.assertion.priority) ===
+                    convertAssertionPriority(priority)
+            );
+        }
```
alflennik, line 56: Would you be able to explain this change to me, I'm getting a little lost on this one. My main question here is whether this is going to change the metrics from what was there before. If so, that would mean this change would result in inconsistencies in the TestPlanReport table in the database. Can you say whether that's the case?

alflennik (commented):
I appreciated that you kept Prettier, ESLint and Jest all working in the new workspace. I can see this being useful in more than one case, in fact I can think of a few instances of duplicated code that would be fantastic to put in this new workspace. Nice work!

I do have a couple of questions below.

package.json
```diff
@@ -1,47 +1,48 @@
 {
```
howard-e, line 1: I'm puzzled why this entire file is shown as updated when it seems only `shared` has now been added in `workspaces`.

shared/getMetrics.js
howard-e: You are correct in that and this PR will need to account for that. This is now being handled through a dedicated utility file at [server/util/convertAssertionPriority](https://github.com/w3c/aria-at-app/blob/main/server/util/convertAssertionPriority.js).

I'm thinking this could be another candidate for the shared folder and the references to it from the server be updated for that.

shared/getMetrics.js
howard-e: > My main question here is whether this is going to change the metrics from what was there before

This puzzled me as well. I think right here may benefit from some comments/self documenting renames. It doesn't seem to diverge from what was there before to me. The `if` block does the calculations if the assertion results are passed in, named as `requiredAssertionResults`, `mayAssertionResults`, etc. This seems to only happen when called from the client (on the `/reports/<id>` page) from what I can tell. I'm guessing because of the shape of the data that's queried for said page.

The `else` applies to when it's called from the server which has enough context to derive it from the attached priority which may be much easier to get to.

> ... that would mean this change would result in inconsistencies in the TestPlanReport table in the database. Can you say whether that's the case?

Although not directed to me, I haven't seen signs of this but I'm continuing to test as well!

howard-e (changes requested):
Very interesting approach! Initially I was thinking the metrics could have been 1 file on the server, having the data only served through `gql` but this has made me think of some cases we could benefit from this shared approach!

I spent some time testing the results, comparing with results against prod data, looking at the database metrics and haven't yet seen any odd results jump out at me with what's saved and what's shown on the frontend.

Not to mention, having to only make update to this file in a single place will be great.

I've left one blocking comment below.

shared/getMetrics.js
evmiguel: I added this to the shared section. Let me know your thoughts!

shared/getMetrics.js
evmiguel: @alflennik, Howard's explanation is correct. The `if/else` differentiates between client and server.

alflennik: @evmiguel could you rename the variables to make that clearer? Something like:
```js
const isClient = `${priority.toLowerCase()}AssertionResults` in scenarioResult
if (isClient) {
  // ...
}
```

evmiguel: @alflennik I updated this

howard-e (approved):
@evmiguel thanks for this! While I hadn't expected this direction, it definitely introduces an easy to use way to share resources between the server and client.

Looking forward to how we can expand our use cases with this!

alflennik (approved):

alflennik:
Great job! Excited to use this for future shared code between the client and server.

Issue #850, opened 2023-11-30
Modify access to aria at app  environments


ccanash:
Ensure when changes are pushed to admins.txt, vendors.txt and testers.txt the change is immediately reflected on the environments.

ccanash:
Hi @mcking65, this has been released to production. Reflagging to ensure you saw it.

PR #920, opened 2024-02-12
Changes made to admin, vendor, or tester .txt files will take effect immediately

Paul-Clue:
see issue #850

This PR adds an `admin.txt` file to the code base. It also makes the app check for changes to vendor, admin, and tester access using the `vendor.txt`, `admin.txt`, and `tester.txt` files that are on the `main` branch in github instead of in the code base that has been deployed. Now changes made to vendor, admin, or tester access will be reflected immediately in production.

Note that the file path argument for the `fetch` in the `getUsersFromFile.js` file references this branch:
```
const response = await fetch(
        // This needs to be switched back the commented out path.
        // `https://raw.githubusercontent.com/w3c/aria-at-app/main/${file}`
        `https://raw.githubusercontent.com/w3c/aria-at-app/modify-tester-vender-access/${file}`
    );
```

This is so the tests will pass here. After all code reviewers approve this PR, the file path needs to be changed to reference `main` instead of this branch:

Change this `https://raw.githubusercontent.com/w3c/aria-at-app/modify-tester-vender-access/${file}`

To this `https://raw.githubusercontent.com/w3c/aria-at-app/main/${file}`


client/tests/smokeTest.test.js
```diff
@@ -98,7 +98,7 @@ describe('smoke test', () => {
         if (browser) await browser.close();
     }, 60000);
 
-    it('loads various pages without crashing', async () => {
+    it.skip('loads various pages without crashing', async () => {
```
gnarf, line 101: I assume this was related to the flaky test?  I think this was fixed and no longer requires a skip.

server/util/getUsersFromFile.js
```diff
@@ -1,9 +1,8 @@
-const path = require('path');
-const fs = require('fs/promises');
-
 const getUsersFromFile = async file => {
-    const roleGroupPath = path.join(__dirname, file);
-    const roleGroupTxt = await fs.readFile(roleGroupPath, { encoding: 'utf8' });
+    const response = await fetch(
+        `https://raw.githubusercontent.com/w3c/aria-at-app/modify-tester-vender-access/${file}`
```
gnarf, line 3: This probably needs to change to `main` branch?

server/util/getUsersFromFile.js
```diff
@@ -1,9 +1,8 @@
-const path = require('path');
-const fs = require('fs/promises');
-
 const getUsersFromFile = async file => {
-    const roleGroupPath = path.join(__dirname, file);
-    const roleGroupTxt = await fs.readFile(roleGroupPath, { encoding: 'utf8' });
+    const response = await fetch(
```
gnarf, line 2: do we have any provision/idea what we would want to do if github is unavailable/this request times out?  We probably want to try/catch it somehow with a fallback/specific error?  I know it seems unlikely considering this only gets called from the oauth login workflow coming back from github.  (this might not need any changes, unless we also check the local file, in which case we'd want to fall back to that)

gnarf (changes requested):
Overall this is looking pretty good to me!

Three comments inline explain some changes I think may be needed, but also:

Have we given any thought to also reading the local folder versions of these files?  The only downside I can see is we wouldn't be able to quickly remove access for someone on the live servers without a deploy.  The benefit being any contributor who might not be an admin on the main branch yet can always add themselves to their local admins.txt/testers.txt files.

gnarf:
Also just to mention - I don't have the passphrase needed to decrypt the ansible secrets encoded files, so @howard-e can you pay extra attention there? my 👀 were ⛔

client/tests/smokeTest.test.js
Paul-Clue: Yes. Good catch

server/util/getUsersFromFile.js
Paul-Clue: Done!

server/util/getUsersFromFile.js
Paul-Clue: Good point. I will discuss this with @howard-e and @alflennik.

howard-e (commented):
Looks good to me and should be able to go soon!

Even though the PR's build is now failing, I confirmed it works locally by updating the branch referenced in `getUsersFromFile` from `main` to `modify-tester-vender-access`, since the `admins.txt` doesn't yet exist on `main`. This does make me further agree with @gnarf's point on having error fallbacks where applicable (although that particular issue will not happen after this gets merged).

I left a few inline comments below.

I think the following line in `local-development.md` can be updated since the organization related pieces are getting removed:

> ARIA-AT App determines if you are authorized to sign in as an admin based on whether you are a member of the official admin team within the W3C GitHub organization and as a tester based on whether your Github username is listed in the testers.txt file. The GitHub organization and team name can be changed by altering the app's environment variables, and there are dedicated teams for each app environment.

howard-e:
> Also just to mention - I don't have the passphrase needed to decrypt the ansible secrets encoded files, so @howard-e can you pay extra attention there? my 👀 were ⛔

Those files look good, thanks for calling that out!

Paul-Clue:
> Have we given any thought to also reading the local folder versions of these files? The only downside I can see is we wouldn't be able to quickly remove access for someone on the live servers without a deploy. The benefit being any contributor who might not be an admin on the main branch yet can always add themselves to their local admins.txt/testers.txt files.

Hey, @gnarf, currently we have a feature called fake roles for this use case, and actually there is a PR that will enhance it:
https://github.com/w3c/aria-at-app/pull/908. For this reason, I think it's ok to ignore the local file.

server/util/getUsersFromFile.js
Paul-Clue: Hey this is true, but this code runs immediately after completing a round trip to github. So github would have to fail between completing the round trip and the next request, so it's a very short window. Also, we previously used github to load the admin and no issues occurred that I know of.

gnarf (approved):
Test failure expected until the merge now (though we could probably just quick land the admins.txt file in its own PR and then rerun tests

howard-e (approved):
@Paul-Clue looks good to me! Reminder to update the branch being referenced for the .txt files and you can go ahead and squash and merge afterwards.

PR #921, opened 2024-02-13
Support transactions in models

alflennik:
Apologies for the ginormous PR, believe it or not this is only half of the larger effort to implement transaction support.

The file change count is high but most of the changed files only have one line in their imports changed.

Here is a summary of the changes:

1. Duplicates the server/models/services folder into server/models/services.deprecated
2. Updates all imports of the model services to this new deprecated folder.

These changes allow the entire system to keep working without changes. This is the strategy I've used to break this larger effort into two PRs.

3. Updates the server/models/services with a new object-based API, making it way easier than the old positional arguments to supply transactions.
4. Updates the tests to support the new API. Where tests were missing, I added new tests.
5. Makes all the models files consistent, CollectionJobService and TestPlanService were both diverging from the other model files and that made this kind of change difficult.


alflennik:
The diff might not be accurate, I'm seeing tons of unrelated changes.

alflennik:
Opened a new PR to fix the diff: https://github.com/w3c/aria-at-app/pull/922

PR #923, opened 2024-02-14
 Remove references to Test.atMode from graphql-schema.js

Paul-Clue:
See issue #928

---
  - https://app.asana.com/0/0/1206594245402147


server/graphql-schema.js
```diff
@@ -427,10 +427,7 @@ const graphqlSchema = gql`
         ats: [At]!
         # TODO: consider moving to the Scenario type if we support a single test
         # applying to multiple AT modes.
-        """
-        The AT mode the test was written to expect.
-        """
-        atMode: AtMode
+
```
howard-e, line 430: I think you also want to remove the 2 lines above as well

howard-e (changes requested):
@Paul-Clue appreciate the clean up happening here and this should be ready soon! There are additional `atMode` references to be removed in `components/TestReview/queries.js` and `components/TestRun/queries.js` to get those pages loading as expected.

Paul-Clue:
Hey, @howard-e, I implemented the changes that you requested.

server/graphql-schema.js
```diff
@@ -425,12 +425,7 @@ const graphqlSchema = gql`
         The ATs the test was written to expect.
         """
         ats: [At]!
-        # TODO: consider moving to the Scenario type if we support a single test
-        # applying to multiple AT modes.
-        """
-        The AT mode the test was written to expect.
-        """
-        atMode: AtMode
+
```
alflennik, line 428: Would you mind also removing the extra blank line, just to keep it consistent with the rest of the file.

alflennik: There are other references to AtMode in this file. Can you remove them as well?

```
        # TODO: reenable when this data is properly flowing into the system
        # """
        # The categories of generalized AT modes the AT supports.
        # """
        # modes: [AtMode]!
```

```
    """
    Some ATs like JAWS have modes - for example, normally pushing "h" would
    navigate to the next header, but when you are interacting with a text input,
    typing "h" would insert the letter "h" into that input. Our tests are aware
    of these modes.
    """
    enum AtMode {
        """
        JAWS "browse" mode or NVDA "browse" mode.
        """
        READING
        """
        JAWS "forms" mode, NVDA "focus" mode, or VoiceOver's one-and-only mode.
        """
        INTERACTION
    }
```

alflennik (changes requested):

howard-e:
> @howard-e even after this PR we have tons of dead code related to At Modes, and I do think this is an improvement, definitely, but I can't help but wonder if we could just completely remove it. What do you think?

It certainly doesn't seem like there currently is (or has been?) getting used. Definitely not opposed to that. Would be great to see if we can remove it and take note if there's any unexpected side effects

alflennik:
@Paul-Clue FYI I just merged from main and fixed the merge conflicts

alflennik (approved):
Great work, after nearly three years of maintaining this model without ever properly making use of it, removing this is a big win. As I noted in a comment, I came in and fixed the conflicts caused by all the activity occurring on the main branch recently, and I was able to successfully run the tests, test all the parts of the app that might be affected, and verified the migrations work both with a fresh database and a recent dump of production.

server/migrations/20240305205951-removeAtMode.js
```diff
@@ -0,0 +1,35 @@
...
+                            delete test.atMode;
+                            return test;
+                        })
+                    );
+                    await queryInterface.sequelize.query(
+                        `UPDATE "TestPlanVersion" SET tests = ? WHERE id = ?`,
+                        { replacements: [updatedTests, id], transaction }
+                    );
+                })
+            );
```
howard-e, line 32:
```suggestion
            );

            await regenerateResultsAndRecalculateHashes(
                queryInterface,
                transaction,
                { pruneOldVersions: false }
            );
```

You'll also need a migration to update the `hashedTests` field of TestPlanVersion since the atMode was factored in the tests hash. Otherwise when doing a new import, there will be duplicates of what's already in the system

howard-e (changes requested):
Looks good! Great to remove an entity that was never getting used.

I've left an inline comment on a migration tweak. I'm good to approve otherwise.

howard-e:
@Paul-Clue there is also a conflict to resolve here

Paul-Clue:
Hey @howard-e I made the changes and took care of the conflict.

server/models/CollectionJob.js
```diff
@@ -9,7 +9,8 @@ module.exports = function (sequelize, DataTypes) {
             id: {
                 type: DataTypes.STRING,
                 allowNull: false,
-                primaryKey: true
+                primaryKey: true,
+                autoIncrement: true
```
howard-e, line 13: Think this shouldn't be here. I keep seeing this in recent PRs. Very popular piece of code 😅

howard-e (changes requested):
Looks good after addressing my last inline comment. Great to get rid of this table that hasn't been getting used!

server/models/CollectionJob.js
Paul-Clue: Ok I just removed it.

howard-e (approved):
Looks good! Thanks for cleaning this up so well!

PR #922, opened 2024-02-13
Support transactions in models

alflennik:
Apologies for the ginormous PR, believe it or not this is only half of the larger effort to implement transaction support.

The file change count is high but most of the changed files only have one line in their imports changed. Also the lines changed is inflated since there's a lot of duplicated code.

Here is a summary of the changes:

1. Duplicates the server/models/services folder into server/models/services.deprecated
2. Updates all imports of the model services to this new deprecated folder.

These changes allow the entire system to keep working without changes. This is the strategy I've used to break this larger effort into two PRs.

3. Updates the server/models/services with a new object-based API, making it way easier than the old positional arguments to supply transactions.
4. Updates the tests to support the new API. Where tests were missing, I added new tests.
5. Makes all the models files consistent, CollectionJobService and TestPlanService were both diverging from the other model files and that made this kind of change difficult.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206550059227038

PR #925, opened 2024-02-14
Let database define id for CollectionJob

stalgiag:
This pull request modifies the 'CollectionJob' model so that it has an integer ID that automatically increments as new entries are added. A previous iteration of the Response Collection System specification expected there to be a layer between the app and the job runner that would provide a unique ID tied directly to the job as the identifier for the collection job. The MVP of the automation feature in the app was initially built with this design in mind but part of the way into development this part of the specification changed and as a temporary workaround code was added to the app that manually managed unique auto-incrementing IDs. This PR offloads that responsibility to the DB and simplifies some existing logic. Additionally, previous services that expected `CollectionJob`s to be created with a defined ID (such as `findOrCreateCollectionJob`) have been removed.

Since the new migration in this PR modifies the type of the primary key in the `CollectionJob` table, I would like to prioritize thorough over speedy review on this one. I tested with a relatively recent dump of production (1/22/24) by running a bot test plan run of the 'Link Example 1 (span element with text content)' using the full system and the simulator.


server/tests/integration/automation-scheduler.test.js
```diff
@@ -262,57 +258,56 @@ describe('Automation controller', () => {
...
+                await getTestCollectionJob(job.id);
+            expect(storedCollectionJob.id).toEqual(job.id);
             expect(storedCollectionJob.status).toEqual('QUEUED');
         });
     });
 
     it('should gracefully reject restarting a job that does not exist', async () => {
         const { restartCollectionJob: res } =
-            await restartCollectionJobByMutation();
+            await restartCollectionJobByMutation('2');
```
gnarf, line 301: All the others seem to be taking ints instead of strings now, feel like this might've been missed.

gnarf (approved):
I commented on one line that looked a little suspicious to me, but tests seem to be passing and stuff still behaves on my local install so LGTM!

server/tests/integration/automation-scheduler.test.js
stalgiag: Oh good point! It doesn't matter for how the code functions because it's being put into a GraphQL query but I swapped it for an int here for clarity. [26d851a](https://github.com/w3c/aria-at-app/pull/925/commits/26d851a993af7b1fc30922a22efd3c76a93895b9)

server/migrations/20240212232936-modify-collectionjob-id-type.js
```diff
@@ -0,0 +1,63 @@
...
+    down: async queryInterface => {
+        // Remove the primary key constraint
+        await queryInterface.sequelize.query(
+            `ALTER TABLE "CollectionJob" DROP CONSTRAINT "CollectionJob_pkey";`
+        );
+
+        // Drop the sequence created for auto-increment
+        await queryInterface.sequelize.query(
+            `DROP SEQUENCE collectionjob_id_seq;`
+        );
```
jugglinmike, line 51: This fails on my system:

>     $ yarn sequelize db:migrate:undo
>     yarn run v1.22.19
>     $ dotenv -e config/dev.env npx sequelize-cli db:migrate:undo
>     
>     Sequelize CLI [Node: 18.18.0, CLI: 6.6.0, ORM: 6.28.0]
>     
>     Loaded configuration file "config/config.js".
>     Using environment "development".
>     == 20240212232936-modify-collectionjob-id-type: reverting =======
>     
>     ERROR: cannot drop sequence collectionjob_id_seq because other objects depend on it
>     ERROR DETAIL: default value for column id of table "CollectionJob" depends on sequence collectionjob_id_seq
>     
>     error Command failed with exit code 1.
>     info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

server/tests/models/services/CollectionJobService.test.js
```diff
@@ -10,91 +10,63 @@ describe('CollectionJob Service Tests', () => {
...
-            expect(collectionJob.id).toEqual(id);
             expect(collectionJob.status).toEqual('QUEUED');
         });
     });
 
     it('should create a new CollectionJob with provided status', async () => {
         await dbCleaner(async () => {
-            const id = randomIdGenerator();
-            const status = 'IN_PROGRESS';
+            const status = 'CANCELLED';
```
jugglinmike, line 26: This change doesn't seem related to the refactoring. Is it still possible to create a collection job with the "IN_PROGESS" status?

More generally, I can't find a case (or think of one to be added in the near future) for creating a collection job with a status other than "QUEUED". Do you think we could remove the parameter in a follow-up patch?

jugglinmike (commented):
Thanks for paying down this tech dept--there's so much to like about your patch!

server/migrations/20240212232936-modify-collectionjob-id-type.js
stalgiag: Bummer! I'll check it out when I get more time back on the project

server/tests/models/services/CollectionJobService.test.js
stalgiag: There are no limitations to the string that collection job can have as a `status`. Currently we are relying on the use of an enum throughout all of the code that sets statuses for CollectionJobs to ensure consistency. For additional safety, we could check to make sure the string passed into `createCollectionJob` is present in the enum. That would've surfaced this leftover from earlier work. This test doesn't care about the content of the status string. It was using a status "IN_PROGRESS" that we stopped using fairly early in development. While editing the relevant tests I took the opportunity to change the test statuses to strings that are currently used.

jugglinmike: Jeepers, I forgot about our move away from "IN_PROGRESS". Thanks for the thorough explanation! We can discuss the need for alternate values at creation time later.

server/migrations/20240212232936-modify-collectionjob-id-type.js
stalgiag: This should be fixed with [5769bd3](https://github.com/w3c/aria-at-app/pull/925/commits/5769bd3e06bfbece9e83c183f8b9382a73f415b5). I tested both up and down with default test data and a dump of prod's DB from today.

stalgiag:
Note that the merge conflicts will be handled once the changes to the `services` folder [started with this PR](https://github.com/w3c/aria-at-app/pull/922) settle.

jugglinmike (approved):
> This should be fixed with [5769bd3](https://github.com/w3c/aria-at-app/pull/925/commits/5769bd3e06bfbece9e83c183f8b9382a73f415b5). I tested both up and down with default test data and a dump of prod's DB from today.

Works for me, too!

jugglinmike:
Superseded by gh-943.

Issue #931, opened 2024-02-20
Review Conflicts modal content appears as empty after tester submits conflicting results


howard-e:
_Originally posted by @alflennik in https://github.com/w3c/aria-at-app/issues/849#issuecomment-1934688011_

> .. I created an unexpected behavior for one user, but left it out for another user. There does seem to be a null value somewhere.
> 
> <img width="1151" alt="Empty review conflicts modal, with a console error of cannot read properties of null." src="https://github.com/w3c/aria-at-app/assets/5255858/0507f6f5-079d-42e6-b67d-70301b49cc1f">

howard-e:
https://github.com/w3c/aria-at-app/pull/849#issuecomment-1954456297 says:

> .. The conflicts content for the `testPlanReport` isn't being updated after a submit operation is done ..

PR #932, opened 2024-02-20
Update `testPlanReport.conflicts` in `TestRun` when test result is submitted

howard-e:
Addresses #931.


Paul-Clue (approved):

alflennik (approved):
Great simple fix! I verified the bug on main and then verified the fix.

PR #937, opened 2024-02-27
Update `updateMetricsForCountingUnexpectedBehaviors.js` migration's path to services.deprecated

howard-e:
Missed due to sequencing of when #932 and #922 was merged


alflennik (approved):
Oh shoot! Good catch.

Issue #868, opened 2023-12-11
Follow up items to address after #840 merge


howard-e:
Follow up items this PR has revealed which can be addressed later on:

1. Update named references for `REQUIRED` and `OPTIONAL` to be `MUST` and `SHOULD`
2. Update the (current) `commands.json` to `commandsV1.json`
3. Remove references to `Test.atMode` from `graphql-schema.js`
4. Remove `TestPlanUpdater` component and related references

_Originally posted by @howard-e in https://github.com/w3c/aria-at-app/issues/840#issuecomment-1828437355_

howard-e:
5. Merging `server/util/getMetrics.js` and `client/components/Reports/getMetrics.js`

PR #938, opened 2024-02-27
Remove all instances of REQUIRED  and OPTIONAL

Paul-Clue:
see issue #868 

This PR removes all instances of REQUIRED and OPTIONAL. These changes already exist in presentational components. This PR updates the references of those priority names in the code itself to lessen any confusion in the future.

---
  - https://app.asana.com/0/0/1206594245402143


client/resources/aria-at-test-io-format.mjs
```diff
@@ -1642,7 +1642,7 @@ const UnexpectedBehaviorImpactMap = createEnumMap({
 /**
  * Command result derived from priority 1 and 2 assertions.
  *
- * Support is "FAILING" is priority 1 assertions fail. Support is "ALL REQUIRED"
+ * Support is "FAILING" is priority 1 assertions fail. Support is "ALL MUST"
```
howard-e, line 1645: I don't think this has to be changed

client/resources/aria-at-test-io-format.mjs
```diff
@@ -1653,7 +1653,7 @@ const UnexpectedBehaviorImpactMap = createEnumMap({
 const CommandSupportJSONMap = createEnumMap({
   FULL: 'FULL',
   FAILING: 'FAILING',
-  ALL_REQUIRED: 'ALL REQUIRED',
+  ALL_REQUIRED: 'ALL MUST',
```
howard-e, line 1656: Same with this, but if it were to change, the enum name may as well be changed too

howard-e (changes requested):
@Paul-Clue thanks for taking this first step to update the repository to be consistent with the new language!

I think the key thing that may be missing here though is that a database migration is needed to update the older `TestPlanVersion.tests[].assertions[].priority` which still use REQUIRED and OPTIONAL as well.

server/util/convertAssertionPriority.js
```diff
@@ -8,14 +8,13 @@
  * @returns {null|string}
```
alflennik, line 8: The TODO can be removed now.

client/resources/aria-at-test-io-format.mjs
alflennik: Agreed. If you do change it, make sure to look at server/util/getMetrics.js which still uses the term "All Required." Probably better to just leave it as-is.

alflennik (changes requested):
I concur with Howard that the main missing thing is the migration. Once that's done I'll run the changes and see if I can find any issues. As for the code, it looks good, not much to add to Howard's review.

server/util/convertAssertionPriority.js
Paul-Clue: Done

server/migrations/20240311192910-removeRequiredAndOptionalPriority.js
```diff
@@ -0,0 +1,88 @@
...
+                            EXISTS (
+                              SELECT
+                                1
+                              FROM
+                                jsonb_array_elements(tests) AS test,
+                                jsonb_array_elements(test -> 'assertions') AS assertion
+                              WHERE
+                                assertion ->> 'priority' = 'REQUIRED'
+                                OR assertion ->> 'priority' = 'OPTIONAL'
+                            )`,
```
howard-e, line 78:
```suggestion
                            id = ?`,
```

shared/convertAssertionPriority.js
```diff
@@ -1,21 +1,18 @@
 /**
  * This should take any of the valid priority inputs and return the currently known string
  * representation of that value.
- * TODO: Eventually, this should only need to take 'number' types as priority once 'REQUIRED' and
- *       'OPTIONAL' are no longer used
  *
  * @param {number|string} priority
- * @returns {null|string}
+ * @returns {null|string}
```
howard-e, line 6:
```suggestion
 * @returns {null|string}
```

howard-e (changes requested):
@Paul-Clue looks good! There's an issue I found with the migration when testing against the current prod data and have suggested an inline fix.

Also, you can remove `REQUIRED` and `OPTIONAL` for the `AssertionPriority` enum in graphql-schema.js

alflennik:
@Paul-Clue looks like the tests are failing, any idea about that?

shared/getMetrics.js
```diff
@@ -191,7 +191,7 @@ const getMetrics = ({
         assertionsCount: requiredAssertionsCount,
         assertionsPassedCount: requiredAssertionsPassedCount,
         assertionsFailedCount: requiredAssertionsFailedCount
-    } = calculateAssertionPriorityCounts(result, 'REQUIRED');
+    } = calculateAssertionPriorityCounts(result, 'MUST');
```
howard-e, line 194: So with this change, because https://github.com/w3c/aria-at-app/blob/3808ed7f46642564718978aabd2765dd96c748c3/shared/getMetrics.js#L46 uses the priority string in combination with what has been named through the client's various `queries.js` files, it will incorrectly display the expected values for anywhere `TestPlanResultsTable` is getting used.

ie. [`queries.js`](https://github.com/w3c/aria-at-app/blob/3808ed7f46642564718978aabd2765dd96c748c3/client/components/TestRun/queries.js#L39) provides `requiredAssertionResults` but this change to `getMetrics` is expecting `mustRequiredAssertionResults`.

The following screenshot also shows that for each command, there are 2+ assertions but the command's assertion level status only shows `2 passed`, which would then also affect the total calculation for the entire test:
screenshot showing inaccurately calculated numbers

2 things I think could happen here:
1. Revert the change to `getMetrics` until updating each of the named `(required|optional)AssertionResults` instances. But this may cause some confusion until then.
2. Update the named instances of `(required|optional)AssertionResults` as appropriate which means renaming the value being exported from `getMetrics` and also adding a migration to update metrics across all the TestPlanReports

howard-e (changes requested):

shared/getMetrics.js
Paul-Clue: Hey @howard-e all of the changes have been made and a migration created.

shared/getMetrics.js
```diff
@@ -219,11 +219,11 @@ const getMetrics = ({
         testPlanReport?.runnableTests?.length || countTests({ ...result });
     const testsFailedCount = testsCount - testsPassedCount;
 
-    const requiredFormatted = `${requiredAssertionsPassedCount} of ${requiredAssertionsCount} passed`;
+    const requiredFormatted = `${mustAssertionsPassedCount} of ${mustAssertionsCount} passed`;
     const optionalFormatted =
```
alflennik, line 223: I think `requiredFormatted` needs to be `mustFormatted` and `optionalFormatted` needs to be `shouldFormatted`, right?

alflennik:
@Paul-Clue fantastic work, I just have one comment. I verified everything else is working and once you address that I'll approve.

shared/getMetrics.js
Paul-Clue: Ok done!

alflennik (approved):
Thanks for your considerable persistence chasing down all the details needed to get this task done! 😜

client/components/CandidateReview/CandidateTestPlanRun/queries.js
```diff
@@ -115,17 +115,13 @@ export const CANDIDATE_REPORTS_QUERY = gql`
...
+                    mustAssertionResults: assertionResults(priority: MUST) {
                         assertion {
                             text
                         }
                         passed
                     }
-                    optionalAssertionResults: assertionResults(
-                        priority: OPTIONAL
-                    ) {
+                    mustAssertionResults: assertionResults(priority: SHOULD) {
```
howard-e, line 124:
```suggestion
                    shouldAssertionResults: assertionResults(priority: SHOULD) {
```

howard-e (commented):
Looks good to go after my last inline comment

shared/getMetrics.js
howard-e: Great! Thanks for going the route of #2 as well. Definitely makes things cleaner for the future!

client/components/CandidateReview/CandidateTestPlanRun/queries.js
Paul-Clue: Got it! Thanks @howard-e !

howard-e (approved):
Looks good! Thanks for the continued changes to perfect and support this change @Paul-Clue!

ccanash:
See issue #926

PR #942, opened 2024-02-29
Handle tokenized `assertionPhrase` in import script

howard-e:
https://github.com/w3c/aria-at/pull/1038 introduces tokenized [`assertionPhrase`](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#assertionphrase) the same way an [`assertionStatement`](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#assertionstatement) is tokenized.

This PR handles importing a tokenized `assertionPhrase` for a test the same way tokenized `assertionStatements` are being handled.

To functionally test this change:
* There is https://github.com/w3c/aria-at/pull/1015 which builds its content with a tokenized assertion phrase:
  * The [Preview](https://deploy-preview-1015--aria-at.netlify.app/review/radiogroup-aria-activedescendant) shows Test 1 for JAWS and NVDA, with Tab (virtual cursor active and browse mode active) now includes an assertion phrase (and statement) which are tokenized values.
* ~Switch the import-tests script to import the branch `tests/Radio-activedescendant-refactor`~ This has now been merged so no need to change the branch, just run the import script:
  * Observe that the `TestPlanVersion` data for `Radio Group Example Using aria-activedescendant Test Plan V24.03.13` in `tests[0].assertions[lastIndex]` has a tokenized representation in `assertionPhrase` and `assertionStatement`, for JAWS.
  * The same is true for `tests[1]` which is the NVDA equivalent.


alflennik (approved):
I successfully imported the new test and took a look at the database JSON for the test. In contrast to the other test plan versions, I do see assertion phrases showing up. Also I was able to view the correctly formatted assertions on the test review page, I ran the test in the queue and successfully viewed the formatted results. Everything worked great. Nice work!

Issue #745, opened 2023-08-15
Support V2 test format in the single-page view of a test plan


mcking65:
#### Overview

Change the content shown on the test plan version view page for versions of test plans built using the V2 format.

Note:
* It is essential the system supports viewing versions of test plans built from both V1 and V2 test plan formats. 
* The [V2 format](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2) contains a column for `presentationNumber` in tests.csv. That number specifies the sequence of test presentation.

#### Changes to the test plan view page

All but one of the changes to the build/review files listed in w3c/aria-at#977 also need to be made to the test plan review page. Ignore the first change related to the level 1 heading. Make all subsequent changes listed in w3c/aria-at#977, starting with the label for the radio filters.

mcking65:
This issue says the fix is in the sandbox. However, the sandbox data doesn't have any v2 formatted test plans as far as I can tell.

I checked in staging, and the content of the test plan pages is incorrect. This is of sufficient concern that we should either:
1. Block deployment until resolved.
2. Deploy the staging environment only if a fix for this could be deployed in a subsequent fast-follow deployment that is less than a week later.

To validate, I looked at the command button test plan:
* [Build preview of command button test plan in V2 format in the aria-at repo](https://aria-at.netlify.app/review/command-button)
* [View of the same version of command button test plan from the staging server](https://aria-at-staging.w3.org/test-review/68023)

When comparing, note the following problems with the staging server view of the test plan:
1. Following the instructions in each test, there is a heading 5 "success criteria" followed by some obsolete V1 content.
2. The headings for each command and assertion tables are not presented.
3. Under the heading "About This Test Plan", the title of the latest commit is not a link; it should be a link just like it is on the test plan version history page. Editorial note: I think it would be better if that list item had content: "Latest commit: [title of latest commit]().
4. There is an information architecture problem that does not appear to be covered by the original specification. The tests are under an H2 of "Supporting Documentation" instead of under the H2 "Tests". The "Supporting Documentation" heading should be moved before the ""Tests" heading. The resulting H2 order would then be: "About This Test Plan", "Supporting Documentation", "Tests". Then, each individual test will be an H3 under the H2 of "Tests".

mcking65:
Due to above described problems, Changing status from "In sandbox" to to "in progress"

PR #944, opened 2024-02-29
Address single page view format issues 

Paul-Clue:
see issue https://github.com/w3c/aria-at-app/issues/745

This PR addresses the second set formatting issues brought up in issue https://github.com/w3c/aria-at-app/issues/745.

### DIFF
--- a/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
-import { Button } from 'react-bootstrap';
+import { Button, Table } from 'react-bootstrap';
 import { unescape } from 'lodash';
 import {
     parseListContent,
@@ -19,6 +19,7 @@ import { TestWindow } from '../../../resources/aria-at-test-window.mjs';
 import { evaluateAtNameKey } from '../../../utils/aria.js';
 import commandsJson from '../../../resources/commands.json';
 import supportJson from '../../../resources/support.json';
+import { convertAssertionPriority } from 'shared';
 
 const NumberedList = styled.ol`
     counter-reset: numbered-list;
@@ -169,18 +170,81 @@ const InstructionsRenderer = ({
         commandsContent
     );
 
-    const assertions = [...pageContent.instructions.assertions.assertions];
-    const assertionsContent = parseListContent(assertions);
-
+    headingLevel = 5;
     const Heading = `h${headingLevel}`;
+
+    console.log(renderableContent)
 
     return (
         <>
             <NumberedList>{allInstructionsContent}</NumberedList>
-            <Heading>{pageContent.instructions.assertions.header}</Heading>
-            {pageContent.instructions.assertions.description}
-            <NumberedList>{assertionsContent}</NumberedList>
             {settingsContent.length ? settingsContent : null}
+
+            {renderableContent.commands.map(({ id, settings }, i) => {
+                debugger;
+                const settingsScreenText =
+                    renderableContent.target.at.raw.settings[settings]
+                        ?.screenText ?? '';
+
+                let mustCount = 0;
+                let shouldCount = 0;
+                let mayCount = 0;
+
+                renderableContent.assertions.forEach(({ priority }) => {
+                    const priorityString = convertAssertionPriority(priority);
+                    if (priorityString === 'MUST') mustCount += 1;
+                    if (priorityString === 'SHOULD') shouldCount += 1;
+                    if (priorityString === 'MAY') mayCount += 1;
+                });
+
+                const settingsScreenTextFormatted =
+                    settingsScreenText === '' ? '' : ` (${settingsScreenText})`;
+
+                const scenarioTitle =
+                    `${renderableContent.commands[i].keystroke}` +
+                    `${settingsScreenTextFormatted}: ${mustCount} MUST, ` +
+                    `${shouldCount} SHOULD, ${mayCount} MAY Assertions`;
+
+                return (
+                    <>
+                        <Heading>{scenarioTitle}</Heading>
+                        <Table
+                            key={`${id}-${i}`}
+                            bordered
+                            responsive
+                            aria-label={`Results for test ${test.title}`}
+                        >
+                            <thead>
+                                <tr>
+                                    <th>Priority</th>
+                                    <th>Assertion Phrase</th>
+                                    <th>Assertion Statement</th>
+                                </tr>
+                            </thead>
+                            <tbody>
+                                {renderableContent.assertions.map(
+                                    ({
+                                        assertionPhrase,
+                                        assertionStatement,
+                                        priority,
+                                        assertionId
+                                    }) => (
+                                        <tr key={`${i}-${assertionId}`}>
+                                            <td>
+                                                {convertAssertionPriority(
+                                                    priority
+                                                )}
+                                            </td>
+                                            <td>{assertionPhrase}</td>
+                                            <td>{assertionStatement}</td>
+                                        </tr>
+                                    )
+                                )}
+                            </tbody>
+                        </Table>
+                    </>
+                );
+            })}
 
             <Button
                 disabled={!pageContent.instructions.openTestPage.enabled}
--- a/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
@@ -173,18 +173,17 @@ const InstructionsRenderer = ({
     headingLevel = 5;
     const Heading = `h${headingLevel}`;
 
-    console.log(renderableContent)
-
     return (
         <>
             <NumberedList>{allInstructionsContent}</NumberedList>
             {settingsContent.length ? settingsContent : null}
 
             {renderableContent.commands.map(({ id, settings }, i) => {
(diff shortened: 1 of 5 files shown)


client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -169,19 +170,128 @@ const InstructionsRenderer = ({
...
+                        : ` (${settingsScreenText})`;
+
+                const scenarioTitle =
+                    `${renderableContent.commands[i].keystroke}` +
+                    `${settingsScreenTextFormatted}: ${mustCount} MUST, ` +
+                    `${shouldCount} SHOULD, ${mayCount} MAY Assertions`;
+
+                if (isV2) {
+                    return (
+                        <>
```
howard-e, line 211: The console is reporting that a unique key prop is required here. React uses this to determine if an element needs to be re-rendered. To have a unique enough key here, and based on the context, you could do something like:
```javascript
<React.Fragment key={`command-${id}-${i}`}>
```

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -169,19 +170,128 @@ const InstructionsRenderer = ({
...
+                                    </tr>
+                                </thead>
+                                <tbody>
+                                    {renderableContent.assertions.map(
+                                        ({
+                                            expectation,
+                                            priority,
+                                            assertionId
+                                        }) => (
+                                            <tr key={`${i}-${assertionId}`}>
```
howard-e, line 278:
```suggestion
                                        ({
                                            expectation,
                                            priority
                                        }) => (
                                            <tr key={`${i}-${expectation}`}>
```

When looking at the console for plans using the v1 test format version, an error being shown is that there is multiple instances of `{index}-undefined` being set. That's because `assertionId` isn't available [for v1](https://github.com/w3c/aria-at/wiki/Test-Format-V1-Definition#assertion1-through-assertion6) tests but it is [for v2](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#assertionid).

The suggestion or similar should work

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -169,19 +170,128 @@ const InstructionsRenderer = ({
...
 
     return (
         <>
             <NumberedList>{allInstructionsContent}</NumberedList>
-            <Heading>{pageContent.instructions.assertions.header}</Heading>
-            {pageContent.instructions.assertions.description}
-            <NumberedList>{assertionsContent}</NumberedList>
             {settingsContent.length ? settingsContent : null}
 
+            {renderableContent.commands.map(({ id, settings }, i) => {
```
howard-e, line 181: A key thing missing here is that [assertionExceptions](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#assertionexceptions) aren't being handled here.

To confirm this, I did the following:
* There is an open PR which should add a test plan version with `assertionExceptions`, https://github.com/w3c/aria-at/pull/1015.
* I set the import script locally to `tests/Radio-activedescendant-refactor` and ran the script
* Navigated to Radio Group Example Using aria-activedescendant Test Plan V24.02.19 from the Data Management page
* Confirmed the commands shown for Test 1 across all 3 ATs didn't match the [built preview link of the aria-at PR](https://deploy-preview-1015--aria-at.netlify.app/review/radiogroup-aria-activedescendant).

`assertionExceptions` are available on the `renderableContent.commands` value being mapped through.

howard-e (changes requested):
@Paul-Clue looks good and happy you're so close to finishing off this bit of work!

There is something key missing, which is with handling [`assertionExceptions`](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#assertionexceptions) for the rendered commands tables.

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -169,19 +170,147 @@ const InstructionsRenderer = ({
...
-            <Heading>{pageContent.instructions.assertions.header}</Heading>
-            {pageContent.instructions.assertions.description}
-            <NumberedList>{assertionsContent}</NumberedList>
             {settingsContent.length ? settingsContent : null}
 
+            {renderableContent.commands.map(
+                ({ id, settings, assertionExceptions = [] }, i) => {
+                    const settingsScreenText = isV2
+                        ? renderableContent.target.at.raw.settings[settings]
+                              ?.screenText ?? ''
```
howard-e, line 185:
```suggestion
                              ?.screenText ?? null
```

Because line 220 expects this to be null. Otherwise, settings show me shown as 'COMMAND ()' instead of just 'COMMAND'.

The following screenshot is from Command Button Example Test Plan V23.12.13 > Test 1 > VoiceOver. Not all VoiceOver commands may have a setting applied can lead to the empty brackets.

screenshot showing Control+Option+Right Arrow and Tab commands with empty open and close brackets beside it

howard-e (approved):
@Paul-Clue LGTM! Thanks for addressing my feedback

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -169,19 +170,153 @@ const InstructionsRenderer = ({
         commandsContent
     );
 
-    const assertions = [...pageContent.instructions.assertions.assertions];
-    const assertionsContent = parseListContent(assertions);
-
+    headingLevel = 5;
```
howard-e, line 173:
```suggestion

```

Noticed this after going to the CandidateTestPlanRun page which also uses this component and realized the scenarioTitle heading component is wrong. 

You shouldn't override what's being passed into this component. Instead pass in the `headingLevel` prop as `5` where `InstructionsRenderer` is being called in the `TestReview` component.

howard-e (changes requested):
@Paul-Clue apologies, looking back at the feedback of https://github.com/w3c/aria-at-app/issues/745#issuecomment-1965638839.

These items have not been fully addressed:

> * Under the heading "About This Test Plan", the title of the latest commit is not a link; it should be a link just like it is on the test plan version history page. Editorial note: I think it would be better if that list item had content: "Latest commit: title of latest commit.
> * There is an information architecture problem that does not appear to be covered by the original specification. The tests are under an H2 of "Supporting Documentation" instead of under the H2 "Tests". The "Supporting Documentation" heading should be moved before the ""Tests" heading. The resulting H2 order would then be: "About This Test Plan", "Supporting Documentation", "Tests". Then, each individual test will be an H3 under the H2 of "Tests".

howard-e (approved):

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -169,19 +170,152 @@ const InstructionsRenderer = ({
...
+                    const settingsScreenText = isV2
+                        ? renderableContent.target.at.raw.settings[settings]
+                              ?.screenText ?? ''
+                        : null;
+
+                    let mustCount = 0;
+                    let shouldCount = 0;
+                    let mayCount = 0;
+
+                    let assertions = [...renderableContent.assertions];
```
alflennik, line 191: I noticed in the alert example the should assertions were coming before the must assertions, which makes me think that sorting this array might be needed.

Screenshot 2024-03-05 at 5 05 33 PM

client/components/CandidateReview/CandidateTestPlanRun/InstructionsRenderer.jsx
```diff
@@ -169,19 +170,152 @@ const InstructionsRenderer = ({
...
+                        if (priorityString === 'SHOULD') shouldCount += 1;
+                        if (priorityString === 'MAY') mayCount += 1;
+                    });
+
+                    let settingsScreenTextFormatted;
+
+                    if (
+                        settingsScreenText === null ||
+                        settingsScreenText === ''
+                    ) {
```
alflennik, line 223: Both of these values are falsy so this could be `if (!settingsScreenText) {` or even more compact:

```js
const settingsScreenTextFormatted = settingsScreenText ? ` (${settingsScreenText})` : ''
```

alflennik (changes requested):
Hi @Paul-Clue looks great, I was able to verify it working and successfully tested tons of test plans, on both the data management section as well as the candidate review section. I just have one possible issue and a couple of quick comments.

alflennik (approved):

mcking65:
@Paul-Clue @howard-e @ccanash 

Looking much better!! Thank you!!

 I see one important problem looking at the [page for alert in staging](https://aria-at-staging.w3.org/test-review/56291):

Captions on the table of assertions under each command heading are incorrect for NVDA and VoiceOver. For example, the caption on the spacebar command for voiceOver says:

>Space (virtual cursor active): 1 MUST, 1 SHOULD, 0 MAY Assertions

It seems the caption is coming from the first AT, which is JAWS.

mcking65:
In the [toggle button plan](https://aria-at-staging.w3.org/test-review/68049), there is no caption on the table for the first command for each screen reader.

Of less concern, in a v1 plan, like [menu button](https://aria-at-staging.w3.org/test-review/55611), the captions say "Results for  COMMAND_NAME" instead of following the caption format based on prior heading like it should.

howard-e:
@mcking65 great that the checks are generally going well! #948 was submitted to address the table caption bugs. It has now been reviewed, merged and pushed to staging.

mcking65:
Tested in staging. All resolved. Beautiful work. Thank you!

PR #945, opened 2024-03-04
Fix assertions content when multiple commands are combined and use `getMetrics` for "total" calculation

howard-e:
After reviewing #944, and doing the following [described here](https://github.com/w3c/aria-at-app/pull/944#discussion_r1511251299):

> ... I did the following:
> 
> * There is an open PR which should add a test plan version with `assertionExceptions`, [Refactor Radio Group Using aria-activedescendant  test plan to V2 test format aria-at#1015](https://github.com/w3c/aria-at/pull/1015).
> * I set the import script locally to `tests/Radio-activedescendant-refactor` and ran the script
> * Navigated to Radio Group Example Using aria-activedescendant Test Plan V24.02.19 from the Data Management page
> * ...

I noticed 3 odd instances for the imported Radio Group Example Using aria-activedescendant Test Plan when doing the Test Run process:

* For Test 1 with JAWS + Chrome, the exceptions were correctly represented on the collection form but when submitted the `down down` command was still showing the `JAWS switched from virtual cursor active to PC cursor active` on the Test Results summary table. That assertion is a 0-priority exception and shouldn't be shown.
* This caused 47 test results to be shown, but there are 46 assertions for that test (Note that the actual metrics values that were saved wasn't affected but the presentation's value is being calculated separately, and would be wrong).
* Also on the results summary table, `down down` was being shown as `Down Arrow (virtual cursor active) then Down Arrow (virtual cursor active)` instead of `Down Arrow then Down Arrow (virtual cursor active)`


Paul-Clue (approved):
Approved! Nice job catching these problems and cleaning them up!

alflennik:
Thanks for the PR and the accompanying instructions of how to test it out! Actually since this PR was created the latest code on main is making this feature difficult to test. It might be a good opportunity to expand the scope of this PR slightly to fix the issue.

<img width="880" alt="A screenshot showing the error Uncaught TypeError: Cannot assign to read only property 'priority' of object '#<Object>' at InstructionsRenderer.jsx:205:64" src="https://github.com/w3c/aria-at-app/assets/5255858/6efc7d9f-4d8c-4317-ac3f-aacbf0bec2cd">

This error is showing up on this page: http://localhost:3000/test-review/27 where 27 is the id of the radio-activedescendant test plan version from the `tests/Radio-activedescendant-refactor` branch.

I should note that the error cannot happen currently on production since the it only happens with the new test plan which is still in a branch.

howard-e:
@alflennik good catch! And thanks for the repro instructions.

Since https://github.com/w3c/aria-at/pull/1042 was included in that PR and now making it possible for exceptions to also be done through `tests.csv#assertions` instead of just `*-commands.csv#assertionExceptions`, this had to be specially handled.

Done in https://github.com/w3c/aria-at-app/pull/945/commits/1d14cb9c97ec7f35168c46f1cd7e7359075e031c

alflennik (approved):
Thanks for fixing that bug I found. Additionally, I verified the three fixes. The collection form as well as the results page are correctly showing the assertions, the right number of assertions (46), and the correctly formatted text for the command. Looks great!

PR #948, opened 2024-03-06
Use uniquely created ids for `aria-labelledby` referenced in `InstructionsRenderer`

howard-e:
Follow up for #944.

Fixes issue where non-unique ids were being used for `aria-labelledby` with the assertions table in the `InstructionsRenderer` component. This can be tested on the Test Review and Candidate Test Plan Run pages.

Addresses:
* https://github.com/w3c/aria-at-app/pull/944#issuecomment-1981463867
* https://github.com/w3c/aria-at-app/pull/944#issuecomment-1981481738


alflennik (approved):
Looks good thanks for catching this one

Issue #914, opened 2024-02-07
Add support for automated results collection with Firefox and NVDA


jugglinmike:
Currently, ARIA-AT App only supports automated results collection using NVDA with the latest version of Chrome. Extend this support to include using NVDA with the latest version of Firefox.

PR #950, opened 2024-03-07
Enable nvda / firefox browser combination for automated testing

gnarf:
Tested locally seems fine!

### DIFF
--- a/server/services/GithubWorkflowService.js
+++ b/server/services/GithubWorkflowService.js
@@ -21,7 +21,7 @@ const ALGORITHM = 'RS256';
 // > the workflow file is on the default branch.
 //
 // https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch
-const WORKFLOW_FILE_NAME = 'nvda-chrome.yml';
+const WORKFLOW_FILE_NAME = 'nvda-test.yml';
 const WORKFLOW_REPO = 'bocoup/aria-at-gh-actions-helper';
 
 // Generated from the GitHub.com UI
@@ -126,13 +126,15 @@ const createGithubWorkflow = async ({ job, directory, gitSha }) => {
         jsonWebToken,
         GITHUB_APP_INSTALLATION_ID
     );
+    const browser = job.testPlanRun.testPlanReport.browser.name.toLowerCase();
     const inputs = {
         callback_url: `https://${callbackUrlHostname}/api/jobs/${job.id}/result`,
         status_url: `https://${callbackUrlHostname}/api/jobs/${job.id}/update`,
         callback_header: `x-automation-secret:${process.env.AUTOMATION_SCHEDULER_SECRET}`,
         work_dir: `tests/${directory}`,
         test_pattern: '{reference/**,test-*-nvda.*}',
-        aria_at_ref: gitSha
+        aria_at_ref: gitSha,
+        browser
     };
     const axiosConfig = {
         method: 'POST',
--- a/client/utils/automation.js
+++ b/client/utils/automation.js
@@ -2,7 +2,10 @@ export const isSupportedByResponseCollector = ctx => {
     if (!ctx || !ctx.at || !ctx.browser) {
         return false;
     }
-    return ctx.at.name === 'NVDA' && ctx.browser.name === 'Chrome';
+    return (
+        ctx.at.name === 'NVDA' &&
+        (ctx.browser.name === 'Chrome' || ctx.browser.name === 'Firefox')
+    );
 };
 
 export const isBot = user => user?.username?.toLowerCase().slice(-3) === 'bot';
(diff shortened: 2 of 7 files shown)


server/services/GithubWorkflowService.js
```diff
@@ -126,13 +126,15 @@ const createGithubWorkflow = async ({ job, directory, gitSha }) => {
         jsonWebToken,
         GITHUB_APP_INSTALLATION_ID
     );
+    const browser = job.testPlanRun.testPlanReport.browser.name.toLowerCase();
```
howard-e, line 129: There are keys provided for the browsers but it's not currently passed from the database or available in the resolvers.

It's definitely not a problem here but wanted to raise that, in case it may lighten and simplify the load on this service trying to derive a key from `"VoiceOver for macOS".toLowerCase()` for example, in the future. It would also maintain predictable key maps for the browsers across this app and the workflow(s) being triggered.

howard-e (approved):
Looks good! Did a test locally (to ensure the chrome build hasn't changed)

_Tiny nitpick - I haven't reviewed this file closely before but there's a typo on L42: `'Envrionment ...` if you'd also like to include that here_

gnarf:
> Tiny nitpick - I haven't reviewed this file closely before but there's a typo on [L42: `'Envrionment ...`](https://github.com/w3c/aria-at-app/blob/279150841938a10c76bed95199739ab8ff0002a5/server/services/GithubWorkflowService.js#L42) if you'd also like to include that here

#954

server/services/GithubWorkflowService.js
howard-e: Oop! I realize now that I gave a bad and **very confusing** example because I must have viewed the `AT` table at the time I made the comment.

> There are keys provided for the browsers but it's not currently passed from the database or available in the resolvers.

I was wrong on this. Still in my confusion, I was referencing ATs **BUT** those keys also aren't being tracked in the db, instead being derived. This comment is then irrelevant to this PR.

> It seems like browser is just id/name (and the name to lower also seem pretty safe in this case)

It is safe!

But the point I'm making could still stand, in that there should be stored keys somehow. Not something for this PR but I think I'm just now becoming aware of a potential data consistency issue.

In the examples you gave here, when "Edge" gets added, they may want to present it as "Microsoft Edge" but the service (or other sections of the app) could want to resolve `"Microsoft Edge"` to `"edge"` for simplicity. Or if it ever comes to adding `"Safari (for iOS)"` -> `"safari_ios"`, it would mean renaming the current "Safari" browser being tracked to something more appropriate like "Safari (for macOS)" which could have been implicitly being thought of as `"safari_macos"` anyways.

All that to say, I'd feel safer using:
```javascript
job.testPlanRun.testPlanReport.browser.key;
```

VS

```javascript
job.testPlanRun.testPlanReport.browser.name.toLowerCase();
```

OR

```javascript
getBrowserKey(job.testPlanRun.testPlanReport.browser.name.toLowerCase());
```

If you or others agree to those keys, then that's a separate task to maybe make future needs easier.

gnarf: ^ i don't feel entirely qualified to make this patch, the database stuff is still a bit of a mystery to me, but i do think that a "key" that isn't the "display name" here is probably the long-term "right" solution. (Followup issue?) I don't think we need to solve it to land this however.

howard-e: > (Followup issue?) I don't think we need to solve it to land this however.

Agreed! I can make a follow up task to discuss

stalgiag (changes requested):
Thanks for starting this. What is here looks good but there are a few related things that need updating at the same time.

[The code that is responsible for making a new browser ](https://github.com/w3c/aria-at-app/blob/main/server/controllers/AutomationController.js#L254) version if the response collector reports a browser version that doesn't exist already always assumes Chrome.

The [local simulated automation scheduler also assumes Chrome](https://github.com/w3c/aria-at-app/blob/main/server/tests/util/mock-automation-scheduler-server.js#L149) when it selects the relevant test plan report from a test plan version.

server/services/GithubWorkflowService.js
stalgiag: Just adding a +1 to what @howard-e is suggesting!

gnarf:
> Thanks for starting this. What is here looks good but there are a few related things that need updating at the same time.

thanks! glad you had the time to provide this input before it was forgotten! I'll get right on these changes!

server/services/GithubWorkflowService.js
howard-e: #958

alflennik (approved):
I looked over all the code, but mainly I focused on the transaction changes, which look to be really solid enhancements. I think we should merge this PR before https://github.com/w3c/aria-at-app/pull/951 because I do expect some conflicts and I don't want them to block this PR.

howard-e (approved):
Looks good to me! Went back over the newer changes since I've last reviewed and all seems in line to me.

Thanks as well for including the todo on what needs to happen when #958 is resolved.

edit: @gnarf I can't merge since there's a conflict to resolve so please feel free to do so after resolving that

jugglinmike:
This patch implements gh-914.

PR #951, opened 2024-03-08
Adding and Removing AT Versions End-to-End Test

alflennik:
This PR implements the first end-to-end test which uses the full end-to-end testing capabilities, including a complete debugging experience, resetting the database state, testing utilities and documentation.

I chose the smallest feature I could to test, settling on the ability to add, edit and remove AT Versions. It was not ultimately super simple to test - I had to update the atLoader and browserLoader slightly, since the caching we recently introduced made using transactions impossible. This is the source of the changes to context and transactions throughout the server.

If reviewers would be able to verify that debugging works that would be hugely appreciated.

Currently there are conflicts and changes showing from other branches. Once the model-transactions-2 branch is merged I will do a pass over this PR and fix them.

---
  - https://app.asana.com/0/0/1206707065690631


alflennik:
@howard-e and @evmiguel thanks for volunteering to review, this is ready for review now.

server/models/loaders/AtLoader.js
```diff
@@ -1,16 +1,10 @@
 const { getAts } = require('../services/AtService');
```
evmiguel, line 1: Was there a reason for removing the singleton implementation on the `AtLoader` and `BrowserLoader`?

server/services/PopulatedData/populateData.js
```diff
@@ -24,11 +24,8 @@ const getTests = require('../../models/services/TestsService');
  * and no database queries will be run by this function.
  * @returns
  */
-const populateData = async (
-    locationOfData,
-    { transaction, preloaded } = {}
-) => {
-    const context = { transaction };
+const populateData = async (locationOfData, { context, preloaded } = {}) => {
+    const { transaction } = context;
```
evmiguel, line 28: Curious about the reasoning behind changing the API to accept the `context` instead of the transaction?

server/models/services/AtService.js
```diff
@@ -1,13 +1,11 @@
 const ModelService = require('./ModelService');
-
 const {
     AT_ATTRIBUTES,
     AT_VERSION_ATTRIBUTES,
     AT_MODE_ATTRIBUTES,
     BROWSER_ATTRIBUTES
 } = require('./helpers');
 const { Sequelize, At, AtVersion, AtMode } = require('../');
-const { clearCachedAts } = require('../loaders/utils');
```
evmiguel, line 10: Can you remind me of the caching issue and the reasoning behind removing this?

evmiguel (approved):
Great work @alflennik! Code looks great. I have some questions around the intentional behind removing caching and some code decisions made, but overall I can confirm that the e2e test works!

client/tests/util/getPage.js
```diff
@@ -113,9 +125,17 @@ const getPage = async (options, callback) => {
         throw new Error('Please provide a URL, even if it it is simply "/"');
     }
 
-    await page.goto(`http://localhost:${CLIENT_PORT}${url}`);
+    if (isDebugMode) {
+        page.setDefaultTimeout(604800000 /* a week should be enough */);
```
howard-e, line 129: > a week should be enough 

I would hope so! :)

server/graphql-context.js
```diff
@@ -1,16 +1,17 @@
...
-        transaction = false;
-    }
+    // Req will not be defined when queries are made with apolloServer.executeOperation()
+    let transaction = req ? req.transaction : false;
+
+    const atLoader = AtLoader();
+    const browserLoader = BrowserLoader();
 
-    return { user, transaction };
+    return { user, atLoader, browserLoader, transaction };
```
howard-e, line 14: Isn't this going back to how this file used to be? Should it? Didn't this have some implications to automation?

server/models/CollectionJob.js
```diff
@@ -9,7 +9,8 @@ module.exports = function (sequelize, DataTypes) {
             id: {
                 type: DataTypes.STRING,
                 allowNull: false,
-                primaryKey: true
+                primaryKey: true,
+                autoIncrement: true
```
howard-e, line 13: Wouldn't this need a migration?

server/models/index.js
```diff
@@ -17,7 +17,8 @@ sequelize = new Sequelize(
             ssl: true,
             native: true
         },
-        logging: false // console.log // eslint-disable-line
+        // logging: console.log // eslint-disable-line
+        logging: false
```
howard-e, line 21: Guessing you want to revert these 2 lines

howard-e (commented):
This is exciting! Thanks for the many previews you've already given and spent some time messing with it myself. Looks good!

I've left a few inline comments and questions but no major blockers here.

.eslintrc.json
```diff
@@ -26,7 +26,8 @@
         "eol-last": ["error", "always"],
         "no-console": ["error", { "allow": ["warn", "error"] }],
         "no-use-before-define": ["off"],
-        "react/display-name": ["off"]
+        "react/display-name": ["off"],
+        "jest/expect-expect": ["off"]
```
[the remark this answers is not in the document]
alflennik: This is intentional but we can talk about it. The way puppeteer works, it will time out and throw an error when expected UI elements fail to appear. In a sense every line of an e2e test can be considered an assertion. That means explicit assertions would be both redundant ... and unreachable!

It's true that the current e2e tests all have assertions in them, but I can 100% envision valid e2e tests that don't include them, that's why I removed the requirement.

Does that clear things up? Any follow up comments?

client/tests/AtVersions.test.js
```diff
@@ -0,0 +1,48 @@
...
+            await openTrayIfClosed();
+            await page.click('button ::-p-text(Add a New Version)');
+            await page.waitForSelector('.modal-title ::-p-text(Add a New Version for JAWS)');
+            await page.waitForSelector('.modal-body .form-group:nth-child(1) ::-p-text(Version Number)');
+            await page.waitForSelector('.modal-body .form-group:nth-child(2) ::-p-text(Approximate date of availability)');
+            await page.type('.modal-body .form-group:nth-child(1) input', '99.0.1');
+            await page.type('.modal-body .form-group:nth-child(2) input', '01-01-2000');
+            await page.click('.modal-footer button ::-p-text(Add Version)');
+            await page.waitForNetworkIdle();
+            await openTrayIfClosed();
```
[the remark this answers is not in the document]
alflennik: Yes! I think it's caused by the page either refreshing the component or possibly refreshing the entire page. It would be great to follow up and address that.

client/tests/AtVersions.test.js
```diff
@@ -0,0 +1,48 @@
...
+            await page.click('.at-versions-container button ::-p-text(Remove)');
+            await page.waitForSelector('.modal-title ::-p-text(Remove JAWS Version 99.0.99)');
+            await page.click('.modal-footer button ::-p-text(Remove)');
+            await page.waitForNetworkIdle();
+            await openTrayIfClosed();
+            const option = await page.$('.at-versions-container option ::-p-text(99.0.99)');
+            expect(option).toBeNull();
+        });
+    });
+});
```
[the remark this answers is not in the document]
alflennik: You know, what I was imagining, if someone needed to come along later and figure out what's happening, is that they would use breakpoints and actually take a look at the real page shown in the browser for each step of the test. This was a motivation for putting extra care into the debugging support for end-to-end tests. I did experiment with creating variables for some of these selectors, but I found that it didn't create a more readable test and in fact may have made it even more intimidating by adding indirection and making it less of a step-by-step "recipe," in addition to making the test quite a bit longer.

With readability in mind, I did try choose selectors not just to point at the elements in question, but also to tell a bit of a story about what is happening. For example, I had `.modal-title ::-p-text(Add a New Version for JAWS)` which hopefully conveys the idea that we're searching inside a modal's title for the text "Add a New Version for JAWS."

server/graphql-context.js
alflennik: Thanks for asking about this. This is definitely one of the larger changes in the PR. So you're right that the automation PR changed the way the atLoader and browserLoader worked. However that change should be considered more of a refactor or a simplification than something strictly required for automation. When I reviewed the first major automation PR I had a chance to go over that change in detail, including discussing with Stalgia's her motivation for making the change. 

When she introduced the "singleton" architecture which is currently still in use on main, she had not realized that it was possible to dynamically add new AT and browser versions. Therefore, in the first version of the PR, the caching inherent to the singleton approach would result in stale data being loaded for AT and Browser versions, until the server was restarted. To fix this she added several functions for clearing the cache. This was a good solution at the time to quickly unblock the PR, but a I think a couple of developments since then made going back to the old approach preferable again.

Number one, one of the big benefits of using a singleton architecture was the fact that passing context to all the resolvers was no longer needed. This is no longer the case. The context is now a required field, because the transaction is a required field.

Number two, the other big benefit of using a singleton architecture was the fact that the cache could be more aggressive and hit the database less often, almost never. In the previous approach we would have had to query the database once during every request. However this aggressive caching was the real show stopper when I came along to test this feature. The issue is that the caching would not respect the current transaction. And in addition to possibly causing issues in production for code that mutates the AT and browser versions before the transaction is committed, the main problem is that this actually made testing with a long-lived transaction completely impossible.

Hopefully that "little" essay is enough to understand the reasoning for the change. Thanks for your patience as I try to explain that.

server/models/loaders/AtLoader.js
alflennik: Hi Erika! Good question, take a look at a "little essay" I wrote about it here: https://github.com/w3c/aria-at-app/pull/951#discussion_r1529061043

server/services/PopulatedData/populateData.js
alflennik: Now that context is a required field for resolvers, this is needed by populateData so that it can call out to any resolvers. And in fact it does call the `testPlanVersionTestPlanResolver` below.

server/models/services/AtService.js
alflennik: Hopefully my big explanatory comment will clear this up! https://github.com/w3c/aria-at-app/pull/951#discussion_r1529061043

server/models/CollectionJob.js
alflennik: Oh shoot this change snuck in from another PR, I will remove it.

server/models/index.js
alflennik: Good guess, but actually this is intentional! I noticed that it's hard to quickly turn logging on and off. I figured out that it would be faster and easier to have two lines where you can comment one and uncomment the other. So that's why I changed this.

alflennik:
@howard-e one quick note! If you do approve this, would you be able to hold off merging it? I want to get https://github.com/w3c/aria-at-app/pull/950 merged first to avoid a high potential for conflicts getting in the way of that PR.

client/tests/util/getPage.js
alflennik: Me too haha 😂 some bugs seem like a week would be necessary.

server/models/index.js
howard-e: Got it. Well wouldn't it be better to control logging behind a flag then? Seems very easy as something to have future merge conflicts if someone forgets to revert.

Since something like this only has value in a dev environment. I figure doing `yarn dev-debug` could enable that flag vs `yarn dev`.

.eslintrc.json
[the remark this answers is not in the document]
howard-e: Interesting! I'm not blocking on this but here are some follow up thoughts that could be discussed later on:
1. While I'm not the biggest fan of using `warn`. This does seem like a case that _could_ warrant it.
2. You could use jest's [overrides](https://eslint.org/docs/latest/use/configure/rules#using-configuration-files-1) to turn that rule off for an e2e test. But it would mean having to now name e2e test suites with a specific name. eg. `AtVersion.test.js` -> `AtVersion.e2e.test.js`
3. Ignoring each e2e file instead.

This is all opinion though, really I don't expect tests outside of the e2e ones as you've explained, to have missing expects anyways and even then, would be caught by a review most likely. But purely a precaution.

client/tests/AtVersions.test.js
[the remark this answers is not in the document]
howard-e: Certainly something to track!

client/tests/AtVersions.test.js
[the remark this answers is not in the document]
howard-e: Understood and I appreciate the care already taken towards authoring these test suites in that regard. I could also be alone in my current thought on it 😅.

But certainly curious about any others' thoughts in the future in case there is a middle ground here!

server/graphql-context.js
howard-e: Makes sense! Haven't noticed any significant setbacks with this change during my perusal of the app with this PR.

> Hopefully that "little" essay is enough to understand the reasoning for the change. Thanks for your patience as I try to explain that.

It does! Thanks for sharing this "essay" :), and I appreciate the work and discussions you and @stalgiag must have to had to ensure continuous changes to this file are as safe as possible to make.

howard-e (approved):
@alflennik looks good! Thanks for addressing my one point of actionable feedback and also replying to my other queries. I've left inline comments on items that could be follow up discussions but nothing I feel needs to be blocking at the moment.

Happy for yet another stepping stone to better testing environment!

> would you be able to hold off merging it? I want to get https://github.com/w3c/aria-at-app/pull/950 merged first to avoid a high potential for conflicts getting in the way of that PR.

Sure, I'll leave the merging to you until #950 is ready.

server/models/index.js
alflennik: I'll just revert it

.eslintrc.json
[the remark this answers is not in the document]
alflennik: I'll revert the change then.

PR #955, opened 2024-03-11
Add npm to ansible

evmiguel:
This PR addresses [this Asana task](https://app.asana.com/0/1206547249735190/1206719837277162), where the import tests job was failing in staging and production. NPM was not installed on provisioning of the server, which is why the import tests job was failing.

### DIFF
--- a/deploy/roles/nodejs/tasks/main.yml
+++ b/deploy/roles/nodejs/tasks/main.yml
@@ -25,6 +25,7 @@
     name:
       - nodejs
       - yarn
+      - npm
     state: present
   become: yes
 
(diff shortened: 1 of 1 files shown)


howard-e (approved):
Looks good! Confirmed this restores the import's functionality on staging. Thanks for finding this!

jugglinmike (approved):
Looks good to me.

Thanks for including your research on that Asana task, @howard-e. This would have been a much more straightforward diagnosis if the script reported errors. [I've filed an issue to improve error reporting along these lines.](https://github.com/w3c/aria-at-app/issues/957)

PR #956, opened 2024-03-11
Support copying of results' assertions even when there is a changed assertion

howard-e:
Currently, when test results are being copied from one `TestPlanReport` for an older `TestPlanVersion`, to another `TestPlanReport` for a newer `TestPlanVersion`, it uses the same functionality that allows unique copies of TestPlanVersions to be imported into the application. What that means is that all the elements of an individual test is being taken into account.

But it also doesn't give any consideration to the cases where only a single assertion has been updated on a test and so. So if a new TestPlanVersion updates an assertion to fix a typo OR changes an `assertionId`, it invalidates all the other results attached to that test.

This PR updates the app to account for per-assertion level results being copied by using the newly introduced `assertionId` from `assertions.csv`


alflennik (approved):
Looks good, thanks for addressing the feedback I gave offline. I was able to replicate the successful case of upgrading the toggle button test plan. I will say that I could envision having to revisit this code in the future for other update scenarios, but that shouldn't be a blocker getting this fix in. Thanks!

PR #960, opened 2024-03-13
Update Test Plan Report Status dialog completion percentage calculation

howard-e:
Currently being shown on aria-at.w3.org for the Toggle Button V24.03.12 Test Plan Report Status Dialog is that all the current reports in the Test Queue are 100% complete:

Screenshot showing 100% complete status across JAWS, NVDA and VoiceOver reports currently in the Test Queue

But the Test Queue shows that only 2 of 8 tests are complete for each listed tester:

Screenshot showing 2 of 8 tests complete for both testers assigned to JAWS and Chrome

That's because the percentage is being calculated against whether the test result for a test exists (which is created once you navigate to the test within the Test Run page) vs checking if the test has actually been completed by the tester.


alflennik (approved):
Looks great. I was able to verify the issue and the fix using a recent dump of the production database.

PR #961, opened 2024-03-14
Break out individual test commands for github actions

gnarf:
This will allow all of the tests to run even if there are errors in other suites (a lint failure will still let tests run, a failure in client won't stop server from running)

### DIFF
--- a/.github/workflows/runtest.yml
+++ b/.github/workflows/runtest.yml
@@ -41,9 +41,28 @@ jobs:
           yarn workspace server db-import-tests:test -c ${IMPORT_ARIA_AT_TESTS_COMMIT_3}
           yarn workspace server db-import-tests:test -c ${IMPORT_ARIA_AT_TESTS_COMMIT_4}
           yarn workspace server db-populate-sample-data:test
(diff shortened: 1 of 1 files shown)


gnarf:
#92 is an example PR using this yml change that introduces a lint and test error in the "shared" folder but all the other bits of the test still run.

gnarf:
screenshot showing the failure status of the individual jobs on linked pull request 962

alflennik (approved):
This is a great change, and honestly I wish I had thought of it 😄 It resolves the issue where a single extra console log would prevent any of the tests from running. And it also resolves another issue that I noticed recently, that when the smoke test fails, indicating an issue with the server, it would then preclude the server from running its tests, making it very unclear what exactly is wrong with the server.

Thank you for including a screen shot of what the output looks like, I was slightly concerned that the failure would be buried somewhere deep in the logs but it's actually going to be quite easy to track down any failures.

howard-e (approved):
Thanks @gnarf! Should make tracking down GH build issues a bit easier

Issue #935, opened 2024-02-26
Improving the test results copy functionality


howard-e:
Currently, the app supports copying results for reports when updating a `TestPlanVersion`'s phase under the following conditions:
* When the earlier `TestPlanReport`s' data to be used is a part of a `DRAFT` `TestPlanVersion`
  * Note that if a report is marked as final while in `DRAFT`, when the newer report is included and copies over the results, it also copies the marked final status and isn't unmarked in the Test Queue to make it clear that those results have been copied over.
* Assertion text for a command for a test has not changed.

The above creates the following problems when:
* A `TestPlanVersion` is in an advanced phase (`CANDIDATE` or `RECOMMENDED`) and it has results which could be copied over into a new version, to make subsequent test collection and review cycles easier
* When assertion text has changed in the instance of just a typo being fixed and now several tests using that assertion has to be completely copied over again.
* When commands have been added, removed or edited and now several unrelated results for other commands in that test have to be completely copied over again

The suggestions as noted through [recent CG discussions](https://www.w3.org/2024/02/22-aria-at-minutes.html) and functional reviews is that the following needs to happen to support a better results copying process between `TestPlanVersion`s and their `TestPlanReport`s:
- [x] During the tests comparison process between versions, relax constraints to allow situations such as typos to be considered "the same" (#956)
- [x] Need to account for when commands have been added, removed or edited for a test (#967)
- [x] Unmark test plan reports as final once the results have been copied into a new TestPlanVersion's TestPlanReports to make it clearer that the results have been successfully copied. Test administrators will be able to make changes and mark as final again, if appropriate (#974)
- [x] Functionality to support copying results when a TestPlanVersion is in an advanced phase (#976)
- [x] Need to be able to pull older results when also adding an individual Test Plan Report to the Test Queue instead instead of just providing that functionality during the advance process from the Data Management page (#978)

Note
* #877 describes additional detailed requirements of this process that are currently in place, and should be maintained.

howard-e:
From CG discussion on Mar 13:
* Need to account for when commands have been added or removed from a test
* Need to be able to pull older results when also adding an individual Test Plan Report to the Test Queue instead instead of just providing that functionality during the advance process from the Data Management page

PR #967, opened 2024-03-18
Support results being copied even when commands have been added or removed between versions

howard-e:
This PR updates `processCopiedReports.js` to account for when commands have been added/removed between versions.

Previously, if a test had a command be included or removed in a subsequent update, all assertion results, unexpected behaviors and output would be discarded during the copy process.

Addresses the following from #935:

> 1. Need to account for when commands have been added, removed or edited for a test
> 2. Unmark test plan reports as final once the results have been copied into a new TestPlanVersion's TestPlanReports to make it clearer that the results have been successfully copied. Test administrators will be able to make changes and mark as final again, if appropriate.
> 3. Functionality to support copying results when a TestPlanVersion is in an advanced phase.
> 4. Need to be able to pull older results when also adding an individual Test Plan Report to the Test Queue instead instead of just providing that functionality during the advance process from the Data Management page

2, 3 and 4 have been individually reviewed in separate PRs through:
- #974 
- #976 
- #978


server/resolvers/helpers/processCopiedReports.js
```diff
@@ -6,22 +6,250 @@ const {
...
+        // settings may not exist if v1 test format
+        if (settings) scenarioId = `${scenarioId}_${settings}`;
+        knownScenarioIdsForTest.push(scenarioId);
+    }
+
+    return [knownScenarioIdsForTest, knownAssertionIdsForTest];
+};
+
+/**
+ * Determine which tests between an 'old' and 'new' TestPlanVersion is the same between both
```
Paul-Clue, line 50:
```suggestion
 * Determine which tests between an 'old' and 'new' TestPlanVersion are the same between both
```

Paul-Clue (approved):
I did highlight one change, but it was something insignificant in a comment, so I removed  the suggestion. The changes here are working fine, and the code seems good to me. Good stuff, as always.

PR #968, opened 2024-03-18
Fix persistence of Aria live region

Paul-Clue:
see issue #797 

This pull request fixes the issues of the Arial live region persisting forever.


alflennik (approved):
I tested this in NVDA and VoiceOver, including triggering a large number of alerts to make sure the timeouts wouldn't cause the output to fail to read or get cut off. But this clean and simple fix worked perfectly. If I did the task it would definitely come out a lot more convoluted. Nice work!

PR #971, opened 2024-03-19
Remove AT and browser alert banner

Paul-Clue:
see issue #797 

This pull request removes the green alert banner from the top of the test run page after saving data and reloading the page.

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -76,6 +76,22 @@ const TestRun = () => {
             variables: { testPlanRunId, testPlanReportId }
         }
     );
+    /* HOMEWORK
+        *try changing or turning off the fetchPolicy so that loading won't be false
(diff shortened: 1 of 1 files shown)


client/components/TestRun/index.jsx
```diff
@@ -1390,6 +1409,7 @@ const TestRun = () => {
                         )}
                         patternName={testPlanVersion.title}
                         testerName={tester.username}
+                        // section:
```
alflennik, line 1412: Would you be able to remove this comment and any other unrelated comments? 🙂

Paul-Clue: Ok I just removed them

alflennik (approved):
Kudos on the efficiency of the fix, I verified that it works both when you click "next test" and also when you navigate with the test navigator.

Issue #946, opened 2024-03-05
Update BrowserVersion table to include releasedAt column


ccanash:
We will need to start tracking the release date for browser versions for the benefits of automation and trend reporting.

This task is to add the release data to the BrowserVersion table, there is no need to populate or make use of the column at this time.

PR #972, opened 2024-03-19
Add releasedAt to browserVersion

alflennik:
See https://github.com/w3c/aria-at-app/issues/946. 

Note that the column is not expected to be populated at this time.


howard-e (approved):
LGTM!

PR #974, opened 2024-03-20
Unmarks newly copied test plan reports as final during results copy process

howard-e:
During the current results copy process, the `markedFinalAt` status is being copied from the older reports. It wouldn't allow the test admin(s) to review the newly copied reports to ensure everything as expected before being displayed on the generated reports again.

This PR addresses the following from #935:

> Unmark test plan reports as final once the results have been copied into a new TestPlanVersion's TestPlanReports to make it clearer that the results have been successfully copied. Test administrators will be able to make changes and mark as final again, if appropriate.

Dependent on #967


alflennik (approved):
I like this! This seems like a positive usability enhancement. Code looks good and I was able to successfully copy data from an old version of Command Button to a new version. 99% of the data was copied which is great.

PR #976, opened 2024-03-20
Copy results from advanced phase

howard-e:
**Problem**

During the current results copy process, only the results from the 'next' available older TestPlanVersion is considered when preserving results during an update on the DataManagement page.

That is, if there is an older `TestPlanVersion` in `CANDIDATE` and a newly imported `TestPlanVersion` in `RD` is being promoted to `DRAFT`, it will not attempt to copy the results from the `CANDIDATE` `TestPlanVersion`. It WOULD copy the results if the older `TestPlanVersion` was in `DRAFT` however.

This creates an issue for ease of testing and preservation of future results capture. It would force testers to have to re-run entire suites and compare against past results manually.

**Solution**

This PR now allows for results to be copied from advanced phases which are 1+ phases away from the current. ie. when promoting to `DRAFT` from `RD`, reports could be copied from the latest imported `CANDIDATE` or `RECOMMENDED` TestPlanVersions if applicable.

This addresses the following from #935:

> Functionality to support copying results when a TestPlanVersion is in an advanced phase.

Dependent on #974


Paul-Clue (approved):
It's working fine in my local setup, and the code looks good to me!

PR #978, opened 2024-03-21
Copy results when adding to Test Queue

howard-e:
**Problem**

Currently, the only way reports are copied from an older test plan version's report(s) into a newer test plan version's report(s) is by the implicit action of just using the `Advance to <Phase>` button on the DataManagement page. In the instance where an update has been missed (because of the current state of the data) or if a copied report was removed or otherwise, it would mean that the ability to now copy those older results which matched for the same Test Plan + AT + Browser is functionally lost.

**Solution**

This PR allows a test admin to choose whether to copy those results or create a new report when adding to the Test Queue if older results are found for the valid combination, using a dialog.

This addresses the following from #935:

>  Need to be able to pull older results when also adding an individual Test Plan Report to the Test Queue instead instead of just providing that functionality during the advance process from the Data Management page

Dependent on #976


server/graphql-schema.js
```diff
@@ -1016,6 +1016,7 @@ const graphqlSchema = gql`
         testPlanVersionId: ID!
         atId: ID!
         browserId: ID!
+        testPlanVersionDataToIncludeId: ID
```
alflennik, line 1019: The part that confuses me is that it reads like we're copying data from a TestPlanVersion to a TestPlanReport, which superficially doesn't make sense, even if there's an algorithm happening under the hood. I find myself wondering if it would be possible to call this something like `copyResultsFromTestPlanReportId`. That seems more direct and intuitive to me. Do you think that would be workable?

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -142,31 +183,72 @@ function AddTestToQueueWithConfirmation({
     };
 
     const renderErrorDialog = () => {
```
alflennik, line 185: Seems like we need to rename the dialog now since the new message doesn't seem like an error anymore!

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -142,31 +183,72 @@ function AddTestToQueueWithConfirmation({
...
+        if (oldReportToCopyResultsFrom) {
+            title = 'Older Results Data Found';
+            content =
+                'Older results with the same AT, browser and test plan ' +
+                'were found for the report being created. Would you like ' +
+                'to copy the older results into the report or create a ' +
+                'completely new report?';
+            actions = [
+                {
+                    label: 'Create new report',
```
alflennik, line 215: How about "Create empty report"?

server/graphql-schema.js
howard-e: Fine with me! Done in https://github.com/w3c/aria-at-app/pull/978/commits/bffd2247824ed3d639b6576f8aad11085d539030.

client/components/AddTestToQueueWithConfirmation/index.jsx
howard-e: Sounds good, done in https://github.com/w3c/aria-at-app/pull/978/commits/bffd2247824ed3d639b6576f8aad11085d539030

client/components/AddTestToQueueWithConfirmation/index.jsx
howard-e: Good point, done in https://github.com/w3c/aria-at-app/pull/978/commits/bffd2247824ed3d639b6576f8aad11085d539030. Let me know if you'd like the workshop the name further.

alflennik (approved):
I'm a big fan of this change, it makes copying results way easier and in my testing it behaved flawlessly. I appreciated that we were able to reuse so much existing functionality. I was able to successfully copy results from a previous final report in a different phase as well as copy results from reports still in the test queue, which is awesome. The system behaves identically to now when adding reports where no previous data exists. Great work!

Issue #958, opened 2024-03-12
Include key columns for `At` and `Browser` tables


howard-e:
In reviewing https://github.com/w3c/aria-at-app/pull/950#discussion_r1520092670, found that with both the `At` and `Browser` tables are missing key values for their respective tables' rows. This may make deriving keys for upcoming Ats and Browsers an increasingly tedious process, as how it is currently being done may require several points of updates for each row to the tables.

For the ATs, keys are provided in [w3c/aria-at/tests/support.json](https://github.com/w3c/aria-at/blob/master/tests/support.json) but they aren't reflected in the database or any easy to use reference point. Instead [a function returns either the lowercase version of the At's name or in a special case, converts it to the same key provided by the source support.json](https://github.com/w3c/aria-at-app/blob/5e2334c6c440c6c44116c0a5f9daf130abb6ef49/server/util/aria.js#L4:L8):

```javascript
const evaluateAtNameKey = atName => {
    // Could probably add back support for AT keys from the database level
    if (atName.toLowerCase().includes('voiceover')) return 'voiceover_macos';
    else return atName.toLowerCase();
};
```

In this particular case, it would become problematic when attempting to also derive a key for `VoiceOver on iOS` if that were ever added and may require other locations to make similar updates as well.

The above is the same for Browser, which is that the lower case version of the currently tracked browsers are being used as the "keys".

To ensure better identification and predictability of keys for the `At` and `Browser` rows as the app grows and other services are connected to it, this proposal is to include something somewhat small but may prevent future data consistency issues.

PR #979, opened 2024-03-21
Add key column to At and Browser tables

Paul-Clue:
see issue #958 

This pull request adds a `key` column to the `At` and `Browser` tables in the database.

---
  - https://app.asana.com/0/0/1206834367657798

### DIFF
--- a/server/seeders/20240321184138-populateKeyForAtAndBrowser.js
+++ b/server/seeders/20240321184138-populateKeyForAtAndBrowser.js
@@ -1,0 +1,34 @@
+'use strict';
+
+module.exports = {
+    async up(queryInterface) {
+        await queryInterface.sequelize.query(`
+          UPDATE
+            "At"
+          SET
+            "key" = CASE
+              WHEN id = 1 THEN 'jaws'
+              WHEN id = 2 THEN 'nvda'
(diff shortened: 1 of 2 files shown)


server/seeders/20240321184138-populateKeyForAtAndBrowser.js
```diff
@@ -0,0 +1,34 @@
...
+              WHEN id = 2 THEN 'chrome'
+              WHEN id = 3 THEN 'safari_macos'
+              ELSE null
+            END;
+        `);
+    },
+
+    async down(queryInterface) {
+        await queryInterface.removeColumn('At', 'key');
+        await queryInterface.removeColumn('Browser', 'key');
```
howard-e, line 32: The reverse here is probably to just set `At.key` and `Browser.key` to null for the rows. You could do that with sequelize's `bulkUpdate`

howard-e (commented):
Looks good! Just left a minor inline comment.

server/seeders/20240321184138-populateKeyForAtAndBrowser.js
Paul-Clue: Ok I just used the `bulkUpdate` and it worked great!

Issue #973, opened 2024-03-20
Candidate review start date is errantly reset when a test plan advances to candidate 


ccanash:
The last deployment of the app broke an important working mode requirement.

When a test plan advances to candidate, if there is an earlier version of that plan in candidate review, the candidate review start and target completion dates should not be changed. Now, the candidate review start dates are incorrect for all test plans in candidate review. In addition, target dates were changed for all test plans advanced after the latest release of the app, so they are also incorrect. We can no longer see how overdue those plans are.

mcking65:
This is partially fixed, but the fix introduces another problem that creates equally severe data integrity issues.

Now, when a test plan version is advanced to candidate, it creates two other data problems:
1. The plan that is in candidate review has the wrong date for end of draft review.
2. Version history shows the incorrect plan version number for the start of candidate phase.

These problems might be two different sides of the same coin.

It seems there is some fundamental problem with the logic for the requirements related to events in the test plan life cycle and how those events define the start and end of phases in the working mode.

1. Draft review phase starts for a given test plan when a version of that plan enters draft review and there are no earlier versions of that plan in draft review at that time.
2. If a test plan version advances to draft, and there is already a version of that plan in draft review, the earlier version of that plan is deprecated.
3. Draft review ends when a test plan version advances from draft to candidate review.

The logic is the same for candidate review.

Recommended phase is only different in that once a test plan has a recommended version, there is no end to the phase because it is the last phase.

For problem 1, look at the Alert example row on the data management page in staging. In the draft review column it says:

>V23.12.06
>Review Completed Dec 9, 2022 

![image](https://github.com/w3c/aria-at-app/assets/7241781/1aebcce2-38d4-409e-8d3d-dca813d73afe)

Obviously, review of a test plan version that was developed in 2023 could not have had draft review complete in 2022. That should say:

>V23.12.06
>Review Completed Feb 26, 2024

I figured out the Feb 26 date by looking at the version history page and found a Feb 26, 2024 deprecation date for V22.12.08 in candidate review.

There are several other places where the incorrect Dec 9, 2022 date shows up instead of the Feb 26, 2024 date. In the version summary on the [version history page](https://aria-at-staging.w3.org/data-management/alert), the first row says:

>V23.12.06 started Candidate on Dec 9, 2022

Similarly, looking at the [test plan review page for V23.12.06 of alert](https://aria-at-staging.w3.org/test-review/56291), it says:

[...]

PR #986, opened 2024-03-26
Copy `recommendedPhaseTargetDate` while in same candidate review window for a test plan version

howard-e:
Address latest comments in https://github.com/w3c/aria-at-app/issues/973#issuecomment-2016916399.

---
  - https://app.asana.com/0/0/1206922478712970


gnarf (commented):
I'm not familiar enough with the workflow to provide a good local test that it is now working correctly, but the code seems correct and the additional comments about the window are useful.

gnarf (approved):

alflennik (approved):
This was a doozy to test! With some tweaking of my system clock I was able to verify that the target dates are preserved when deprecating a candidate test plan, and that the target dates are NOT preserved when there is no candidate test plan.

PR #987, opened 2024-03-26
Clean up incorrect database values in production

howard-e:
Found 4 issues of varying importance while making changes to support #986:

1. There were unexpected versions included for all 35 test plans which seemed to have been created during the transition to the v2 test format and the `hashTests` function wasn't properly implemented at that time and the imports were missed. These instances are all still in RD, and so can be removed without issue.
2.  As a result of no.1, there were 8 valid instances of test plan versions which were unnecessarily deprecated so we can revert that.
3. The current [aria-at.w3.org' toggle-button page](https://aria-at.w3.org/data-management/toggle-button) has an additional version in DRAFT phase which shouldn't be possible. That is V23.12.13 should be shown as being deprecated by V23.12.14. I can only guess as to why this is but I assume this also stemmed from unexpected handling of data during the v2 test format update changes.
4. [Radio Group Example Using aria-activedescendant Test Plan V24.03.13](https://aria-at.w3.org/test-review/75322) currently shows assertion phrases which should be using replaced by intended token strings as the generic string. ie `Test 1 > JAWS > Tab (virtual cursor)`'s last assertion phrase is saying _"switch from reading mode to interaction mode"_ instead of _"switch from virtual cursor active to PC cursor active"_ and the same is true for additional JAWS and NVDA assertions where this tokenization should have happened. That's because [the PR](https://github.com/w3c/aria-at-app/pull/942) which would support this happening during the import (and only during the import) has not yet been included in production and so it warrants an update happening in place.

This PR addresses the above.

### DIFF
--- a/server/migrations/20240325205951-cleanupDuplicatesAndRegenerateHashes.js
+++ b/server/migrations/20240325205951-cleanupDuplicatesAndRegenerateHashes.js
@@ -1,0 +1,124 @@
+'use strict';
+
+const { regenerateResultsAndRecalculateHashes } = require('./utils');
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+    async up(queryInterface, Sequelize) {
+        return queryInterface.sequelize.transaction(async transaction => {
+            // Remove unnecessarily created test plan versions for ALL 35 test
+            // plan versions for commit,
+            // https://github.com/w3c/aria-at/commit/836fb2a997f5b2844035b8c934f8fda9833cd5b2
+            // None of the following ever left R&D
+            await queryInterface.sequelize.query(
+                `delete
+                 from "TestPlanVersion"
+                 where id in
+                       (55591, 55592, 55593, 55594, 55595, 55596, 55597, 55598, 55599, 55600, 55601, 55602, 55603, 55604, 55605, 55606,
+                        55607, 55608, 55609, 55610, 55611, 55612, 55613, 55614, 55615, 55616, 55617, 55618, 55619, 55620, 55621, 55622,
+                        55623, 55624, 55625)`,
+                {
+                    transaction
+                }
+            );
+
+            // Reset the phases and remove deprecatedAt dates for versions affected by above wrongly created versions:
+            // checkbox-tri-state
+            // link-css
+            // link-img-alt
+            // main
+            // meter
+            // radiogroup-roving-tabindex
+            // slider-multithumb
+            // switch
+            await queryInterface.sequelize.query(
+                `update "TestPlanVersion"
+                 set phase          = 'RD',
+                     "deprecatedAt" = null
+                 where id in (1566, 1591, 1618, 1798, 2129, 2162, 2245, 2347)`,
+                {
+                    transaction
+                }
+            );
+
+            // Deprecate older uncaught instance of a 2nd DRAFT version
+            // Toggle Button V23.12.13 and set deprecatedAt to date of when
+            // newer Toggle Button V23.12.14 would have overwritten it,
+            // minus 2 seconds to simulate what would happen in
+            // updatePhaseResolver
+            await queryInterface.sequelize.query(
+                `update "TestPlanVersion"
+                 set phase          = 'DEPRECATED',
+                     "deprecatedAt" = '2023-12-14 21:51:35.527000 +00:00'
+                 where id = 62309`,
+                {
+                    transaction
+                }
+            );
+
+            // Update Radio Group Example Using aria-activedescendant V24.03.13 to include the tokenized assertion phrase for JAWS and NVDA
+            // because it wasn't properly handled by the import script because #942 was not yet in production when version was imported
+            const testPlanVersions = await queryInterface.sequelize.query(
+                `SELECT id, tests FROM "TestPlanVersion" WHERE directory = 'radiogroup-aria-activedescendant' AND "versionString" = 'V24.03.13'`,
+                {
+                    type: Sequelize.QueryTypes.SELECT,
+                    transaction
+                }
+            );
+            await Promise.all(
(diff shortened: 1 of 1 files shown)


server/migrations/20240325205951-cleanupDuplicatesAndRegenerateHashes.js
```diff
@@ -0,0 +1,124 @@
...
+            // plan versions for commit,
+            // https://github.com/w3c/aria-at/commit/836fb2a997f5b2844035b8c934f8fda9833cd5b2
+            // None of the following ever left R&D
+            await queryInterface.sequelize.query(
+                `delete
+                 from "TestPlanVersion"
+                 where id in
+                       (55591, 55592, 55593, 55594, 55595, 55596, 55597, 55598, 55599, 55600, 55601, 55602, 55603, 55604, 55605, 55606,
+                        55607, 55608, 55609, 55610, 55611, 55612, 55613, 55614, 55615, 55616, 55617, 55618, 55619, 55620, 55621, 55622,
+                        55623, 55624, 55625)`,
```
gnarf, line 18: Since this is just a raw DELETE, and you have no guarantee this migration is running on a version of the production database, I feel like a backup clause like:

```sql
and "gitSha" = '836fb2a997f5b2844035b8c934f8fda9833cd5b2'
```

should probably be added to this one

server/migrations/20240325205951-cleanupDuplicatesAndRegenerateHashes.js
```diff
@@ -0,0 +1,124 @@
...
+            // main
+            // meter
+            // radiogroup-roving-tabindex
+            // slider-multithumb
+            // switch
+            await queryInterface.sequelize.query(
+                `update "TestPlanVersion"
+                 set phase          = 'RD',
+                     "deprecatedAt" = null
+                 where id in (1566, 1591, 1618, 1798, 2129, 2162, 2245, 2347)`,
```
gnarf, line 37: Again here, this feels like it needs some secondary confirmation that this is the correct update to apply on these rows

server/migrations/20240325205951-cleanupDuplicatesAndRegenerateHashes.js
```diff
@@ -0,0 +1,124 @@
...
+            // Deprecate older uncaught instance of a 2nd DRAFT version
+            // Toggle Button V23.12.13 and set deprecatedAt to date of when
+            // newer Toggle Button V23.12.14 would have overwritten it,
+            // minus 2 seconds to simulate what would happen in
+            // updatePhaseResolver
+            await queryInterface.sequelize.query(
+                `update "TestPlanVersion"
+                 set phase          = 'DEPRECATED',
+                     "deprecatedAt" = '2023-12-14 21:51:35.527000 +00:00'
+                 where id = 62309`,
```
gnarf, line 52: same here

gnarf (changes requested):
Having a migration that has no down step feels.... wrong.... Doing this in a migration feels... wrong....

I understand that a migration for this task might currently be the best option to patch/finesse the data on production, but we may want to consider finding a better one.
I am mostly worried about the inline comments and the possibility of deleting/updating random rows by id only out of an unknown database state.

howard-e:
@gnarf pushed an update in https://github.com/w3c/aria-at-app/pull/987/commits/e154971c1c1e46b1742db3bc0ce710556fca69d2

gnarf (approved):
If we are going to move ahead with the migration approach, the extra specificity fixes my main concerns - I assume the resulting information still checks out from a local test on your side (I'm not 100% sure where to verify it in the frontend on my side)

gnarf:
> But outside of a migration, it could be resolved to a one time patch done directly on the server, and noted as an incidence report since there isn't really a "down" to return here unless we're preserving all the current data for instances like this.

I feel like unless it was seeded data, manual SQL updates could be peer-reviewed and executed on the live database (with appropriate backups) in a future where we try to ensure an up/down path for migrations is viable.

The main complaint I have seeing migrations like this comes from testing pull requests locally, if you can't safely up/down the migrations you can't safely switch between branches with any migration diffs. You are also a lot more restricted in ability to restore previous states to aid with bisecting a bug in the history. It could even be an issue on production.  If there is some extreme issue with the next deployment for instance, we will be unable to revert the deploy to the previous version because of this migration alone (though I think an empty `down()` would maybe solve that for this one, now that it could be safely `up()`ed again without worry)

howard-e:
> > But outside of a migration, it could be resolved to a one time patch done directly on the server, and noted as an incidence report since there isn't really a "down" to return here unless we're preserving all the current data for instances like this.
> 
> I feel like unless it was seeded data, manual SQL updates could be peer-reviewed and executed on the live database (with appropriate backups) in a future where we try to ensure an up/down path for migrations is viable.

I'm all for that approach for the future. So PRs like this could be drafted or issues created to discuss the approach. With the result of the approval being a direct execution on the server and writing it up as a [new incident report](https://github.com/w3c/aria-at-app/blob/main/CONTRIBUTING.md#technical-incident-reporting) perhaps.

> The main complaint I have seeing migrations like this comes from testing pull requests locally, if you can't safely up/down the migrations you can't safely switch between branches with any migration diffs. You are also a lot more restricted in ability to restore previous states to aid with bisecting a bug in the history. It could even be an issue on production. If there is some extreme issue with the next deployment for instance, we will be unable to revert the deploy to the previous version because of this migration alone (though I think an empty `down()` would maybe solve that for this one, now that it could be safely `up()`ed again without worry)

Makes sense. This just reminded me that sequelize explicitly requires the `down()` to support their cli command. At a minimum, I agree that an empty down should at least be included (which I've just added in https://github.com/w3c/aria-at-app/pull/987/commits/ebc4825cb2b16ae07a56eff6db8fb54aefed9e89)

Issue #984, opened 2024-03-25
Timeout error in Graphql


ccanash:
When accessing https://aria-at.w3.org/api/graphql a timeout error message appears

gnarf:
pretty sure this is related to #984 and has been fixed

Issue #983, opened 2024-03-25
Intermittent automation API Error with transactions


gnarf:
While @jugglinmike was testing this weekends issues with the github automation workflow, the job status URL threw an error related to transactions.

https://github.com/bocoup/aria-at-gh-actions-helper/actions/runs/8426667180/job/23075465091#step:3:15

```
SequelizeConnectionAcquireTimeoutError: Operation timeout
    at ConnectionManager.getConnection (/home/aria-bot/aria-at-report/node_modules/sequelize/lib/dialects/abstract/connection-manager.js:204:15)
    at async Transaction.prepareEnvironment (/home/aria-bot/aria-at-report/node_modules/sequelize/lib/transaction.js:85:24)
    at async Sequelize.transaction (/home/aria-bot/aria-at-report/node_modules/sequelize/lib/sequelize.js:499:7)
    at async middleware (/home/aria-bot/aria-at-report/server/middleware/transactionMiddleware.js:30:23)
```

This error happened on the public server for aria-at-app, and the last deployment was on March 11th.

gnarf:
Additional note: We have observed some other intermittent failure states on the app in the last hour, this might of be symptom a deeper server issue

gnarf:
We actually ran into this error twice - https://github.com/bocoup/aria-at-gh-actions-helper/actions/runs/8426935982/job/23076374771#step:3:15 is the second run that hit this error

alflennik:
Thanks for raising this to my attention. Since we added transactions recently I've been paying close attention to any issues and so I wanted to investigate this one quickly.

The code called out in the stack trace looks innocuous enough. It just starts a transaction. It was not initially obvious why a call like that would time out. 

But thinking about it, that line of code is special in that it is the first call to the database. So if the database became unresponsive this would be the first call to fail.

Therefore my current feeling is that this error might indicate some issue with our database, which may or may not be related to transactions at all. That said, transactions are certainly still a candidate for causing havoc until proven otherwise. I think a good next step might be to check the performance strain of running the database - my thinking is along of lines of, is the database sporadically crashing and restarting? Is it possible that a large query is slowing it down so much that it grinds to a halt? This kind of thing.

PR #990, opened 2024-03-27
Fix intermittent connection timeout

alflennik:
See https://github.com/w3c/aria-at-app/issues/984 and https://github.com/w3c/aria-at-app/issues/983.

Here are steps to replicate the error:

1. Check out the main branch.

2. In the `server/app.js` file, find the following code:

```js
apolloServer.start().then(() => {
    apolloServer.applyMiddleware({ app });
});
```

And right after it, paste in the following endpoint:

```js
app.get('/connectiontest', async (req, res) => {
    console.log('connectiontest');

    const transactions = [];
    for (let i = 0; i < 4; i += 1) {
        transactions.push(await sequelize.transaction());
    }
    console.log('transactions created');

    for (const transaction of transactions) {
        await transaction.commit();
    }
    console.log('transactions committed');

    res.status(200).send();
});
```

Also import the sequelize variable which will be undefined.

3. Trigger the endpoint by navigating to `http://localhost:3000/api/connectiontest`. It should load a white page successfully and the terminal should read "transactions created" and "transactions committed".

4. Edit the code `i < 4` to `i < 5`

5. Trigger the endpoint again and verify that the request will time out with the below error after 60 seconds, before which the log will only show "connectiontest".

```
SequelizeConnectionAcquireTimeoutError: Operation timeout
    at ConnectionManager.getConnection (/Users/Shared/Projects/aria-at-app/node_modules/sequelize/lib/dialects/abstract/connection-manager.js:204:15)
    at async Transaction.prepareEnvironment (/Users/Shared/Projects/aria-at-app/node_modules/sequelize/lib/transaction.js:85:24)
    at async Sequelize.transaction (/Users/Shared/Projects/aria-at-app/node_modules/sequelize/lib/sequelize.js:499:7)
    at async /Users/Shared/Projects/aria-at-app/server/app.js:36:27
```

This error occurs when five transactions are opened concurrently. This is because the default sequelize maximum connection count is 5.

After some investigation I found the default postgres connection count is 100, and the fix was to update the sequelize maximum connection count to that number.

6. Stash the changes, check out this branch, the stash apply the changes. Try the endpoint again. It should work without any timeout. Try increasing the `i < 5` to `i < 90` I was able to get it to work with up to 93 simultaneous transactions. 

7. Try increasing `i < 90` to `i < 101`. There should be a more more actionable error message, and it should show immediately instead of slowly timing out. If we see this error in the future it will mean there is probably a leak in our transaction management which is failing to close connections.

The new error:

```
SequelizeConnectionError: remaining connection slots are reserved for non-replication superuser connections
    at Client._connectionCallback (/Users/Shared/Projects/aria-at-app/node_modules/sequelize/lib/dialects/postgres/connection-manager.js:143:24)
    at Client._handleErrorWhileConnecting (/Users/Shared/Projects/aria-at-app/node_modules/pg/lib/client.js:305:19)
    at Client._handleErrorMessage (/Users/Shared/Projects/aria-at-app/node_modules/pg/lib/client.js:325:19)
    at Connection.emit (node:events:513:28)
    at /Users/Shared/Projects/aria-at-app/node_modules/pg/lib/connection.js:114:12
    at Parser.parse (/Users/Shared/Projects/aria-at-app/node_modules/pg-protocol/dist/parser.js:40:17)
    at Socket.<anonymous> (/Users/Shared/Projects/aria-at-app/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
```

I did extensive benchmark testing while trying to replicate the error and I found that in the real world getting 5 transactions to be open at the same time was extremely difficult even under extreme circumstances. It did happen, only once, after dozens and dozens of tests. Assuming there isn't a leak in the future, I'm quite confident that hitting 100 is functionally impossible.

### DIFF
--- a/server/models/index.js
+++ b/server/models/index.js
@@ -17,6 +17,9 @@ sequelize = new Sequelize(
             ssl: true,
             native: true
         },
+        pool: {
+            max: 100
+        },
         logging: false // console.log // eslint-disable-line
     }
 );


howard-e (commented):
Neat! Thanks so much the detailed instructions on how to reproduce locally. I was able to simulate what we experienced and was also able to confirm this fixed it.

> It did happen, only once, after dozens and dozens of tests. Assuming there isn't a leak in the future, I'm quite confident that hitting 100 is functionally impossible.

That does give me some reason to _want to_ investigate further but it isn't of immediate concern as yet. I wasn't able to hit it myself (and I TRIED, ha). So I can also feel some confidence in that statement but it doesn't fully account for the real world case. I think we should definitely stay aware of this and with some targeted benchmarking, determine if we could also benefit from adjusting [pool.idle, pool.acquire or pool.evict](https://sequelize.org/api/v6/class/src/sequelize.js~sequelize#instance-constructor-constructor) should we notice anything like this again.

PR #994, opened 2024-04-01
Address date inconsistencies for candidate phase TestPlanVersions

howard-e:
Addressing latest comments of https://github.com/w3c/aria-at-app/issues/973#issuecomment-2028893100:

**Data Management Overall Status Column date being incorrect**

The Data Management page never kept track of the deprecated versions, so the earliest version date could not have been presented if the version it belonged to was now deprecated. This PR now includes deprecated versions in that date calculation.

**Incorrect phase change date recorded on Command Button versions page for V23.12.13**

This issue isn't related to this PR because it is also [on production](https://aria-at.w3.org/data-management/command-button). I figure something odd happened around this time and this is a one-off situation. There was also an oddity observed for [Toggle Button](https://aria-at.w3.org/data-management/toggle-button) on the same date, with there being a 2nd DRAFT version (but this will be resolved once #987 is deployed to production).

I think fixing this could be a simple SQL update. Using V22.05.04 deprecation date + 1 second, as the guide for the "correct" phase change date as it was the version deprecated in CANDIDATE by V23.12.13, the statement should be:

```sql
update "TestPlanVersion"
set "candidatePhaseReachedAt" = '2024-01-24 02:06:55.460000 +00:00'
where id = 62353
  and "gitSha" = '565a87b4111acebdb883d187b581e82c42a73844'
  and directory = 'command-button';
```

Reviewers, would you rather there be a migration for this change or it be performed on the deployed environment and noted? I have tested this locally on a copy of the production data and found no issues.

**Candidate review page**

This PR makes use of option 2

> Change the column title and leave the data as is.

And updates the `Candidate Phase Start Date` title to `Plan Last Updated`

### DIFF
--- a/client/components/CandidateReview/TestPlans/index.jsx
+++ b/client/components/CandidateReview/TestPlans/index.jsx
@@ -438,9 +438,7 @@ const TestPlans = ({ testPlanVersions }) => {
                         <thead>
                             <tr>
                                 <th>Candidate Test Plans</th>
-                                <CenteredTh>
-                                    Candidate Phase Start Date
-                                </CenteredTh>
+                                <CenteredTh>Plan Last Updated</CenteredTh>
                                 <CenteredTh>Target Completion Date</CenteredTh>
                                 <CenteredTh>Review Status</CenteredTh>
                                 <CenteredTh>Results Summary</CenteredTh>
--- a/client/components/CandidateReview/TestPlans/index.jsx
+++ b/client/components/CandidateReview/TestPlans/index.jsx
@@ -438,7 +438,7 @@ const TestPlans = ({ testPlanVersions }) => {
                         <thead>
                             <tr>
                                 <th>Candidate Test Plans</th>
-                                <CenteredTh>Plan Last Updated</CenteredTh>
+                                <CenteredTh>Last Updated</CenteredTh>
                                 <CenteredTh>Target Completion Date</CenteredTh>
                                 <CenteredTh>Review Status</CenteredTh>
                                 <CenteredTh>Results Summary</CenteredTh>
--- a/client/components/DataManagement/DataManagementRow/index.jsx
+++ b/client/components/DataManagement/DataManagementRow/index.jsx
@@ -432,7 +432,7 @@ const DataManagementRow = ({
          * @param {"RECOMMENDED"|"CANDIDATE"|"DRAFT"} phase
          * @returns {TestPlanVersion[]}
          */
-        const deriveVersionsByDates = phase => {
+        const getVersionsThatReachedPhase = phase => {
             switch (phase) {
                 case 'RECOMMENDED':
                     return testPlanVersions.filter(
@@ -462,13 +462,15 @@ const DataManagementRow = ({
             }
         };
 
-        const derivedRecommendedVersions = deriveVersionsByDates('RECOMMENDED');
-        const derivedCandidateVersions = deriveVersionsByDates('CANDIDATE');
-        const derivedDraftVersions = deriveVersionsByDates('DRAFT');
+        const versionsThatReachedRecommended =
+            getVersionsThatReachedPhase('RECOMMENDED');
+        const versionsThatReachedCandidate =
+            getVersionsThatReachedPhase('CANDIDATE');
+        const versionsThatReachedDraft = getVersionsThatReachedPhase('DRAFT');
 
-        if (derivedRecommendedVersions.length) {
+        if (versionsThatReachedRecommended.length) {
             const { earliestVersionDate } = getVersionData(
-                derivedRecommendedVersions,
+                versionsThatReachedRecommended,
                 'recommendedPhaseReachedAt'
             );
             const phase = 'RECOMMENDED';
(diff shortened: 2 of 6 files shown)


client/components/CandidateReview/TestPlans/index.jsx
```diff
@@ -438,9 +438,7 @@ const TestPlans = ({ testPlanVersions }) => {
                         <thead>
                             <tr>
                                 <th>Candidate Test Plans</th>
-                                <CenteredTh>
-                                    Candidate Phase Start Date
-                                </CenteredTh>
+                                <CenteredTh>Plan Last Updated</CenteredTh>
```
alflennik, line 441: How do you feel about dropping the word "Plan", leaving just "Last Updated"? After all it's not the plan that's being updated here, but the overall candidate review process.

server/migrations/20240401203810-fixCommandButtonPhaseChangeDate.js
```diff
@@ -0,0 +1,61 @@
...
+                 where id = 62353
+                   and "gitSha" = '565a87b4111acebdb883d187b581e82c42a73844'
+                   and directory = 'command-button'`,
+                {
+                    transaction
+                }
+            );
+
+            // Update deprecation date for Command Button V23.12.06 to be a
+            // second before it's current time. It was the same time Command
```
alflennik, line 21:
```suggestion
            // second before its current time. It was the same time Command
```

client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -425,12 +425,53 @@ const DataManagementRow = ({
...
+                            !!draftPhaseReachedAt &&
+                            !candidatePhaseReachedAt &&
+                            !recommendedPhaseReachedAt
+                    );
+            }
+        };
+
+        const derivedRecommendedVersions = deriveVersionsByDates('RECOMMENDED');
+        const derivedCandidateVersions = deriveVersionsByDates('CANDIDATE');
+        const derivedDraftVersions = deriveVersionsByDates('DRAFT');
```
alflennik, line 467: Just an idea to call this `versionsThatReachedDraft`? And `versionsThatReachedRecommended`? When I read the code with the word "derived" I originally felt unsure about why this value needed to be derived and I didn't get the connection to the deprecated phase.

Likewise, for `deriveVersionsByDates` could we call that `getVersionsThatReachedPhase`?

client/components/DataManagement/DataManagementRow/index.jsx
```diff
@@ -425,12 +425,53 @@ const DataManagementRow = ({
...
+         * Uses the truthy state of defined dates for a TestPlanVersion to
+         * determine if the version was ever in that phase at any given point.
+         * Useful if the phase of the TestPlanVersion is currently DEPRECATED.
+         * @param {"RECOMMENDED"|"CANDIDATE"|"DRAFT"} phase
+         * @returns {TestPlanVersion[]}
+         */
+        const deriveVersionsByDates = phase => {
+            switch (phase) {
+                case 'RECOMMENDED':
+                    return testPlanVersions.filter(
```
alflennik, line 438: I verified that this is working, and that's the important thing, but I was wondering, looking down at the query below, how does this work when the testPlanVersions variable apparently doesn't include deprecated versions?

alflennik (commented):
I verified that the command button is fixed as well as the column in the candidate review page and the date in the data management page. Nice work! Once you answer my questions I'll approve and merge.

client/components/CandidateReview/TestPlans/index.jsx
howard-e: Good point and maybe either could work. It was also explained the same way with the rename suggestion from #973.

Given the context of the page and table, it could be clearer without the 'Plan' or need of the accompanying explanation. What do you think @mcking65?

client/components/DataManagement/DataManagementRow/index.jsx
howard-e: Sure, +1 for the consistency. 'Reached' is the verb that's currently being used in that context.

client/components/CandidateReview/TestPlans/index.jsx
howard-e, line 441:
```suggestion
                                <CenteredTh>Last Updated</CenteredTh>
```

client/components/DataManagement/DataManagementRow/index.jsx
alflennik: Thanks for clarifying offline.

alflennik (approved):

client/components/CandidateReview/TestPlans/index.jsx
mcking65: I went back and forth on this. I am OK with dropping "Plan" and seeing how it is understood by users as we work with them in the candidate review process.

When considering the label of the date column, one possible source of confusion that I considered is whether users might wonder if the date referred to the content of the plan or the status of the candidate review. Since the label of each row is essentially the first column, and since the date is the second column, it seems unlikely that people would interpret the date as specifying when status of work within the candidate review process last changed.

PR #995, opened 2024-04-02
Adjust connection pooling limit

alflennik:
Increasing the connection limit fixed the issue we were seeing about too many simultaneous connections but whenever the pooling function is needed it immediately errors without timing out with a complaint about exceeding the number of superuser connections. The solution was to lower the connection count slightly so that connection pooling can work again. That way we keep the increased number of simultaneous transactions while also keeping the ability to use connection pooling during intense workloads like the tests.

### DIFF
--- a/server/models/index.js
+++ b/server/models/index.js
@@ -18,7 +18,7 @@ sequelize = new Sequelize(
             native: true
         },
         pool: {
-            max: 100
+            max: 80
         },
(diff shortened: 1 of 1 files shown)


howard-e (approved):
Thanks for continuing to do the research to make this less flaky. Confirmed this **surprisingly** fixed it for me locally.

Issue #791, opened 2023-09-26
Update test queue to support generating reports that require a specific AT or browser version


mcking65:
#### ARIA-AT goal

One goal of the ARIA-AT working mode specific to test plans in the recommended phase is to keep their interop reports current. That means, for each AT/Browser combo that has a report, generate a new report when a new version of an AT or browser is released. In addition to keeping data current, over time, this will also create data that can be used to publish trend charts.

The CG has not yet defined criteria for which releases of AT or browsers would updates. That criteria could change over time.

As long as reporting involves a substantial amount of manual work, the community group may choose to be fairly selective about the releases for which new reports are required. For example, we could require new reports only for major AT releases and not for new browser releases. Any recent browser version could be considered sufficient in that scenario.

#### Test queue design update objective

Enable the test queue to include test plan runs that require a specific version of an AT or browser. If a test plan run does not require a specific version of either AT or browser, clearly indicate that any recent version is sufficient.

#### Design discussion

Should we fit version requirements into the current tables? Or, should we use this issue to drive a redesign of the information architecture of the queue?

Expedient option A: Add a table column "Version Requirements". Show the AT /browser name and versions required.

Examples:
* JAWS Any Recent/ChromeAny recent 
* JAWS 2023.2303.144 /Chrome Any Recent
* JAWS 2023.2303.144.400/Chrome 116.0.5845.188

Note that when specific versions are required, opening the test plan should require the user to confirm their versions match the requirements.

Option B: Consider restructuring the information on the test queue page to better align with how the CG has been working, e.g., a heading for each test plan and a table with columns for AT, Browser, assignees, status, and actions. The version information could be specified in the AT and Browser columns.

howard-e:
Discussions outside this issue and also described at https://github.com/w3c/aria-at-app/issues/792#issuecomment-1971331518 note that specifying the browser version will not be required for the short term.

mcking65:
Since encountering the conditions that led to w3c/aria-at#1047, and recalling comments that @jscholes  has made in the past about clearly defining "recent", I want to open discussion about changing "any recent" to "VX or Later".

In other words, when adding a run to the queue in draft or candidate review, the admin would be able to choose between specifying either a minimum version or a required version for an AT. Then in the test queue table, for each run, the AT and version cell in the table would say something like "JAWS 2024.2310.70.400 or Later" or "JAWS 2024.2310.70.400".

This would eliminate ambiguity about version requirements.

[...]

Issue #792, opened 2023-09-26
Update report status dialog for recommended test plans to include information about required AT/Browser versions


mcking65:
The following requirement is listed in #648 under the heading "Test plan report status dialog Description":

>TODO: If the phase is recommended, the algorithm for determining what rows to put in the table will be described in the next update of this specification.

This issue will complete that specification and track development of the requisite changes to the report status dialog.

#### ARIA-AT goal

One goal of the ARIA-AT working mode specific to test plans in the recommended phase is to keep their interop reports current. That means, for each AT/Browser combo that has a report, generate a new report when a new version of an AT or browser is released. In addition to keeping data current, over time, this will also create data that can be used to publish trend charts.

As of September 27, 2023, the CG plans to update recommended reports for each public release of a supported screen reader; see [Minutes for ARIA and Assistive Technologies Community Group Weekly Teleconference – 27 September 2023](https://www.w3.org/2023/09/27-aria-at-minutes.html#t04). Until the ARIA-AT project has sufficient automation and compute resources, new browser releases will not trigger report updates. The browser requirement will be "Any Recent Version". In the event of report run conflicts that are attributable to browser version differences, data from the more recent browser version will be published.

#### Report status dialog design update objective

When viewing the report status for a recommended test plan, make it easy to determine which AT/Browser version combinations are missing reports and add them to the test queue.

#### Test queue dependency

The test queue does not currently support version-specific test plan runs. Issue #791 must be resolved at the same time or before this issue.

#### Primary test plan run dependency

The meet the data processing requirements of this solution, the system will need to enable admins to mark one of the test plan runs as "primary" when finalizing a report. The requirements for updates to the report finalization process are defined in issue ...

#### Definition of "First Required AT Versions" for a test plan

**Incomplete DRAFT CONTENT**

The goal is to update reports for all AT versions that are newer than the versions used to generate the data that supported the candidate review process of the test plan. In other words, if new AT versions were released during the review process for which reports were not generated, reports using those versions will need to be generated after the test plan becomes recommended. This prevents important gaps in the history of the AT support levels.

Thus, each recommended test plan will have its own set of "First Required AT Versions" or "First Required AT Version Dates". Whether date or version is used is an implementation detail; either will do. For the purpose of the following definition, version number is used.

A test plan has a "First Required AT Version" for each AT that meets the following two criteria:
* The AT is covered by the test plan
* the AT is designated as requiring reports for recommended test plans

[...]

PR #997, opened 2024-04-03
Add Minimum / Exact AT Version to Test Plan Reports

alflennik:
See issue https://github.com/w3c/aria-at-app/issues/791 - this PR addresses the server part of this task.

- Adds minimumAtVersionId and exactAtVersionId to the TestPlanReports table and allows those fields to be passed in through the whole server from GraphQL. Note there is not currently any UI to provide these fields.
- Removes limitation on duplicate TestPlanReports for a single AT / Browser combination. Since the frontend can't accommodate duplicates yet I made sure there is still no way for the frontend to generate any duplicates.
- Removes some old migrations and endpoints that are no longer needed that would be incompatible with this change.


client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -149,24 +149,29 @@ function AddTestToQueueWithConfirmation({
...
-                            } else {
-                                await addTestToQueue();
-                            }
-                        }
-                    }
-                ]}
+                // The proceed button is disabled because it will now create
+                // duplicate testPlanReports, and support has not yet been
+                // implemented across the full frontend to account for
+                // duplicates.
```
howard-e, line 160: Probably want to defend against this on the backend as well.

client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -149,24 +149,29 @@ function AddTestToQueueWithConfirmation({
                 content={
                     'The report could not be created because an existing ' +
                     'report was found on the reports page with the same AT, ' +
-                    'browser and test plan version. Would you like to return ' +
-                    'the existing report back to the test queue?'
+                    'browser and test plan version. Work is currently ' +
+                    'underway to remove this limitation.'
```
howard-e, line 153: This makes me wonder if we don't actually want to include this change in `main` until the rest of #791 and #792 is also ready. So we can instead treat this as a current dev "base"? Do you have any thoughts on that?

howard-e (commented):
Looks good and a promising start to an upcoming big process change here! Exciting! It's also great to see a lot of unnecessary code being removed.

I've left a few inline comments.

client/components/Home/Home.jsx
```diff
@@ -178,9 +178,9 @@ const Home = () => {
                     <p>
                         Enabling AT interoperability is a large, ongoing
                         endeavor that requires industry-wide collaboration and
-                        support. The W3C ARIA-AT Community Group is focusing on
-                        a stable and mature test suite for five screen readers
-                        by the end of 2023. In the future, we plan to test
+                        support. The W3C ARIA-AT Community Group is currently
+                        focusing on a stable and mature test suite for five
+                        screen readers. In the future, we plan to test
```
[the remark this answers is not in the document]
alflennik: Okay I created. https://github.com/w3c/aria-at-app/pull/999 Who do you think should review it?

server/models/services/TestPlanReportService.js
```diff
@@ -215,7 +215,14 @@ const getTestPlanReports = async ({
...
  */
 const createTestPlanReport = async ({
-    values: { testPlanVersionId, atId, browserId },
+    values: {
+        testPlanVersionId,
+        atId,
+        browserId,
+        exactAtVersionId,
+        minimumAtVersionId,
+        vendorReviewStatus
```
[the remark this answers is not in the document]
alflennik: Oh you're right, I removed it. Previously I got confused by code in the `updateTestPlanReportTestPlanVersion` file and that was why I abortively changed this function.

client/components/AddTestToQueueWithConfirmation/index.jsx
alflennik: Sure, I added temporary code to enforce that.

client/components/AddTestToQueueWithConfirmation/index.jsx
alflennik: I do think it might become necessary at some point to have a branch for trend report changes. However personally I feel like this code is still mergeable to main, and we haven't arrived at the "ok now we need a separate branch" point. 

Other than removing the "proceed" button, I don't think there's much impact on the user experience. In the event the admins do want to move a report from the reports page back to the test queue, we can accommodate that through the API.

That's my perspective but I'm happy to defer to whatever you think is better.

Do you think removing the "work is underway" sentence would mitigate your concern?

alflennik:
@howard-e should be ready for a second pass.

server/migrations/20230608171911-moveTestPlanReportValuesToTestPlanVersion.js
```diff
@@ -1,634 +0,0 @@
-'use strict';
```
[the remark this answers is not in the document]
alflennik: felt good!

howard-e (approved):
@alflennik thanks for addressing the feedback! This is great and all very exciting for the next phase or process change!

Changes LGTM and I'm approving but also note that I don't think we should push to `main` just yet for reasons described in https://github.com/w3c/aria-at-app/pull/997#discussion_r1552281387

client/components/AddTestToQueueWithConfirmation/index.jsx
alflennik: @howard-e do you want to merge it into a branch with a nice name like "trends"?

howard-e: Sure!

alflennik:
This has been superseded by https://github.com/w3c/aria-at-app/pull/1055!

PR #999, opened 2024-04-03
Tweak home copy

alflennik:
I noticed this sentence read strange in 2024 so I updated it!


howard-e (approved):
@alflennik thanks for noticing and fixing this! Agreed internally that this copy change is okay

Issue #818, opened 2023-10-23
Automation: Research protocol for communicating “per-test” status updates


ccanash:
The objective of this tasks is to research and define what the best protocol will be to communicate "per-test"status updates.

gnarf:
In order to bring anyone following this issue up to speed, I'm going to post a quick summary of the API in the app for storing statuses / automation jobs (as is).

We have defined a `CollectionJobStatus` type that has the following values: `QUEUED`, `RUNNING`, `COMPLETED`, `CANCELED`, `ERROR`.

The `CollectionJob` itself has an `externalLogsUrl` string state to store additional log file link.

## App API: `POST /jobs/:jobId/update`
### Update job status for entire collection job

Parameters:
* `status` - a CollectionJobStatus
* `externalLogsUrl` - to update the log url if you have it - only updates when present, empty key is okay

This updates the overall collection job status, this currently isn't displayed anywhere in the app, but is used to clue some calculations based on CANCELLED and QUEUED.

## App API: `POST /jobs/:jobId/result`
### Send test results for a specific test in the job
Note: Only works when job status is `RUNNING`

Parameters:
* `testCsvRow` or `presentationNumber` - the v1 or v2 identifier for the test in the job
* `responses` - an array of strings - one collected response for each action in the test
* `capabilities` - `{ atName, atVersion, browserName, browserVersion }` to store which browser/at combo has generated results.

This endpoint collects and stores the responses, and checks for "historical responses" match to promote the old assertion values that it can copy.

## Displaying the status in the app
![image](https://github.com/w3c/aria-at-app/assets/549355/6f233713-42e0-4b85-a1bb-557e397edb37)

The data for this is currently coming from this query:

```graphql
    query TestPlanRunsTestResults($testPlanReportId: ID!) {
        testPlanRuns(testPlanReportId: $testPlanReportId) {
            id
            tester {
                username
            }
            testResults {
                id
                scenarioResults {
                    assertionResults {
                        passed
                        failedReason
                    }
                }
            }
        }
    }
```

* "Completed" is increased by 1 for every test result existing in the `testResults` 
* "Queued" is the number of totals tests minus the "Completed" count (if job has any status other than `CANCELLED`
* "Cancelled" ^^^ but when cancelled.

-- Proposal for updates coming in next message

gnarf:
# Proposal for updates

## App/Database Side:
[...]

PR #1000, opened 2024-04-04
Collection Job - Per-test status updates

gnarf:
Implementing the updates for #818 

---
  - https://app.asana.com/0/0/1206980916128609


server/models/CollectionJob.js
```diff
@@ -45,6 +45,12 @@ module.exports = function (sequelize, DataTypes) {
             sourceKey: 'testPlanRunId',
             as: 'testPlanRun'
         });
+
+        Model.hasMany(models.CollectionJobTestStatus, {
+            as: 'testStatus',
+            foreignKey: 'collectionJobId',
+            sourceKey: 'id'
+        });
```
alflennik, line 53: Do we want the collection job to be deleted when the test plan run is deleted?

server/migrations/20240404171101-addCollectionJobTestStatus.js
```diff
@@ -0,0 +1,58 @@
...
+                    },
+                    testId: {
+                        type: Sequelize.STRING,
+                        allowNull: false
+                    },
+                    collectionJobId: {
+                        type: Sequelize.INTEGER,
+                        allowNull: false,
+                        onDelete: 'CASCADE',
+                        onUpdate: 'CASCADE',
```
alflennik, line 23: Interesting feature! Definitely going to file this trick away in my brain.

alflennik (approved):
I mostly reviewed this from the perspective of GraphQL, the database and the general common practices on the backend. On those accounts, it looks great. I also did try to put on an automation hat and I was able to successfully get "QUEUED", "COMPLETED", "CANCELLED" statuses, but I would appreciate if the other reviewers could give the automation side a good thorough test. I didn't see any unexpected behavior.

It will be great to hugely reduce the number of server requests soon!

server/models/CollectionJob.js
gnarf: I debated this myself - we have some recovery potential via the database if not, but we might want to delete them

server/controllers/AutomationController.js
```diff
@@ -99,6 +100,23 @@ const updateJobStatus = async (req, res) => {
         ...(externalLogsUrl != null && { externalLogsUrl })
     };
 
+    // When new status is 'COMPLETED' or 'ERROR' or 'CANCELLED'
+    // update any CollectionJobTestStatus children still 'QUEUED' to be 'CANCELLED'
+    if (
+        status === COLLECTION_JOB_STATUS.COMPLETED ||
+        status === COLLECTION_JOB_STATUS.CANCELLED ||
+        status === COLLECTION_JOB_STATUS.ERROR
+    ) {
```
jugglinmike, line 109: A utility function like `isFinal` (maybe defined as a static method on the new model) would surface the intent and make this more readable.

jugglinmike (changes requested):
I'd like to formalize the distinction between "new way" and "old way" a bit. Doing so will further clarify the differences, give us more confidence when removing the deprecated logic in the future, and reduce the number of unnecessary code paths in the interim.

Basically, if we start with something like,

```javascript
const isEnhanced = 'testRowNumber' in req.params;
```

Then we could switch on `isEnhanced`. (That name could probably be improved, though.)

jugglinmike:
I think that could work. Don't we also have to update [the callback URLs](https://github.com/w3c/aria-at-app/blob/8b46827288eaea1a72c583199f6303c47933d0ff/server/services/GithubWorkflowService.js#L131-L132)?

gnarf:
I realized while updating the PR to remove the back-compat URLs that some mock work I needed had already been done on the working UI branch so I merged that into this one instead of keeping them separate. I could tease the frontend updates out of this branch if people still prefer.  The changes from [d87280b](https://github.com/w3c/aria-at-app/pull/1000/commits/d87280b44d672dff3f58587f222291be3d988b1f) adds the frontend for the test queue and manage nvda run dialogs and some ways to test other status.

gnarf:
Alright - front end tests are now passing as well, so this should be all good - I removed the back-compat URLs as well because the automation harness is properly working with the new style URLs already.

alflennik (approved):
So this is a huge improvement vastly reducing the load on the server and preventing issues we've already seen that took down the server.

I don't see any reason not to get this in, it looks great and I see a lot of small improvements sprinkled here and there which is great.

I verified that it works as intended with both the mock and genuine automation systems.

I feel like in a future update it might be nice to improve a few quirks of the system as it stands now. For example I noticed it doesn't check for status updates on the test run page. This is definitely an improvement we can consider separately and not a blocker.

Secondly, there are still multiple requests happening every second (I see a "TestPlanRunAssertionResultsByTestPlanRunId" query as well as the "TestPlanRunsTestResults" query) and I feel like the architecture of GraphQL should theoretically allow them to all be combined into one. Again, not a blocker.

The main thing which is worth looking at down the line, but not a blocker, is that fact that the system still makes one request for every active collection job in progress. It feels like this will only be an issue if the admins are trying to test every dang test plan that exists, all at once, and while that doesn't seem likely to happen anytime soon, it is pretty conceivable that it could start to happen eventually. That would be a nice enhancement, but this PR solves the much more serious issue of making one request per test.

Again these notes really have nothing to do with the PR itself, which is all about "per test status updates." So I'm definitely happy to approve.

This ended up being a pretty large-scale change, which was not the expectation going into it, so nice job chasing it down through the whole stack and seemingly every system we got 🙏

jugglinmike (approved):
Way to go, @gnarf!

PR #1001, opened 2024-04-04
Specify primary Test Plan Run when finalizing report and track first required AT Version

howard-e:
This PR adds support for selecting a **"Primary test plan run"** as outlined in #792.

* Updates `TestPlanRun` table to track `isPrimary` as an attribute
* Updates `markAsFinal` resolver to take a `primaryTestPlanRunId` which indicates which test run a report should be considered the "primary"
* Adds a dialog that pops up when marking a test plan report as final on the Test Queue for an admin. It allows them to select from the possible options which tester's run should be the primary

Note: This PR also includes #1051 which was reviewed and approved separately


client/components/TestQueueRow/index.jsx
```diff
@@ -370,12 +391,69 @@ const TestQueueRow = ({
...
+    };
+
+    const renderPrimaryRunSelectionDialog = testers => {
+        return (
+            <BasicModal
+                show={showPrimaryTestPlanRunModal}
+                title="Select Primary Test Plan Run"
+                content={
+                    <>
+                        When a tester&apos;s run is marked as primary, it means
```
evmiguel, line 406: Can we style this at the break point to have whitespace?

client/components/TestQueueRow/TestQueueRow.css
```diff
@@ -77,6 +77,10 @@ button.more-actions:active {
     width: initial;
 }
 
-/* tr.test-queue-run-row td:first-child {
-    padding: 0.75rem;
-} */
+.primary-test-run-select {
+    padding: 0 !important;
```
evmiguel, line 81: Can we add a padding left to the select?

evmiguel (approved):
LGTM! Minor comments on the styling.

client/components/TestQueueRow/TestQueueRow.css
howard-e: I realize now what was being shown is because the dialog still appears when the number of testers to choose from is 1 and `<select>.size` would render differently.

I've pushed https://github.com/w3c/aria-at-app/pull/1001/commits/b738bc05a30fcee945642a0cae3ac2b6fd431315 which should won't display the dialog at all if there's only 1 tester to choose from. Thanks for pointing this out!

evmiguel:
Looks good @howard-e!

alflennik:
Following a discussion with @howard-e we are changing the base branch to https://github.com/w3c/aria-at-app/pull/1055

PR #1050, opened 2024-04-08
Add workflow to deploy to production

howard-e:
Add workflow to automate deployment to aria-at.w3.org on a push to the `main` branch.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -1,12 +1,12 @@
-name: Deploy to staging (aria-at-staging.w3.org)
+name: Deploy to production (aria-at.w3.org)
 
 on:
   push:
     branches:
-      - development
+      - main
 
 jobs:
-  deploy-staging:
+  deploy-production:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-production.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  deploy-production:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install SSH key for deploying
+        uses: shimataro/ssh-key-action@v2
+        with:
+          key: ${{ secrets.DEPLOY_SSH_PRIVATE_KEY }}
+          known_hosts: ${{ secrets.DEPLOY_KNOWN_HOSTS_PRODUCTION }}
```
alflennik, line 17: This variable doesn't seem to be defined in GitHub! Is that expected?

alflennik (commented):
Looks great, this type of automation will make the project a lot more approachable and easier to manage. See my one comment but other than that looks good.

.github/workflows/deploy-production.yml
howard-e: Added!

alflennik (approved):
Thanks for the fix!

PR #1051, opened 2024-04-09
Add resolver for tracking first required AT Version

howard-e:
This PR adds support for tracking **"First Required AT Versions** as outlined in #792.

* Includes `firstRequiredAtVersion` on graphql schema's `TestPlanVersion`. When provided an AT id, given that a report has been generated for that test plan version with that AT, it will follow defined rules to return the first required AT version for a TestPlanVersion once it has gotten to RECOMMENDED.

Since this is work is also dependent on a "primary test plan run", this is branched off of the work being done on #1001.

### DIFF
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -378,15 +378,26 @@ const graphqlSchema = gql`
         """
         testPlanReports(isFinal: Boolean): [TestPlanReport]!
         """
-        The first required AT Version of a specified AT used to collect results
-        for this TestPlanVersion.
+        For each report under this TestPlanVersion, if the report's combination
+        is indicated as required and the report is marked as final at the time
+        the TestPlanVersion is updated to RECOMMENDED then by checking the
+        testers' runs which have been marked as primary, the earliest found AT
+        version for the respective ATs should be considered as the
+        first required AT version or "firstRequiredAtVersion".
 
-        A first required AT Version is the earliest version used to collect
-        results for a test plan run marked as primary, and is a required report.
+        The "earliest" is determined by comparing the recorded AtVersions'
+        releasedAt value.
 
-        This should be used to prevent any older AT Versions from being used to
-        create subsequent reports after the Test Plan Version has reached
-        RECOMMENDED.
+        Required Reports definition and combinations are defined at
+        https://github.com/w3c/aria-at-app/wiki/Business-Logic-and-Processes#required-reports.
+
+        Primary Test Plan Run is defined at
+        https://github.com/w3c/aria-at-app/wiki/Business-Logic-and-Processes#primary-test-plan-run.
+
+        After this TestPlanVersion is updated to RECOMMENDED, this should be
+        used to ensure subsequent reports created under the TestPlanVersion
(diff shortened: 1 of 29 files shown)


alflennik:
Cannot read properties of undefined while querying the new field.

I got this error with the query
```
query { 
  testPlanVersions {
    id
    title
    versionString
    firstRequiredAtVersion(atId: 1) {
 			id
    }
  }
}
```

Any idea what might be causing this?

This is with a fresh database and the sample data.

server/graphql-schema.js
```diff
@@ -377,6 +377,18 @@ const graphqlSchema = gql`
...
+        for this TestPlanVersion.
+
+        A first required AT Version is the earliest version used to collect
+        results for a test plan run marked as primary, and is a required report.
+
+        This should be used to prevent any older AT Versions from being used to
+        create subsequent reports after the Test Plan Version has reached
+        RECOMMENDED.
+        """
+        firstRequiredAtVersion(atId: ID!): AtVersion
```
alflennik, line 391: I was wondering, how do you feel about calling this `earliestMissingAtVersion`? I feel like "missing" might be more descriptive than "required", possibly?

server/graphql-schema.js
```diff
@@ -377,6 +377,18 @@ const graphqlSchema = gql`
...
+        """
+        The first required AT Version of a specified AT used to collect results
+        for this TestPlanVersion.
+
+        A first required AT Version is the earliest version used to collect
+        results for a test plan run marked as primary, and is a required report.
+
+        This should be used to prevent any older AT Versions from being used to
+        create subsequent reports after the Test Plan Version has reached
+        RECOMMENDED.
```
alflennik, line 389: I appreciate the depth of this explanation, but even so I had a hard time understanding what this field really means. Would it be possible to describe it in terms of how exactly it fits into the workflow for test plans?

alflennik (changes requested):
Just a couple of notes! The main thing is that I encountered an error in the graphql playground.

howard-e:
> Cannot read properties of undefined while querying the new field.
> I got this error with the query
> 
> ```
> query { 
>   testPlanVersions {
>     id
>     title
>     versionString
>     firstRequiredAtVersion(atId: 1) {
>  			id
>     }
>   }
> }
> ```
> 
> Any idea what might be causing this?
> 
> This is with a fresh database and the sample data.

https://github.com/w3c/aria-at-app/pull/1051/commits/21a0c72ac25921d84de96cee99ad3125a41d3cfa should fix this. Because of how the `testPlanVersionsResolver` is done the, the `testPlanReports` and `phase` variables were being left out which would then throw the error.

server/graphql-schema.js
howard-e: Updated in https://github.com/w3c/aria-at-app/pull/1051/commits/b916cafdc60e66919b9306991eee53cf1e283970. Does that help to clarify?

server/graphql-schema.js
alflennik: Thanks!

PR #1052, opened 2024-04-09
Display AT Version for finalized reports when TestPlanVersion is RECOMMENDED

howard-e:
This PR now displays an AT version on the following report pages when the TestPlanVersion it is linked to is in the RECOMMENDED phase.

**`/report/<id>`**

* Includes AT version for each AT + Browser h2 heading

**`/report/<id>/targets/<id>`**

* Now includes AT Version in title of page
* Now includes AT Version in h1 on page
* Now includes AT Version in navigation breadcrumb for current location

This follows discussions from https://github.com/w3c/aria-at-app/issues/792#issuecomment-1971331518:
> * When a test plan version is promoted to recommended, the AT version should be included in the title (eg. from "Alert with NVDA and Firefox" to "Alert with NVDA 2023.3 and Firefox")
> * When upgrading a report to recommended, the first required AT version will be what’s displayed. Otherwise it will be what’s specified by the test admin when adding new reports to the test queue for a test plan that has already been recommended.


alflennik (approved):
Code looks good, tests are passing, verified the new UI is showing as intended, also nice job on the resolver that was really easy to follow. Happy to approve!

PR #1053, opened 2024-04-10
feat: Restructure APG Support Tables

Paul-Clue:
see issue #https://github.com/w3c/aria-practices/issues/2971

This pull request changes the structure of the support tables. The rows are now AT and browser combinations and the columns are now "MUST" and "SHOULD" assertions. The support percentage now represents the degree of support for "MUST" assertions and "SHOULD" assertions.

### DIFF
--- a/server/apps/embed.js
+++ b/server/apps/embed.js
@@ -220,7 +220,7 @@ const getLatestReportsForPattern = ({ allTestPlanReports, pattern }) => {
         latestReports
     };
 };
-const priorities = ['MUST', 'SHOULD'];
+const priorities = ['"Must" Assertion Priority', '"Should" Assertion Priority'];
 const renderEmbed = ({
     ats,
     allTestPlanReports,
--- a/server/apps/embed.js
+++ b/server/apps/embed.js
@@ -116,6 +116,7 @@ const queryReports = async testPlanDirectory => {
 // As of now, a full query for the complete list of reports is needed to build
 // the embed for a single pattern. This caching allows that query to be reused
 // between pattern embeds.
+
 const queryReportsCached = staleWhileRevalidate(queryReports, {
     millisecondsUntilStale
 });
@@ -137,7 +138,7 @@ const getLatestReportsForPattern = ({ allTestPlanReports, pattern }) => {
     let testPlanVersionIds = new Set();
     const uniqueReports = [];
     let latestReports = [];
-    // section:
+
     testPlanReports.forEach(report => {
         allAts.add(report.at.name);
         allBrowsers.add(report.browser.name);
@@ -175,14 +176,7 @@ const getLatestReportsForPattern = ({ allTestPlanReports, pattern }) => {
 
         testPlanVersionIds.add(report.testPlanVersion.id);
     });
-    // section:
-    // const numberArray = allAts.map((at) => {
-    //     return allBrowsers.map((browser) => {
-    //         return {at, browser}
-    //     })
-    // })
-    // const numberArray =
-    // console.log(numberArray);
+
     uniqueReports.forEach(group => {
         if (group.length <= 1) {
             latestReports.push(group.pop());
@@ -196,7 +190,6 @@ const getLatestReportsForPattern = ({ allTestPlanReports, pattern }) => {
                 .pop();
 
             latestReports.push(latestReport);
-            // use latest reports for the loop for the table
         }
     });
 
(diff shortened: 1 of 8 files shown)


server/handlebars/embed/views/main.hbs
```diff
@@ -37,78 +37,36 @@
...
                         <td></td>
-                        {{#each allBrowsers}}
+                        {{#each priorities}}
                         <th>{{this}}</th>
                         {{/each}}
                     </tr>
                 </thead>
                 <tbody>
-                    {{#each reportsByAt}}
+                    {{log latestReports}}
```
alflennik, line 46: Some debugging code here

alflennik:
So I think the design still looks pretty good, but I was wondering @isaacdurazo, given that this is by far the most public representation of the ARIA-AT projects work, if you have any ideas to spruce it up? One possible tweak might be "MUST" and "SHOULD" being all caps, in the past we've gotten accessibility notes about all caps words being hard for people with dyslexia to read. I do wonder if "Must Assertions" or something like that might be easier to understand? But the main thing motivating me to ask for your input is just the `VoiceOver for macOS 14.2 - Safari` part, is seems like we could improve it somehow.

What the new embed looks like embedded in the APG website.

server/apps/embed.js
```diff
@@ -42,6 +42,30 @@ const queryReports = async () => {
...
+                            }
+                            latestAtVersionReleasedAt {
+                                id
+                                name
+                                releasedAt
+                            }
+                        }
+                    }
+                }
                 testPlanReports(
```
alflennik, line 69: Is it possible to remove this part of the query?

server/apps/embed.js
```diff
@@ -233,10 +256,12 @@ const renderEmbed = ({
         allAtBrowserCombinations,
         title: queryTitle || title || 'Pattern Not Found',
         pattern,
+        priorities,
         phase,
         allBrowsers,
```
alflennik, line 261: There's a lot of dead code still in this file, like this variable is not used anywhere for example. Would it be possible to simplify this file a bit? It's currently getting pretty intimidating. Acknowledging of course that I contributed more than anyone to the current situation.

alflennik (changes requested):
Fantastic work, it's working perfectly. I was even able to test it integrated into a development version of the APG website. I do have a couple notes on the implementation itself. And I pinged Isaac in a comment to weigh in on the design given that this is arguably the most important page in the app.

server/handlebars/embed/views/main.hbs
Paul-Clue: This is done.

server/apps/embed.js
Paul-Clue: I don't think so. `testPlanReports` is being used in the code.

server/apps/embed.js
alflennik: So yes the variable appears 7 times in this file, but all of those uses are only related to setting the variable, not using it. Then if you look at the helpers file and main.hbs you'll see that this variable is passed around a bit but never used. That's what I mean when I say it's dead code.

The larger issue, beyond this one variable, is that a lot of the code in this file is passed around and set but never actually utilized, and I think it's definitely worth the effort to overhaul the whole file and remove all the dead code.

server/apps/embed.js
alflennik: As I said in another comment, you're right that it's being used. My point is more that it's only being used by dead code and the whole network of dead code present in this file can be removed.

server/handlebars/embed/views/main.hbs
alflennik: Thx!

Paul-Clue:
This pull request is ready for review.

ccanash:
Latest submitted PR addresses https://github.com/w3c/aria-practices/issues/3017

alflennik (approved):
Fantastic work!

Please don't merge this yet since we are still awaiting feedback from the APG task force.

howard-e:
The APG TF [discussed this on May 21, 2024](https://www.w3.org/2024/05/21-aria-apg-minutes.html#t03) and confirmed this work is now ready to be merged. I will be merging this later today.

mcking65:
@howard-e 
I thought the tables would update on the APG site when this merged. The old tables are still being rendered. Do we need to change something in the APG content?

howard-e:
@mcking65 this work being merged into the `development` branch makes it immediately available for the staging environment. It will become available to the production environment that APG targets once this is merged to `main` but no content change will be required of the APG then.

PR #1055, opened 2024-04-22
Minimum and Exact AT Version

alflennik:
Adds a new "minimum" and "exact" AT Version picker to the "Add to test queue" UI. This is part of https://github.com/w3c/aria-at-app/issues/791.

Note that the test queue status dialog has not yet been updated and is not expected to work.

This PR resolves also conflicts with the development branch that currently prevent merging `can-duplicate-test-plan-reports`. Therefore this diff includes all the changes from that branch. Sorry for the noise. If we use `can-duplicate-test-plan-reports` as a base the number of files changed would actually increase.


client/components/common/RadioBox/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+                            {label}
+                        </TextThatWontShiftWhenBold>
+                    </Label>
+                );
+            })}
+        </ContainerDiv>
+    );
+};
+
+const TextThatWontShiftWhenBold = ({ isBold, children: text }) => {
```
howard-e, line 58: Descriptive!

server/graphql-schema.js
```diff
@@ -1015,8 +1015,10 @@ const graphqlSchema = gql`
     input TestPlanReportInput {
         testPlanVersionId: ID!
         atId: ID!
+        exactAtVersionId: ID
+        minimumAtVersionId: ID
         browserId: ID!
-        copyResultsFromTestPlanReportId: ID
+        copyResultsFromTestPlanVersionId: ID
```
howard-e, line 1021: Another good catch

howard-e:
To reviewers: In a separate discussion with @alflennik we agreed this now introduces breaking changes around the Test Plan Report status dialog. I've added a `do not merge` label to reflect that. So until additional parts of #791's and #792's implementation requests are ready, this PR should not yet be merged. This will be treated as a current "base" for those issues.

Currently #1001 is also expected to branch off of this work.

PR #1056, opened 2024-04-23
Redesign support table warning

Paul-Clue:
see issue [#212 ](https://github.com/w3c/wai-aria-practices/issues/212) (in wai-aria-practices)

This pull request changes the warning at the top of the APG support table so that it matches the style of the warnings on the rest of the page.

### DIFF
--- a/server/handlebars/embed/public/style.css
+++ b/server/handlebars/embed/public/style.css
@@ -96,7 +96,7 @@ body {
     padding-right: var(--left-right-padding);
   }
   
-  #candidate-title > span > h1 {
+  #candidate-title > span > h4 {
     font-family: Arial, Helvetica, sans-serif;
     font-size: 12px;
     font-weight: bold;
--- a/server/handlebars/embed/public/style.css
+++ b/server/handlebars/embed/public/style.css
@@ -61,7 +61,7 @@ body {
     background-repeat: no-repeat;
     position: absolute;
     right: 0.5em;
-    top: 0.5em;
+    top: 1.5em;
     background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='12' height='6.861' viewBox='0 0 12 6.861'%3E%3Cpath id='Icon_ionic-ios-arrow-down' data-name='Icon ionic-ios-arrow-down' d='M12.19,16.039,16.727,11.5a.854.854,0,0,1,1.211,0,.865.865,0,0,1,0,1.215L12.8,17.858a.856.856,0,0,1-1.183.025L6.438,12.717A.858.858,0,1,1,7.649,11.5Z' transform='translate(18.188 18.108) rotate(180)' fill='%2360470c'/%3E%3C/svg%3E%0A");
     background-color: #fce1a4;
     transform: rotate(180deg);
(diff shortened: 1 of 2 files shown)


server/handlebars/embed/views/main.hbs
```diff
@@ -9,7 +9,7 @@
     {{#unless dataEmpty}}
         {{#if (isCandidate phase)}}
         <details id="embed-report-phase-container">
-            <summary id="candidate-title"><span>Warning! Unapproved Report</span></summary>
+            <summary id="candidate-title"><span><span>Warning! Unapproved Report</span></span></summary>
```
evmiguel, line 12: I think this could be improved semantically with a heading and a span instead of two spans.

server/handlebars/embed/public/style.css
```diff
@@ -41,45 +41,77 @@ h3#report-title {
...
     padding: 0.5em 0.5em 0.5em 1em;
     cursor: pointer;
-}
-
-#candidate-title > span {
+    --space-between-elements: 1em;
+    --left-right-padding: 55px;
+  }
+  
+  #candidate-title > * {
```
evmiguel, line 90: I think instead of applying this to all the children, make this individual margin left and right on the span and icon. I also think that the space between the icon and span can be smaller.

evmiguel (changes requested):
Great work, @Paul-Clue! I have some requests in the comments.

server/handlebars/embed/views/main.hbs
```diff
@@ -9,7 +9,7 @@
     {{#unless dataEmpty}}
         {{#if (isCandidate phase)}}
         <details id="embed-report-phase-container">
-            <summary id="candidate-title"><span>Warning! Unapproved Report</span></summary>
+            <summary id="candidate-title"><span><h1>Warning! Unapproved Report</h1></span></summary>
```
alflennik: Could we make this an h4 so it matches the flow of headings in the APG website?

server/handlebars/embed/views/main.hbs
Paul-Clue: This has been taken care of.

server/handlebars/embed/public/style.css
Paul-Clue: This has been taken care of.

server/handlebars/embed/views/main.hbs
Paul-Clue: This has been taken care of.

evmiguel (approved):

alflennik:
Code-wise it looks great. It's great to be able to use the exact same styling in both places. However there are some differences still. Take a look at the screenshot.

Comparison of existing disclosure and the iframe disclosure

I think we are not currently using the same font, font-size and the border width looks different. Also I'm not 100% convinced by the pill-shaped background on the h4, something feels off about it and I feel like something more similar to the other disclosure might be preferable.

alflennik (changes requested):
Code looks great! See comment above for a couple of enhancements I requested.

alflennik (approved):
Looks fantastic! Thanks for taking the time to get it so polished.

PR #1061, opened 2024-04-24
Use assertion phrase in test result table

Paul-Clue:
see issue #744 

This pull request adds the assertion phrase instead of the assertion statement to the test results table.

### DIFF
--- a/client/components/TestRun/queries.js
+++ b/client/components/TestRun/queries.js
@@ -204,7 +204,6 @@ export const TEST_RUN_PAGE_ANON_QUERY = gql`
                         id
                         text
                         phrase
-                        phrase
                     }
                 }
                 conflictingResults {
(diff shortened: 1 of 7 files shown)


client/components/TestRun/queries.js
```diff
@@ -198,6 +203,8 @@ export const TEST_RUN_PAGE_ANON_QUERY = gql`
                     assertion {
                         id
                         text
+                        phrase
```
evmiguel, line 206: Will the query work if you delete the duplicate phrase?

client/components/TestRun/queries.js
```diff
@@ -766,6 +784,8 @@ export const SAVE_TEST_RESULT_MUTATION = gql`
                             assertion {
                                 id
                                 text
+                                phrase
```
evmiguel, line 787: Same comment about the duplicate phrase.

evmiguel (changes requested):
@Paul-Clue great work! It works well for me, but I do notice that the assertion phrase is not capitalized. Could you make sure that it is capitalized?

PR #1064, opened 2024-04-29
Add support in import test script for importing test harness

evmiguel:
This PR adds support for importing the test harness into the ARIA-AT App repo from the ARIA-AT repo. To test, run:

```
yarn db-import-tests:dev;
```

Verify that the harness files are in `client/resources`. Additionally, this PR sets the `commands.json` in `server/resources` to `commandsV1.json`. My intent was to make the import minimally invasive. Open to other thoughts!


.gitignore
```diff
@@ -125,3 +125,5 @@ server/migrations/test_plan_target_id.csv
 
 # Private Key files (installed by deploy)
 jwt-signing-key.pem
+
+client/resources
```
howard-e, line 129: Since there are still updates to `client/resources` in this PR, you'll also need to untrack `client/resources` by doing: `git rm -r --cached client/resources`, and push again afterwards.

Although, I am now wondering if this should actually needs to be done? It will require that contributors be *somewhat* aware of when last changes to the `w3c/aria-at` files have happened so they may re-run the import script locally. Otherwise there could be confusion if some `w3c/aria-at`-dependent operations aren't working as expected if the client application's files are updated but not the now ignored aria-at harness changes.

My above concern is relatively minor though and seems like a `docs/local-development.md` update could be useful here.

client/resources/aria-at-test-io-format.mjs
```diff
@@ -1455,7 +1455,7 @@ export class TestRunInputOutput {
         output: command.atOutput.value,
         assertionResults: command.assertions.map(assertion => ({
           assertion: {
-            priority: assertion.priority === 1 ? 'MUST' : 'SHOULD',
+            priority: assertion.priority === 1 ? 'REQUIRED' : 'OPTIONAL',
```
howard-e, line 1458: Was wondering why this was getting updated in this PR and realized this change was never captured in `w3c/aria-at` where these files should have been copied over from. So I've submitted https://github.com/w3c/aria-at/pull/1056 to make sure this change is captured for the future.

server/resources/commandsV1.json
```diff
@@ -0,0 +1,530 @@
+[
```
howard-e, line 1: You can also remove `server/resources/commands.json` in this PR too

server/resolvers/helpers/retrieveCommands.js
```diff
@@ -1,4 +1,4 @@
-const commands = require('../../resources/commands.json');
+const commands = require('../../resources/commandsV1.json');
```
howard-e, line 1:
```suggestion
const commandsV1 = require('../../resources/commandsV1.json');
```

server/scripts/import-tests/index.js
```diff
@@ -251,6 +253,38 @@ const readDirectoryGitInfo = directoryPath => {
     return { gitSha, gitMessage, gitCommitDate };
 };
 
+const importHarness = () => {
+    const sourceFolder = path.resolve(`${testsDirectory}/resources`);
+    const targetFolder = path.resolve('../', 'client/resources');
+    console.info('Clearing harness directory ...');
```
howard-e, line 259:
```suggestion
    console.info(`Updating harness directory, ${targetFolder} ...`);
```

This and the following message suggestions are nitpicks and you can feel free to ignore the exact wording although I do feel better about seeing an explicit path being shared in operations like this.

server/scripts/import-tests/index.js
```diff
@@ -251,6 +253,38 @@ const readDirectoryGitInfo = directoryPath => {
...
+    fse.rmSync(targetFolder, { recursive: true, force: true });
+
+    // Copy source folder
+    console.info('Importing harness ...');
```
howard-e, line 263:
```suggestion
    console.info('Importing latest harness files ...');
```

server/scripts/import-tests/index.js
```diff
@@ -251,6 +253,38 @@ const readDirectoryGitInfo = directoryPath => {
...
+    const supportJson = 'support.json';
+    fse.copyFileSync(
+        `${testsDirectory}/${commandsJson}`,
+        `${targetFolder}/${commandsJson}`
+    );
+    fse.copyFileSync(
+        `${testsDirectory}/${supportJson}`,
+        `${targetFolder}/${supportJson}`
+    );
+    console.info('Import harness complete.');
```
howard-e, line 285:
```suggestion
    console.info('Harness files update complete.');
```

server/scripts/import-tests/index.js
```diff
@@ -251,6 +253,38 @@ const readDirectoryGitInfo = directoryPath => {
     return { gitSha, gitMessage, gitCommitDate };
 };
 
+const importHarness = () => {
```
howard-e, line 256: Nice!

howard-e (commented):
@evmiguel great stuff! Definitely looking forward to getting this in and reducing the overhead involved with certain `w3c/aria-at` updates! I've left a few inline thoughts but this seems very close.

This also made me notice `w3c/aria-at` is missing some changes which this would now remove so I've also submitted https://github.com/w3c/aria-at/pull/1056. This work made that disconnect super clear.

server/scripts/import-tests/index.js
```diff
@@ -251,6 +253,38 @@ const readDirectoryGitInfo = directoryPath => {
...
+        },
+    });
+
+    // Copy files
+    const commandsJson = 'commands.json';
+    const supportJson = 'support.json';
+    fse.copyFileSync(
+        `${testsDirectory}/${commandsJson}`,
+        `${targetFolder}/${commandsJson}`
+    );
```
howard-e, line 280: This file MAY exist if a much older commit is used to run the import script

evmiguel: I put in a check to see if that file exists.

.gitignore
evmiguel: I added a brief notes in the docs.

client/resources/aria-at-test-io-format.mjs
evmiguel: Should that PR be merged first?

server/resources/commandsV1.json
evmiguel: I've removed it.

client/resources/aria-at-test-io-format.mjs
howard-e: Yep, I went ahead and merged!

.gitignore
howard-e: Looks good! After untracking the `client/resources` folder, this PR should be good to go.

evmiguel: All set!

PR #1065, opened 2024-04-30
Update status dialog for minimum exact AT versions

alflennik:
Updates the TestPlanStatusDialog to support minimum and exact AtVersions, which required extensive changes to the algorithm that populates the content of that dialog.

One difference worth highlighting is that I updated the sorting of the dialog to sort by AT before sorting by "required" - the dialog feels a lot more organized when all of the reports are gathered together for one AT, especially because recommended reports can have a huge number of rows in the dialog.

One limitation of this PR is the fact that the automatic switch from minimumAtVersion to exactAtVersion which occurs when a testPlanVersion goes from candidate to recommended has not yet been implemented, mostly because this PR is already bulging at the seams and also because the PR enabling it has not yet been merged into #1055. That seems like a good follow-on task to this one.


server/resolvers/TestPlanVersion/testPlanReportStatusesResolver.js
```diff
@@ -0,0 +1,179 @@
+const atResolver = require('../TestPlanReport/atResolver');
```
howard-e, line 1: This was a lot but the tests definitely make understanding this much easier!

howard-e (commented):
@alflennik this looks great and certainly another big step towards supporting #791 and #792!

My main hang up is with the newly named `TestPlanReportStatus` type. I think it may cause some confusion so would like for an alternate but if I'm alone in my thought on it, it does not have to be considered as blocking since it can be easily refactored afterwards if needed.

howard-e (approved):
As my major bit of feedback was on the name used here, `TestPlanReportStatus`, I don't think it's anything this particular PR has to be held up on.

Definitely something that can be revised in the future if needed.

Issue #993, opened 2024-03-28
Assign Tester Menu options are incorrectly sorted


howard-e:
The Assign Tester Menu options for an admin are being incorrectly sorted. The values are being sorted alphabetically but not in a case-insensitive way. The following screenshot shows a small subset of the tester options as being:

* AlyssaGourley, IsaDC, Paul-Clue, SamuelHShaw, a11ydoer, alflennik ... and so on

The first 4 are taking precedence because they are capitalized. The expected order should be:

* a11ydoer, alflennik, AlyssaGourley, IsaDC, Paul-Clue, SamuelHShaw ... and so on

<img width="204" alt="screenshot showing small subset of tester options" src="https://github.com/w3c/aria-at-app/assets/7191577/baa71bee-558a-4a39-8169-2fcbdd020bf9">

PR #1067, opened 2024-05-01
Correct sorting for the tester assignment menu

Paul-Clue:
see issue #993 

This pull request correctly sorts the tester names in the dropdown for assigning testers and the dropdown for running a test as specific tester.

### DIFF
--- a/client/components/TestQueue/index.jsx
+++ b/client/components/TestQueue/index.jsx
@@ -48,7 +48,6 @@ const TestQueue = () => {
                 testPlanReports = [],
                 testPlans = []
             } = data;
-            // section:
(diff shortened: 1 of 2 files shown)


client/components/TestQueue/index.jsx
```diff
@@ -48,6 +48,7 @@ const TestQueue = () => {
                 testPlanReports = [],
                 testPlans = []
             } = data;
+            // section:
```
alflennik, line 51: Debugging code?

alflennik: I went ahead and removed it

Issue #1085, opened 2024-05-02
Type casting error shown for pages related to Modal Dialog v24.05.02


howard-e:
Modal Dialog v24.05.02 has been imported into the app, and the [current version page](https://aria-at.w3.org/test-review/117356) shows a potential type casting error which seem to match the presentationNumber values of the source csv files.

The issue also affects the test run page being rendered when it has been added to the Test Queue

<details>

<summary>Error shown</summary>

Int cannot represent non-integer value: 14.1 Int cannot represent non-integer value: 14.1 Int cannot represent non-integer value: 14.1 Int cannot represent non-integer value: 16.1 Int cannot represent non-integer value: 16.1 Int cannot represent non-integer value: 16.1 Int cannot represent non-integer value: 40.1 Int cannot represent non-integer value: 40.1 Int cannot represent non-integer value: 40.1

</details>

To replicate this locally:

* `git checkout main`
* `yarn db-import-tests:dev;`
* Run the app
* Visit the Modal Dialog v24.05.02 version page from the Data Management page and observe the error

PR #1086, opened 2024-05-02
fix: Change cast type for `rowNumber` from `Int!` to `Float!`

howard-e:
Address #1085

### DIFF
--- a/server/graphql-schema.js
+++ b/server/graphql-schema.js
@@ -393,8 +393,11 @@ const graphqlSchema = gql`
         """
         Since each TestPlan originates from a CSV, this number corresponds to
         the row within the CSV where this test originated.
+
+        Float type because presentationNumber fields from the source *.csv
+        files use decimal numbers in the v2 format.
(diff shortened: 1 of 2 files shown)


server/graphql-schema.js
```diff
@@ -393,8 +393,11 @@ const graphqlSchema = gql`
         """
         Since each TestPlan originates from a CSV, this number corresponds to
         the row within the CSV where this test originated.
+
+        Float type because presentationNumber fields from the source *.csv
```
howard-e, line 397: For reviewers, I cannot at this moment say with certainty why this didn't affect previous v2 test format imports that also use decimal numbers the same way. I'll continue to investigate why through the commit which caused it: https://github.com/w3c/aria-at/commit/af80e48dc9156f2749b2ecb83d93826d6bf07e6e.

alflennik (approved):
I verified the bug as well as the fix, thanks for the quick turnaround.

Issue #1083, opened 2024-05-01
Implement trend reports UI


ccanash:
Implement trend reports UI as defines in #963 mockups

ccanash:
These can be tested in sandbox here: [](https://aria-at-app-sandbox.bocoup.com/report/56291/targets/204)

Issue #963, opened 2024-03-14
Trend reports design proposal


boazsender:
This issue describes a design for trend reports to meet the following requirement from the project [readme](https://github.com/orgs/w3c/projects/67/views/1?pane=info):

> Access to reports on specific AT/browser versions: Minimal UI changes that enable publication of reports for a specific AT/browser version and visibility of reports by AT and browser version.

This design proposal consists of two tables outlined below.

## 1) Trend reports table
The trend reports will consist of a summary table of all historical runs, where each row links to a detail table. This summary table will take the place of the existing table at https://aria-at.w3.org/reports which currently displays the lates trun.
### Summary table heading (column names): 
* Test plan version (git sha of test suite)
* One column for every AT/Browser combo
* Date of run

### Summary table rows
* One row for each run
* A check mark in the cell for each AT/Browser combo if the combo was run with a link to that run

## 2) Detail table 
The detail table will be the same as the current table that exists at https://aria-at.w3.org/reports, but reproduced for every historical run and moved to a url dedicated to the git sha of the given run (e.g. https://aria-at.w3.org/reports/runs/$SHA)

mcking65:
I don't understand the base concept underpinning this design. The test plan version and run date are not relevant dimensions of the trend data. The only relevant dimension for any given AT/Browser combination is the version of the AT. In the future, we might also show a trend by browser version. It remains to be determined whether or not browser version trends have any value.

For now, I think we should keep the top-level page as is so that it always shows the latest data. That is still the most important data. We want anyone coming to the site to easily see that information.

One light-weight approach to providing access to data about all versions of an AT that have been tested for a given test plan might be to provide a AT versions summary table at the top of each detailed report page. So, users would continue to see the latest data on the top-level page and on the second-level page. But when they get to the detailed report on a test plan for a given AT/browser combination, they would see a summary of all versions of that AT tested for that plan.

[...]

PR #1092, opened 2024-05-09
feat: Include Trend Reports table on reports page

howard-e:
Address #1083 and implements the design shared in https://github.com/w3c/aria-at-app/issues/963#issuecomment-2087744440.

This PR adds a versions summary and results table on the final reports page (AT + Browser report page): `/reports/<id>/targets/<id>`


gnarf (approved):
Seems pretty simple and straight forward.  👍

Might be worth adding some simple tests for things like only showing for recommended reports, but I wouldn't necessarily block on this

Issue #934, opened 2024-02-22
Designate "bot" users in the database


jugglinmike:
Currently, the application logic uses [a heuristic based on username](https://github.com/w3c/aria-at-app/blob/a78a9786914737912ad3a4537c60f831bc12ae79/client/utils/automation.js#L8) to differentiate between so-called "bot" users (i.e. those not associated with a human being or GitHub account) and human users. In the interest of improving maintainability, we would prefer to reflect that information in the data model itself.

PR #1093, opened 2024-05-09
Fix: Add isBot flag to users to simplify matching - Ack #934

gnarf:
Adds an `isBot` field to `User` table instead of relying on string matching.

### DIFF
--- a/client/components/ManageBotRunDialog/index.jsx
+++ b/client/components/ManageBotRunDialog/index.jsx
@@ -10,7 +10,6 @@ import {
 } from './queries';
 import DeleteButton from '../common/DeleteButton';
 import BotRunTestStatusList from '../BotRunTestStatusList';
-import { isBot } from '../../utils/automation';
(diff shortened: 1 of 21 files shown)


gnarf:
@alflennik I'll leave it up to you if this needs a second reviewer to flag someone?

howard-e (commented):
Looks good to me!

A small error being silently thrown related to the mocks though [shown here](https://github.com/w3c/aria-at-app/actions/runs/9020381882/job/24785430991#step:15:198). There's a `boo-far` tester at [the bottom of `TestQueuePageTesterPopulatedMock`](https://github.com/w3c/aria-at-app/blob/c786a3ef5f360c77169fce32e75a0cc27a01d9f1/client/tests/__mocks__/GraphQLMocks/TestQueuePageTesterPopulatedMock.js#L310) missing the `isBot` attribute.

Should be good to go after

PR #1094, opened 2024-05-09
Fix: status for RUNNING tests updates when job status becomes finalized

gnarf:
If a test is still marked as RUNNING when the job becomes COMPLETED, ERROR, or CANCELLED - the RUNNING test should be updated to be `ERROR` or `CANCELLED` (`COMPLETED` should become `CANCELLED` because the test results were not received)

---
  - https://app.asana.com/0/0/1207279995378747

### DIFF
--- a/server/tests/integration/automation-scheduler.test.js
+++ b/server/tests/integration/automation-scheduler.test.js
@@ -667,6 +667,274 @@ describe('Automation controller', () => {
         });
     });
 
+    it('should convert RUNNING test status to ERROR when job state becomes ERROR', async () => {
+        await apiServer.sessionAgentDbCleaner(async transaction => {
+            const { scheduleCollectionJob: job } =
+                await scheduleCollectionJobByMutation({ transaction });
+            const collectionJob = await getCollectionJobById({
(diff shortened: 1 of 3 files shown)


server/tests/integration/automation-scheduler.test.js
```diff
@@ -667,6 +667,274 @@ describe('Automation controller', () => {
...
+                let expectedStatus = COLLECTION_JOB_STATUS.QUEUED;
+                if (testStatus.test.id === selectedTest.id) {
+                    foundStatus = true;
+                    expectedStatus = COLLECTION_JOB_STATUS.RUNNING;
+                }
+                expect(testStatus.status).toEqual(expectedStatus);
+            }
+            expect(foundStatus).toEqual(true);
+
+            // Leave our test RUNNING but ERROR the job
```
howard-e, line 901:
```suggestion
            // Leave our test RUNNING but COMPLETED the job
```

?

howard-e (approved):
Looks good! Reviewing the tests made going through this easier as well.

Left an inline comment for clarifying a comment.

Issue #991, opened 2024-03-28
Update the Assign Tester Menu to support type-ahead


howard-e:
During a [recent CG meeting](https://www.w3.org/2024/03/27-aria-at-minutes.html#t04), it came up that with more testers now being added into the system, it is getting increasingly difficult for an admin to find the tester they'd like to assign from the list of options.

A suggestion is to make the menu support type-ahead to make it easier for an admin to assign a tester, given that they already know the username of the tester to be assigned. ie. pressing M to jump to testers beginning with M.

jscholes:
@howard-e I don't think @mcking65 was asking for a combobox with autocomplete/search.  Rather, just that the menu support type-ahead, i.e. you can press M to jump to testers beginning with M, as per the [APG example](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions-active-descendant/).  Multi-select comboboxes are hard to implement and often hard to use.

howard-e:
@jscholes thanks for clarifying! I can update the top comment to reflect

PR #1095, opened 2024-05-14
Add type-ahead to assign tester dropdown

Paul-Clue:
see issue #991 

This pull request updates the assign tester menu to support type-ahead.

### DIFF
--- a/client/components/TestQueue/AssignTesterDropdown/index.jsx
+++ b/client/components/TestQueue/AssignTesterDropdown/index.jsx
@@ -123,7 +123,7 @@ const AssignTesterDropdown = ({
                     return menuItem.innerText
                         .trim()
                         .toLowerCase()
-                        .startsWith(key);
+                        .startsWith(key.toLowerCase());
                 }
             );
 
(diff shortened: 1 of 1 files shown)


client/components/TestQueue/AssignTesterDropdown/index.jsx
```diff
@@ -113,9 +113,33 @@ const AssignTesterDropdown = ({
     const clearAriaLiveRegion = () => {
         setAlertMessage('');
     };
+
+    const handleKeyDown = event => {
```
howard-e, line 117: Interesting!

client/components/TestQueue/AssignTesterDropdown/index.jsx
```diff
@@ -113,9 +113,33 @@ const AssignTesterDropdown = ({
...
+        const { key } = event;
+        if (key.match(/[a-z]/)) {
```
howard-e, line 119:
```suggestion
        if (key.match(/[0-9a-zA-Z]/)) {
```

While there are none currently tracked by the app, I've come to find out GitHub usernames can begin with a number, so just to be safe. Uppercase is there as well in case one is holding `Shift` or `CapsLock` is enabled. It seems to also align with browsers' native behavior for `<select>`

howard-e (commented):
Well done! I've left an inline comment on the regex requiring a change, should be good to go after that!

howard-e (approved):
Looks good!

PR #1096, opened 2024-05-15
Update design for recommended Test Plans in the support tables

Paul-Clue:
see issue [#3018](https://github.com/w3c/aria-practices/issues/3018)

This pull request adds support for tests with a "Recommended" phase in the APG support tables.

### DIFF
--- a/server/handlebars/embed/public/style.css
+++ b/server/handlebars/embed/public/style.css
@@ -1,37 +1,39 @@
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@0,100;0,300;0,600;0,800;1,700&family=Noto+Sans:ital,wght@0,100..900;1,100..900&family=Roboto:wght@100;400&display=swap');
+
 body {
     margin: 0;
     /* Parent should use the posted message and set an explicit height */
     overflow: hidden;
-}
-
-#main {
+  }
+  
+  #main {
     width: 100%;
-}
-
-a {
+  }
+  
+  a {
     color: #036;
-}
-a:focus-visible {
+  }
+  a:focus-visible {
     outline: 2px solid #3a86d1;
-}
-
-h3#report-title {
+  }
+  
+  h3#report-title {
     margin-top: 0;
     margin-bottom: 0.5rem;
-
+  
     font-family: Arial, Helvetica, sans-serif;
     font-size: 18px;
-}
-
-#copied-message {
+  }
+  
+  #copied-message {
     align-self: center;
-}
-
-#at-version {
+  }
+  
+  #at-version {
     font-weight: lighter;
-}
-
-#no-data-content-container {
+  }
+  
+  #no-data-content-container {
     border-left: 1.5px solid #c0c0c0;
     border-right: 1.5px solid #c0c0c0;
     border-bottom: 1.5px solid #c0c0c0;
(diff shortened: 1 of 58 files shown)


server/handlebars/embed/public/style.css
```diff
@@ -43,54 +27,108 @@ h3#report-title {
...
+    padding-left: var(--left-right-padding);
+    padding-right: var(--left-right-padding);
+}
+
+.details > summary > span > h4 {
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 1em;
+    color: #60470c;
+    display: inline-block;
+}
```
howard-e, line 116: It is a bit lost on me why an extra `<span>` is needed looking at the structure. But I also acknowledge that I'm looking at this after several revisions so perhaps this has already been explained.

Seems to me like the following should give the same result:

```css
.details > summary > h4 {
    position: relative;
    padding-left: var(--left-right-padding);
    padding-right: var(--left-right-padding);

    font-family: Arial, Helvetica, sans-serif;
    font-size: 1em;
    color: #60470c;
    display: inline-block;
}
```

server/handlebars/embed/views/main.hbs
```diff
@@ -8,9 +8,9 @@
     {{/if}}
     {{#unless (dataEmpty testPlanReports)}}
         {{#if (isCandidate phase)}}
-            <details id='embed-report-phase-container'>
-                <summary id='candidate-title'><span>Warning! Unapproved Report</span></summary>
-                <div id='candidate-content-container'>
+            <details class="details">
```
howard-e, line 11: While not wrong to have the class name be the same as the element name, and there may be cases why you'd want to do that, I'm not sure why the need here. To me, it seems better to directly target the `details` element rather than define a `.details` class, but this is purely opinion.

server/handlebars/embed/views/main.hbs
```diff
@@ -8,9 +8,9 @@
...
+                <summary><span><h4>Warning! Unapproved Report</h4></span></summary>
```
howard-e, line 12:
```suggestion
                <summary><h4>Warning! Unapproved Report</h4></summary>
```

Based on my above comment on not needing to target `<span>`. Also typically, inline elements shouldn't contain block elements and I can't say I come across much instances of `<h*>` inside a `<span>` (but I'm sure there are exceptions!)

server/handlebars/embed/views/main.hbs
```diff
@@ -23,20 +23,17 @@
...
-                    assistive technology developers and represent consensus
-                    regarding:
-                    <ol>
-                        <li>applicability and validity of the tests, and</li>
-                        <li>accuracy of test results.</li>
-                    </ol>
-                </div>
-            </details>
+            <details class="details">
+            <summary class="recommended"><span><h4>Recommended Report</h4></span></summary>
```
howard-e, line 27:
```suggestion
            <summary class="recommended"><h4>Recommended Report</h4></summary>
```

Same thought as before

howard-e (changes requested):
Looks good! Thanks for taking another look at the design supporting the RECOMMENDED copy and color scheme as well!

I've left some additional inline thoughts below.

But the key "issue" is a color change has been missed, when the details panel is open. I tested it on the WAI website to see how it looks and noticed the borders in the expanded view are don't match the green accent. You can also see it on the `/embed` link by zooming in your browser to the maximum %.

Screenshot 2024-05-16 at 9 27 35 AM

server/handlebars/embed/views/main.hbs
```diff
@@ -23,20 +23,17 @@
...
-                    tests. Recommended ARIA-AT tests have been reviewed by
-                    assistive technology developers and represent consensus
-                    regarding:
-                    <ol>
-                        <li>applicability and validity of the tests, and</li>
-                        <li>accuracy of test results.</li>
-                    </ol>
-                </div>
-            </details>
+            <details >
```
howard-e, line 26:
```suggestion
            <details>
```

howard-e:
@Paul-Clue thanks for addressing so much here. Last issue I just came across is that the default `<summary>` arrow is still showing when on Safari. Please see the following screenshot:

screenshot show default summary arrow

You could fix this by adding the following to the css:

```css
details summary::-webkit-details-marker {
    display: none;
}
```

Should be good to merge after!

howard-e (approved):
Looks good, thanks for addressing the feedback!

PR #1098, opened 2024-05-16
Add AT version column to Test Queue

alflennik:


server/resources/ats.json
```diff
@@ -6,9 +6,8 @@
         "defaultConfigurationInstructionsHTML": "Configure JAWS with default settings. For help, read &lt;a href=&quot;https://github.com/w3c/aria-at/wiki/Configuring-Screen-Readers-for-Testing&quot;&gt;Configuring Screen Readers for Testing&lt;/a&gt;.",
```
howard-e, line 6: Looks like you're accidentally removing newly added tokens coming from w3c/aria-at. You may want to remove these changes.

client/components/TestQueueRow/index.jsx
```diff
@@ -530,6 +530,11 @@ const TestQueueRow = ({
         <LoadingStatus message={loadingMessage}>
             <tr className="test-queue-run-row">
                 <th>{renderAssignedUserToTestPlan()}</th>
+                <td style={{ padding: '0.75rem' }}>
```
howard-e, line 533: We don't do these inline styles but considering this is only temporary patch, it's fine!

client/components/TestQueue/index.jsx
```diff
@@ -102,6 +102,7 @@ const TestQueue = () => {
                     <thead>
                         <tr>
                             <th className="test-plan">Test Plan</th>
+                            <th className="test-plan">AT Version</th>
```
howard-e, line 105:
```suggestion
                            <th className="test-plan">AT Version Requirement</th>
```

That language may be preferred.

howard-e (changes requested):
@alflennik thanks for getting this done so quickly! The column looks good and left a few inline comments.

The main blocker here is that when assigning and unassigning myself or others to any report, it crashes the page. Could you take a look at why?

howard-e (approved):

Issue #957, opened 2024-03-12
Report errors from subprocesses


jugglinmike:
While investigating an issue with this project's "import" script, @howard-e found that a process spawned by the script had crashed. Although this exceptional circumstance should be plainly evident from server logs, Howard could only intuit its occurrence indirectly (that is: by interpreting an error which occurred after the script incorrectly assumed the subprocess had exited cleanly).

Whenever a subprocess is created, the parent should account for the possibility that the child does not complete successfully (even if it is only to report the failure and exit).

A cursory review of the source identifies a number of instances of hazardous subprocess management:

    $ git grep spawn '**/*.js'
    client/tests/util/getPage.js:const spawn = require('cross-spawn');
    client/tests/util/getPage.js:        const server = spawn('yarn', ['workspace', serverOrClient, 'dev'], {
    server/scripts/import-tests/index.js:const spawn = require('cross-spawn');
    server/scripts/import-tests/index.js:    return spawn
    server/scripts/import-tests/index.js:    const installOutput = spawn.sync('npm', ['install'], {
    server/scripts/import-tests/index.js:    const buildOutput = spawn.sync('npm', ['run', 'build'], {
    server/scripts/import-tests/index.js:    spawn.sync('git', ['clone', ariaAtRepo, gitCloneDirectory]);

...so addressing this problem should begin with a formal audit of the project.

PR #1102, opened 2024-05-20
Report errors in subprocesses

evmiguel:
This PR addresses #957, which raised the issue of a lack of error reporting in subprocesses.

### DIFF
--- a/client/tests/util/getPage.js
+++ b/client/tests/util/getPage.js
@@ -40,6 +40,7 @@ const startServer = async serverOrClient => {
 
         server.on('error', error => {
             console.info(`Error found in startServer process: ${error}`); // eslint-disable-line no-console
+            process.exit(1);
         });
 
         server.on('exit', (code, signal) => {
--- a/client/tests/util/getPage.js
+++ b/client/tests/util/getPage.js
@@ -38,6 +38,20 @@ const startServer = async serverOrClient => {
             }
         );
 
+        server.on('error', error => {
+            console.info(`Error found in startServer process: ${error}`); // eslint-disable-line no-console
+        });
+
+        server.on('exit', (code, signal) => {
+            if (code) {
+                console.error('startServer exited with code', code);
+            } else if (signal) {
+                console.error('startServer was killed with signal', signal);
+            } else {
+                console.info('startServer exited with no errors.'); // eslint-disable-line no-console
+            }
+        });
+
         const killServer = async () => {
             await new Promise((resolve, reject) => {
                 treeKill(server.pid, error => {
--- a/client/tests/util/getPage.js
+++ b/client/tests/util/getPage.js
@@ -39,8 +39,9 @@ const startServer = async serverOrClient => {
         );
 
         server.on('error', error => {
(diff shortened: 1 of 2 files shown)


client/tests/util/getPage.js
```diff
@@ -38,6 +38,20 @@ const startServer = async serverOrClient => {
             }
         );
 
+        server.on('error', error => {
+            console.info(`Error found in startServer process: ${error}`); // eslint-disable-line no-console
```
jugglinmike, line 42: It seems as though the behavior of the tests will be meaningless from this moment forward, and the test runner will report that misleading error much more prominently than a stray console message. Shouldn't we interrupt test execution, instead?

server/scripts/import-tests/index.js
```diff
@@ -53,10 +53,18 @@ const builtTestsDirectory = path.resolve(gitCloneDirectory, 'build', 'tests');
...
 const gitRun = (args, cwd = gitCloneDirectory) => {
-    return spawn
-        .sync('git', args.split(' '), { cwd })
-        .stdout.toString()
-        .trimEnd();
+    const gitRunOutput = spawn.sync('git', args.split(' '), { cwd });
+
+    if (gitRunOutput.error) {
+        console.info(
+            `'git ${args.split(' ')} ${cwd}' failed with error ${
```
jugglinmike, line 60: Splitting on the empty space is necessary for `spawn.sync`. Here, though, by coercing the resultant array back into a string, we'll end up with a misrepresentation of the command executed (e.g. `git fetch origin main` becomes `git fetch,origin,main`).

jugglinmike (changes requested):
Thanks for taking this up, @evmiguel!

A few of the new error messages present the current-working directory as though it were part of the command. It's good to provide the information as context, but we should probably describe it as distinct from the command which was executed.

client/tests/util/getPage.js
evmiguel: I agree. I've updated the code to exit on error.

server/scripts/import-tests/index.js
evmiguel: Thanks for catching this! I've updated the code to remove the split.

evmiguel:
@jugglinmike Thank you for taking a look! I've removed the directories from the console messages and addressed your feedback.

jugglinmike (approved):
Introducing `process.exit` to communicate an error gave me pause because that appeared to stop the test runner itself. Doing so wouldn't be the end of the world because we'd still be surfacing the problem along with some useful information (thanks to the use of `console.info`), but we'd also be interfering with the harness's other work (e.g. running other tests, performing cleanup logic, etc.).

I found some time to experiment today, and I learned that Jest appears to overwrite the built-in `process.exit` so that it can gracefully handle uses like this. However, I'd wager that it was implemented to handle invocations from the application under test. Using the overridden `process.exit` in the test environment may technically work, but it seems a bit misleading. I've rewritten the setup logic to throw an exception which will not be caught due to its origin in an event handler. Jest also accounts for this scenario, so the result is the same (the harness reports the error and continues with the rest of its work), but this way, the code doesn't insinuate that the process will immediately exit.

Issue #1107, opened 2024-05-22
Add test plan report with decimal test.rowNumber to test seed data


gnarf:
Need this to be able to write a test (#1106)

Issue #1106, opened 2024-05-22
Write tests for decimal test rowNumbers in automation-scheduler.test.js


gnarf:
When writing test for #1105 I was blocked by needing some data, opening an issue to remind me to do this after the data I need is in the test database.

PR #1111, opened 2024-05-23
Include the latest w3c/aria-at data imports in tests and update sample data

howard-e:
Address #1107 by doing the following:
1. Allowing the most recent commits from w3c/aria-at to be imported when importing data from w3c/aria-at again.
2. This also updates the sample data population script to include a modal-dialog example to test [tests which has decimal value `presentationNumber`s](https://github.com/w3c/aria-at/blob/master/tests/modal-dialog/data/tests.csv).

Also took the opportunity in this PR to make the `yarn db-import-tests:*` process multiple commits at once, with `-c "commit_1 commit_2"`. This saves ~30s for me.


howard-e:
@gnarf `TestPlanVersion 160` and `TestPlanReport 19` should have the data you intend to test for #1106.

gnarf (approved):
Looks good to me!

Issue #820, opened 2023-10-23
Update UI in real time to receive the "per-test” status updates: "started" and "error".


ccanash:

jugglinmike:
### Background

As of this writing, the ARIA-AT App user interface currently does not change when the system begins any given response-collection task. Each task is considered "queued" immediately after being created, and it remains in that state until the response-collection system reports the AT response, at which time, the task's status changes to "completed." The goal of this enhancement is to surface information about two other states: the period where the response-collection system is actively working on a given test (i.e. "in progress"), and the period where an exceptional circumstance has forced the system to abort.

Unfortunately, the UI is currently inconsistent in how it describes the "queued" state. On the Test Queue page, it is described as "Queued", while in the Test Plan Run page, it is described as "In Progress by Bot." Although this problem did not motivate the enhancement described by this issue, the solution should address it.

### User interface design requirements

1. Communicate the following states for response-collection tasks (in addition to "queued" and "completed"):
   - "In progress"
   - Halted due to an "error"
2. Update the existing UI to accurately describe the "queued"

### Process for determining which aspects of the UI should be updated

1. Evaluate the existing UI which communicates the status of response-collection tasks
  - Test Queue page: the "Testers" column (e.g. "Responses for 3 of 5 tests recorded")
  - Test Queue page: the "Report Status" column (e.g. "11 Tests Queued")
  - Test Plan Run page: the label for each testing task (e.g. "In Progress by Bot")
2. Identify any additional areas of the UI which should surface this information

isaacdurazo:
Here's an updated mockup of the Test Queue with the new states added.

I am also including a screenshot of the previously proposed states for the Test Run page. It might be worth revisiting these for implementation. When I designed these I think I remembered an issue with using custom icons and having to go with the existing icon library being used on ARIA AT.

![ARIA-AT - Test Queue – 1](https://github.com/w3c/aria-at-app/assets/1379244/a4960d0c-997c-4717-a99f-c9498faaff32)

PR #1125, opened 2024-06-06
feat: Test Run / Navigator polling updates for collection jobs

gnarf:
---
  - https://app.asana.com/0/0/1207383005191146

### DIFF
--- a/client/components/TestRun/CollectionJobContext.js
+++ b/client/components/TestRun/CollectionJobContext.js
@@ -1,0 +1,64 @@
+import PropTypes from 'prop-types';
+import React, { createContext, useState, useEffect } from 'react';
+import { useLazyQuery } from '@apollo/client';
+import { COLLECTION_JOB_UPDATES_QUERY } from './queries';
+import { isJobStatusFinal } from '../../utils/collectionJobStatus';
+const pollInterval = 5000;
+
+export const Context = createContext({
+    state: {
+        collectionJob: null
+    },
+    actions: {}
+});
+
+export const Provider = ({
+    children,
+    testPlanRun: { id: testPlanRunId, collectionJob: initialCollectionJob }
+}) => {
+    const [providerValue, setProviderValue] = useState({
+        state: { collectionJob: initialCollectionJob },
+        actions: {}
+    });
+
+    const [, { data: collectionJobUpdateData, startPolling, stopPolling }] =
+        testPlanRunId
+            ? useLazyQuery(COLLECTION_JOB_UPDATES_QUERY, {
+                  fetchPolicy: 'cache-and-network',
+                  variables: { collectionJobId: initialCollectionJob?.id },
+                  pollInterval
+              })
+            : {};
+
+    // control the data flow, turn on polling if this is a collection job report
+    // that still has possible updates.
+    useEffect(() => {
+        // use the colllection job from the polling update first priority
+        // otherwise, default to the first data fetch from the API
+        const collectionJob =
+            collectionJobUpdateData?.collectionJob ?? initialCollectionJob;
+        const status = collectionJob?.status;
+        if (collectionJob && !isJobStatusFinal(status)) {
+            startPolling(pollInterval);
+        } else {
+            stopPolling();
+        }
+        setProviderValue({ state: { collectionJob }, actions: {} });
+    }, [collectionJobUpdateData]);
+
+    return (
+        <Context.Provider value={providerValue}>{children}</Context.Provider>
+    );
+};
+
+Provider.propTypes = {
+    children: PropTypes.node,
+    testPlanRun: PropTypes.shape({
+        id: PropTypes.string,
+        collectionJob: PropTypes.shape({
+            id: PropTypes.string.isRequired,
+            status: PropTypes.string.isRequired,
+            testStatus: PropTypes.arrayOf(PropTypes.object).isRequired
+        })
+    }).isRequired
+};
--- a/client/components/TestRun/CollectionJobContext.js
+++ b/client/components/TestRun/CollectionJobContext.js
@@ -12,10 +12,19 @@ export const Context = createContext({
     actions: {}
 });
 
-export const Provider = ({
-    children,
-    testPlanRun: { id: testPlanRunId, collectionJob: initialCollectionJob }
-}) => {
+export const Provider = ({ children, testPlanRun }) => {
+    if (!testPlanRun) {
+        // Anonymous page / not working, just viewing the tests, no need for the
+        // provider to provide any data or updates, but to be consistent we will
+        // still wrap with a provider with static data
+        return (
+            <Context.Provider value={{ state: {}, actions: {} }}>
+                {children}
+            </Context.Provider>
+        );
+    }
+    const { id: testPlanRunId, collectionJob: initialCollectionJob } =
+        testPlanRun;
     const [providerValue, setProviderValue] = useState({
         state: { collectionJob: initialCollectionJob },
         actions: {}
@@ -60,5 +69,5 @@ Provider.propTypes = {
             status: PropTypes.string.isRequired,
             testStatus: PropTypes.arrayOf(PropTypes.object).isRequired
         })
-    }).isRequired
+    })
 };
--- a/client/components/TestRun/TestRun.css
+++ b/client/components/TestRun/TestRun.css
@@ -176,8 +176,8 @@ button.test-navigator-toggle:focus {
     color: white;
     font-size: 10px;
     position: relative;
-    top: -3px;
-    left: 4px;
+    top: -4px;
+    left: 3px;
 }
 .test-name-wrapper.bot-queued .progress-indicator {
     background: #295fa6;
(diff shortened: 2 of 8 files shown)


client/components/TestRun/TestRun.css
```diff
@@ -158,6 +158,27 @@ button.test-navigator-toggle:focus {
     left: 4px;
 }
 
+.test-name-wrapper.bot-running .progress-indicator {
+    background: #d2d5d9;
```
stalgiag, line 162: not sure how strict this requirement is:

"in progress (blue with green circle around)"

client/components/TestRun/index.jsx
```diff
@@ -68,21 +68,26 @@ const TestRun = () => {
...
     const { runId: testPlanRunId, testPlanReportId } = params;
 
-    const { loading, data, error } = useQuery(
+    const { loading, data, error, startPolling, stopPolling } = useQuery(
         testPlanRunId ? TEST_RUN_PAGE_QUERY : TEST_RUN_PAGE_ANON_QUERY,
         {
             fetchPolicy: 'cache-and-network',
-            variables: { testPlanRunId, testPlanReportId }
+            variables: { testPlanRunId, testPlanReportId },
+            pollInterval
```
stalgiag, line 76: I understand why this is a good idea but I wonder if the polling could be more targeted.  It feels extensive and triggers a re-render of most of the page. Fine with it for now though!

client/components/TestRun/index.jsx
```diff
@@ -68,21 +68,26 @@ const TestRun = () => {
 
     const { runId: testPlanRunId, testPlanReportId } = params;
 
-    const { loading, data, error } = useQuery(
+    const { loading, data, error, startPolling, stopPolling } = useQuery(
```
stalgiag, line 71: I wonder if we want to announce new test results coming in via screenreader. I know that isn't in the spec but VoiceOver is very quiet during these updates.

stalgiag (approved):
Looks good to me! Few comments inline but no blockers. Thanks!

client/components/TestRun/TestRun.css
gnarf: I went with the colors in the image from issac linked on #820 (assuming the one from the designer was correct)

client/components/TestRun/index.jsx
gnarf: Yeah - I was thinking about this and it would be a bit nicer to only poll the "status"es - I might give this a revision for this purpose

client/components/TestRun/index.jsx
gnarf: I wonder if setting up the overall status section (upper right) as a live region when the job is running might be nice - but I don't necessarily like to assume I know what the screen reader user would want in this situation.  Happy to bring this up and maybe we push the screen reader update to it in the next issue?

client/components/TestRun/index.jsx
gnarf: I should be around on monday, but if not @stalgiag - I think we could merge this as is (unless others think we should wait) and I could do this part on a second pass

client/components/TestRun/TestRun.css
stalgiag: Ah gotcha! Didn't see those

client/components/TestRun/index.jsx
stalgiag: Sounds good!

gnarf:
Decided at the meeting today to continue on with splitting up this page such that only the updating parts care about updates so it does less "whole page refresh" feeling tasks

howard-e (changes requested):
@gnarf changes look good to me in the best case scenarios (admin, tester, vendors, etc). But I have a consistent crash when viewing anything in the test queue as an anonymous user. That seems to be because of a **valid** assumption here that a `testPlanRun` variable would always be populated. But the anon user [operates on `testPlanReport](https://github.com/w3c/aria-at-app/blob/756672fddded26f18c6acf40c19731d667a5169a/client/components/TestRun/queries.js#L217).

I've long since thought this file should be severely refactored (with myself being the culprit from some time back) and there's certainly room for breaking more things out here that shouldn't be included

client/components/TestRun/index.jsx
```diff
@@ -1013,8 +954,7 @@ const TestRun = () => {
             ];
         }
 
-        const externalLogsUrl =
-            collectionJobQuery?.collectionJobByTestPlanRunId?.externalLogsUrl;
+        const externalLogsUrl = testPlanRun.collectionJob?.externalLogsUrl;
```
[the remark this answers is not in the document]
gnarf: Fixed!

client/components/TestRun/index.jsx
```diff
@@ -1319,94 +1214,98 @@ const TestRun = () => {
...
-                            item => item.name
-                        )}
-                        patternName={testPlanVersion.title}
-                        testerName={tester.username}
-                        handleAction={handleAtAndBrowserDetailsModalAction}
-                        handleClose={handleAtAndBrowserDetailsModalCloseAction}
-                    />
-                )}
-            </Container>
+            <CollectionJobContextProvider testPlanRun={testPlanRun}>
```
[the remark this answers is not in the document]
gnarf: also fixed!

howard-e (commented):
The fix and tests look good to me! (although there is a merge conflict to resolve here now since #1123 just got merged). Should be good to go afterwards

howard-e (approved):

client/components/TestRun/CollectionJobContext.js
```diff
@@ -0,0 +1,73 @@
...
+        // use the colllection job from the polling update first priority
+        // otherwise, default to the first data fetch from the API
+        const collectionJob =
+            collectionJobUpdateData?.collectionJob ?? initialCollectionJob;
+        const status = collectionJob?.status;
+        if (collectionJob && !isJobStatusFinal(status)) {
+            startPolling(pollInterval);
+        } else {
+            stopPolling();
+        }
```
stalgiag, line 54:
```suggestion
        }
        return () => {
          stopPolling();
        }
```
Returning a cleanup function in the event of a component unmount could be good

client/components/TestRun/CollectionJobContext.js
```diff
@@ -0,0 +1,73 @@
...
+const pollInterval = 5000;
+
+export const Context = createContext({
+    state: {
+        collectionJob: null
+    },
+    actions: {}
+});
+
+export const Provider = ({ children, testPlanRun }) => {
```
stalgiag, line 15: I like this way of sharing the polling across multiple components!

client/components/TestRun/CollectionJobContext.js
```diff
@@ -0,0 +1,73 @@
...
+        // otherwise, default to the first data fetch from the API
+        const collectionJob =
+            collectionJobUpdateData?.collectionJob ?? initialCollectionJob;
+        const status = collectionJob?.status;
+        if (collectionJob && !isJobStatusFinal(status)) {
+            startPolling(pollInterval);
+        } else {
+            stopPolling();
+        }
+        setProviderValue({ state: { collectionJob }, actions: {} });
```
stalgiag, line 55:
```suggestion
        setProviderValue({ state: { collectionJob }, actions: {} });
        
        return () => {
          stopPolling();
        }
```
Could be good to return a cleanup function in the event of a component unmount

stalgiag (approved):
Tested this and the rendering is so much smoother now. Major improvement! At some point it might be cool to try and abstract that `Context` even further to make a general polling provider. I can imagine that could be useful elsewhere in the future. Thanks for going the extra mile on this one!

client/components/TestRun/CollectionJobContext.js
gnarf: Not sure what this means exactly.  Would this to be to enable the page to stop polling when the components on the page don't need it anymore?  Seems like this case wouldn't be necessary, a minor save for what is unlikely to be a permanent situation (most up to date CollectionJob info for the current TestPlanRun is what this provider is providing) The `useContext` mechanism should automatically unmount/disconnect from the updates when it does.

gnarf:
> At some point it might be cool to try and abstract that `Context` even further to make a general polling provider

I feel like if we end up needing it again, I could pretty easily do a `<PollingContext.Provider query={gql}>` - but given how much the abstraction layers are hurting us right now, I think waiting until we need to write it twice, we should keep this one purpose built.

client/tests/smokeTest.test.js
```diff
@@ -111,7 +111,33 @@ describe('smoke test', () => {
                 const h1Handle = await page.waitForSelector('h1');
                 const h1Text = await h1Handle.evaluate(h1 => h1.innerText);
                 expect(h1Text).toBe('Data Management');
-            })
+            }),
+            getPage(
```
howard-e, line 115: Thanks for updating the tests to account for this

howard-e (approved):

client/components/TestRun/CollectionJobContext.js
stalgiag: This is just a [cleanup function](https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/) to ensure that the polling is stopped even if the context is unmounted at an unexpected time. My understanding is that this is best practice for any `useEffect` with an async side effect.

PR #1128, opened 2024-06-10
fix: use correct test number sequence for raising github issues

stalgiag:
This PR modifies `createIssueLink()` so that it accepts a new property in the options param, `testSequenceNumber`. In all instances where this function is used, the generated number used for rendering the test number ("Test X:") is passed. This handles confusion about which test is referenced when creating issues. The test row number which is now decoupled from the specific report's sequence is still stored in the `hiddenMetadata` of the issue.

I wanted to add a unit test but this would be difficult with the current design of this method since the testable logic for this particular problem has more to do with the generated sequence passed in. I can think of ways to adjust the design but none is an absolute win so I decided this was out of scope. Let me know if reviewers think otherwise.

addresses #1110 

---
  - https://app.asana.com/0/0/1207383005191128

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -367,7 +367,7 @@ const TestRun = () => {
                 'YY.MM.DD'
             )}`,
             testTitle: currentTest.title,
-            testRowNumber: currentTest.rowNumber,
+            testSequenceNumber: currentTest.seq,
             testRenderedUrl: currentTest.renderedUrl,
             atName: testPlanReport.at.name,
             browserName: testPlanReport.browser.name,
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -367,6 +367,7 @@ const TestRun = () => {
                 'YY.MM.DD'
             )}`,
             testTitle: currentTest.title,
+            testRowNumber: currentTest.rowNumber,
             testSequenceNumber: currentTest.seq,
             testRenderedUrl: currentTest.renderedUrl,
             atName: testPlanReport.at.name,
--- a/client/utils/createIssueLink.js
+++ b/client/utils/createIssueLink.js
@@ -9,6 +9,28 @@ const atLabelMap = {
     NVDA: 'nvda'
 };
 
+/**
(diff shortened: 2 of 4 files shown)


client/utils/createIssueLink.js
gnarf, line 1: Might be worth adding some propTypes for verification of params to this.

Also noticed at the bottom there is a "query" still using the old testRowNumber

gnarf (changes requested):
very simple change needed IMO (the query should be updated to use the correct sequence number instead of row number)

client/utils/createIssueLink.js
stalgiag: Good call! Didn't even notice the query down there.

Addressed in [f04ad78](https://github.com/w3c/aria-at-app/pull/1128/commits/f04ad78a5f5a0b826ae97f2d77ad744fbf8e9b7a) and [ddb515b](https://github.com/w3c/aria-at-app/pull/1128/commits/ddb515b9e38f9f6a663fc3139be54bfbb6ace6b6) respectively.

gnarf (approved):

client/components/TestRun/index.jsx
```diff
@@ -367,7 +367,7 @@ const TestRun = () => {
                 'YY.MM.DD'
             )}`,
             testTitle: currentTest.title,
-            testRowNumber: currentTest.rowNumber,
+            testSequenceNumber: currentTest.seq,
```
howard-e, line 370:
```suggestion
            testRowNumber: currentTest.rowNumber,
            testSequenceNumber: currentTest.seq,
```

Looks like `testRowNumber` should be re-added here to make sure the test number continues to be included when created from `<TestRun/>` as well, and to also preserve the `issuesResolver` search.

howard-e (changes requested):
@stalgiag left an inline change request. All the other changes look good to me

client/components/TestRun/index.jsx
stalgiag: Committed. Thank you!

howard-e (approved):
Looks good! Thanks for addressing the feedback

PR #1131, opened 2024-06-17
feat: Support minimum and exact AT Version requirements in Test Run's version selection modal

howard-e:
This is related to discussions shared in #791. This PR makes the following changes to the `<AtAndBrowserDetailsModal>` component:
1. If a minimum AT Version is set for a `TestPlanReport`, it prevents AT Versions released before the minimum specified AT Version to be selected on the Test Run page.
2. If an exact AT Version is set for a `TestPlanReport`, it Includes a disclaimer message that it's expected that the tester is using the specified AT Version and prevents any AT Version selection. The disclaimer message is:

> Results collected for this test plan require {atName} {atVersion}. By continuing, you confirm that your results are being recorded using the specified version of the Assistive Technology.

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -1390,9 +1390,24 @@ const TestRun = () => {
                         isAdmin={isAdminReviewer}
                         atName={testPlanReport.at.name}
                         atVersion={currentTest.testResult?.atVersion?.name}
-                        atVersions={testPlanReport.at.atVersions.map(
-                            item => item.name
-                        )}
+                        atVersions={testPlanReport.at.atVersions
+                            .filter(item => {
+                                // Only provide at version options that released
+                                // at the same time or later than the minimum
+                                // AT version
+                                let earliestReleasedAt = null;
+                                if (testPlanReport.minimumAtVersion) {
+                                    earliestReleasedAt = new Date(
+                                        testPlanReport.minimumAtVersion.releasedAt
+                                    );
+                                    return (
+                                        new Date(item.releasedAt) >=
+                                        earliestReleasedAt
+                                    );
+                                }
+                                return item;
+                            })
+                            .map(item => item.name)}
                         browserName={testPlanReport.browser.name}
                         browserVersion={
                             currentTest.testResult?.browserVersion?.name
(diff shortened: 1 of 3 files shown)


client/components/common/AtAndBrowserDetailsModal/index.jsx
```diff
@@ -326,42 +349,59 @@ const AtAndBrowserDetailsModal = ({
...
+                                            <Form.Label>
+                                                Assistive Technology Version
+                                            </Form.Label>
+                                            <Form.Control
+                                                disabled
+                                                type="text"
+                                                value={exactAtVersion.name}
+                                            />
+                                        </>
+                                    ) : (
```
stalgiag, line 364: Good UI choice!

client/components/TestRun/index.jsx
```diff
@@ -1390,9 +1390,24 @@ const TestRun = () => {
...
+                                // at the same time or later than the minimum
+                                // AT version
+                                let earliestReleasedAt = null;
+                                if (testPlanReport.minimumAtVersion) {
+                                    earliestReleasedAt = new Date(
+                                        testPlanReport.minimumAtVersion.releasedAt
+                                    );
+                                    return (
+                                        new Date(item.releasedAt) >=
+                                        earliestReleasedAt
```
stalgiag, line 1405: At some point it might be good to break this out into a utility function so that we can unit test these comparisons to harden against any future changes to ATVersion. Not a requirement right now and I can easily add when doing the automation at version comparison work if preferred.

stalgiag (approved):
Code changes are clear and straightforward. I tested manually with VoiceOver by adding a test plan report with both minimum and exact requirements. With a minimum requirement, version options that do not meet the requirement are not shown in the dropdown. With an exact the warning is displayed and the version select is shown as a disabled text field with the required version. This reads/navigates well with VO.

Thanks for these changes!

client/components/TestRun/index.jsx
howard-e: I think I'll leave it up to a future PR because since there's a couple places where some date comparison is being done. Would rather some dedicated time be taken to pull those all together if creating a utility on that.

> ... and I can easily add when doing the automation at version comparison work if preferred

@stalgiag that's my preference right now, but great call out on this and definitely include me as a reviewer when doing so!

Issue #1143, opened 2024-07-01
Handling limitations on specificity in macOS runner images in VoiceOver version requirements 


stalgiag:
## Overview
While AT version requirements for running automation jobs, I encountered a limitation on the specificity of macOS version supported by Github runners that affects how we do VoiceOver version requirements. This issue is intended to bring this to the attention of the working group so that they can approve the proposed solution or offer an alternative.

## Outline of issue:
- [Github Actions](https://github.com/actions/runner-images?tab=readme-ov-file#available-images) do not support specifying that a runner with a specific minor version is used for macOS.
- Only the major version can be specified and the actual minor version is not known until the job is completed.
- Admins can specify a minor version in the required AT for a Test Plan Report. This makes version comparison to select a runner challenging.
- For NVDA these version comparisons and selections are much more straightforward because the AT version isn't attached to the runner OS in the same way.

## Example
If we were to continue to allow minor version requirements for VoiceOver, an example such as the following could occur:
1. Admin creates a new VoiceOver Test Plan Report with a minimum version of 13.0.1
2. The app can only specify that a workflow runs on a runner with a major version. The relevant major versions in this case are 13 or 14.
3. The only version that meets the minimum requirement of 13.0.1 is 14 even though it likely has less in common with 13.0.1 than version 13 does.  Adding more complexity, Github could even be running 13.5 for jobs that specify version 13 but we do not know this before starting the job.
4. The app has to decide whether automation supports this Test Plan Report. Since macOS 14 technically meets the criteria, the app offers automation as an option for running the Test Plan Run, and upon starting the collection job uses 14.

## Proposed solution:
We limit the required and minimum AT version selection for VoiceOver to major versions. The bot can still report which minor version it used during the Test Plan Run.

PR #1144, opened 2024-07-02
Specify AtVersion for automation

stalgiag:
## Overview
This PR adds AtVersion specification for automated jobs to the app.

Dependent on [aria-at-gh-actions-helper](https://github.com/bocoup/aria-at-gh-actions-helper/pull/21)

## Notes:
- Issue with Github only allowing selection of major macOS Version #1143
- Decided to go with a virtual property in the AtVersions model since separate app databases shouldn't diverge on these values. This also prevents needing to write multiple seeders as new versions are supported. This means that supported automation versions are maintained as a single array in `server/utils/constants.js`


server/util/getAtVersionWithRequirements.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+      pagination: {
+        order: [['releasedAt', 'ASC']]
+      },
+      transaction
+    });
+
+    const supportedAts = await Promise.all(
+      matchingAts.map(async version => {
+        const supportedByAutomation = await version.supportedByAutomation;
```
stalgiag, line 40: The asynchronous request for the `supportedByAutomation` property is necessary since it is virtual and the normal sequelize queries won't return it. Not sure if there are better options.

server/scripts/populate-test-data/pg_dump_test_data.sql
```diff
@@ -31,6 +31,8 @@ $$;
 -- Data for Name: AtVersion; Type: TABLE DATA; Schema: public; Owner: atr
 --
 
+INSERT INTO "AtVersion" ("atId", "name", "releasedAt") VALUES (2, '2023.3.3', '2023-12-02');
+
```
stalgiag, line 35: Added an AtVersion to the test database so that we can have a supported AtVersion in the DB. This is thought to be safe because it is the test database only but please let me know if this is not the case.

server/services/GithubWorkflowService.js
```diff
@@ -158,7 +164,8 @@ const createGithubWorkflow = async ({ job, directory, gitSha }) => {
       'X-GitHub-Api-Version': '2022-11-28'
     },
     data: JSON.stringify({
-      ref: 'main',
+      // TODO: Set back to 'main' once this branch is merged
+      ref: 'support-specific-at-runs',
```
stalgiag, line 168: Note to remember to change this before merging

server/util/constants.js
```diff
@@ -1,6 +1,15 @@
 // Mirror of constant in /client/components/TestRenderer/OutputTextArea/constants.js
 const NO_OUTPUT_STRING = 'No output was detected.';
 
+const AT_VERSIONS_SUPPORTED_BY_COLLECTION_JOBS = {
+  // These should be tracked with the available version in the github workflow
+  // TODO: Add link once feature branch for these changes is merged in aria-at-gh-actions-helper
```
stalgiag, line 6: Note to remember to add this link before merging

server/util/getAtVersionWithRequirements.js
```diff
@@ -0,0 +1,61 @@
...
+        return supportedByAutomation ? version.toJSON() : null;
+      })
+    );
+
+    const latestSupportedAt = supportedAts.filter(Boolean)[0];
```
gnarf, line 45:
```suggestion
    const latestSupportedAt = supportedAts.find(Boolean);
```

gnarf (commented):
I'm a little curious the reasoning behind setting up the `supportedByAutomation` flag as a "code" data source vs making it a bool in the sql table then having a "helper" for writing a quick "seeder" to insert (on conflict update) a specific supported version into the table.  It feels very weird to have to `await atVersion.supportedByAutomation`

stalgiag:
Since this is something that is currently only updated by coders I thought it would be nice to have a single source of truth. Otherwise, we could potentially have dozens of related seeders down the road, and we would still need to maintain the constant array as well because an `AtVersion` might not be in the database when it becomes supported by automation. I have no indication that we will be automatically creating those `AtVersion`s so that column will need to be determined on row creation. The wonkiness to get the virtual property seemed like a worthy tradeoff at the time but I could be missing complexity that will emerge down the road. I am open to changing it.

gnarf:
> The wonkiness

I think the root of "the wonk" here is trying to tie it into the `Model` at all.  It might make more sense as a `resolvers/AtVersion/supportedByAutomationResolver.js`  with a cache & stored strategy for the atNameById map.

It's always going to be a little wonk needing at least one async step, but I do think you could have a trio of methods exported from that resolver that were `supportedByAutomationResolver(atVersion)` which internally did `await prepareAtNameMap(); return supportedByAutomationResolverSync(atVersion);` - `prepareAtNameMap` would be setup on only "execute" the query once, and store the results in some var the resolver can use forever more.

Of course, this is it's own version of the jank, but does have a considerable advantage of at least not needing to do a roundtrip to postgres for every AtVersion in `getAtVersionWithRequirements` - and not having to do an extra roundtrip to postgres for every AtVersion exported in a gql query returning a bunch of them.

I'm not pushing super hard for this change, just putting it out there in case you also think it sounds "better" in some way

stalgiag:
This is an interesting suggestion and I can see the value. I am definitely operating on the assumption that AtVersions will never be a huge collection but I think your solve would be better. I am not sure if it will make sense for me to prioritize that refactor and I'll have to check in with Carmen on that. 

On another note: I just merged `development` in and the new tests are failing. Some work will need to be done to make these changes work with the new end-to-end tests. I will tackle the issue but just wanted to note this here!

stalgiag:
Just noticed the new conflicts. Will handle today after my meetings

server/scripts/populate-test-data/pg_dump_test_data.sql
howard-e: Seems fine to me.

server/services/GithubWorkflowService.js
```diff
@@ -147,6 +147,12 @@ const createGithubWorkflow = async ({ job, directory, gitSha }) => {
   if (atKey === 'nvda') {
     inputs.test_pattern = '{reference/**,test-*-nvda.*}';
     inputs.browser = browser;
+    inputs.nvda_version = atVersion?.name;
+  }
+  if (atKey === 'voiceover_macos') {
+    // We just want the whole number of the macOS version
+    // due to limitations on Github workflow runners
```
howard-e, line 154: Thanks for leaving this comment! Outside the context of this PR, future readers may be curious on which limitation. Could you also link https://github.com/w3c/aria-at-app/issues/1143 here?

howard-e (commented):
Looks good! Also confirmed I can't assign a bot for versions of ATs not listed in `server/util/constants.js`. I've left some additional inline comments.

I also agree with Corey's above suggestions but not something to block on

client/utils/automation.js
```diff
@@ -1,16 +1,80 @@
...
-      (browserKey === 'chrome' || browserKey === 'firefox')) ||
-    (atKey === 'voiceover_macos' && browserKey === 'safari_macos')
-  );
+
+  const { at, browser } = ctx;
+  const { key: atKey } = at;
+  const { key: browserKey } = browser;
+
+  // Check if the AT and browser are supported by the automation
+  if (
```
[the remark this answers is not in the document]
stalgiag: Agreed. Addressed in this context. We may want to consider more centralized constants at another point

server/models/services/CollectionJobService.js
```diff
@@ -383,15 +384,15 @@ const getCollectionJobs = async ({
  * @param {*} options.transaction - Sequelize transaction
  * @returns Promise<CollectionJob>
  */
-const triggerWorkflow = async (job, testIds, { transaction }) => {
+const triggerWorkflow = async (job, testIds, atVersion, { transaction }) => {
```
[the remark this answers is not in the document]
stalgiag: Thank you! I could've sworn I did a search but I clearly missed one. Thanks for this!

server/services/GithubWorkflowService.js
stalgiag: Agree. Done!

client/utils/automation.js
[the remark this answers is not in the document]
howard-e: > We may want to consider more centralized constants at another point.

Good point. Thanks for addressing for now!

server/models/services/CollectionJobService.js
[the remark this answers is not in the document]
howard-e: > Looks like that unit tests for retry cancelled jobs tech debt task is actually needed.

True! Do you want to do it in this PR?

howard-e (approved):
@stalgiag looks good! Thanks for addressing my recent PR feedback

server/models/services/CollectionJobService.js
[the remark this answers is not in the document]
stalgiag: I think Corey already [has an Asana task assigned for it](https://app.asana.com/0/1207770287602779/1206550116189397/f)

gnarf (approved):
Still lookin good! (the conflict game continues though....)

stalgiag:
Okay! Going to merge this quick before another set of conflicts comes through 😅 

Carmen made a [follow-up task](https://app.asana.com/0/1207768658408676/1207826403026130/f) based on @gnarf's [recommended approach](https://github.com/w3c/aria-at-app/pull/1144#issuecomment-2223304706). If there is time during this cycle I can take it on. Thanks all!

PR #1146, opened 2024-07-08
fix: Remove TestQueue v1 and update Test Queue e2e tests

howard-e:
This removes the original `components/TestQueue` and related components in favor of the updated Test Queue introduced in #1124.

This PR also updates and reorganizes tests related to the the original Test Queue and removes tests done using mocks and replaces them with e2e tests where appropriate.


client/components/TestRun/TestIframe/index.jsx
stalgiag, line 1: Looking at `main` it seems that this was no longer used and this is clean up because it isn't likely to be used in the future? Adding this comment as a verification measure

stalgiag (approved):
Tested manually thoroughly with both mouse + screen and screenreader and didn't encounter any bugs. Nice job removing this. Also really excited about the extensive new end-to-end tests!

Small confirmation comment inline on something that seemed a bit out of place. Otherwise approved!

client/components/TestRun/TestIframe/index.jsx
howard-e: Right! It was meant for a much older implementation of showing the test collection form before the `<TestRenderer>` was a thing. I was pretty surprised it was still there so using this as a chance to remove it.

PR #1148, opened 2024-07-10
feat: Update Dependencies

howard-e:
This PR updates several dependencies to the latest available minor and patch versions. Major version bumps are also included which weren't found to cause major breaking changes.

Also included is a revised `dependabot.yml` to lessen the noise and duplication of some dependabot PRs.


stalgiag (approved):
Tested rebuilding the database and the app and all works as expected. Changes make sense and are welcome. Thanks!

Issue #1145, opened 2024-07-06
Update reports to use new assertion verdict terminology for priority 3 (May) assertions


mcking65:
Per agreement in w3c/aria-at#1050, update report tables to express results for "May-Have" behaviors using the word "Supported" instead of "Passed" and "Unsupported" instead of "Failed".

In summary tables, such as [Alert Example | ARIA-AT Reports](https://aria-at.w3.org/report/111731), in the column "May-Have Behaviors", change string from "X of Y Passed" to "X of Y Supported".

In command results tables, such as [JAWS and Chrome for Alert Example | ARIA-AT Reports](https://aria-at.w3.org/report/111731/targets/250#result-NGNhZeyIxMiI6NTcwfQGMyNj), change the string in the "Result" column on rows where priority is "May" to "Supported" instead of "Passed" and "Unsupported instead of "failed".

Note that command results tables appear on:
1. Rreport pages
2. In the candidate review pages
3. In the test runner for tests that are complete.

ccanash:
Noted @mcking65 . We will prioritise it.

PR #1149, opened 2024-07-10
feat: terminology update, passed/failing ->  supported/unsupported

stalgiag:
This PR addresses #1145 

I changed all of the string constants in the UI code, updated the `getMetrics` function to use the updated terms in its `__Formatted` template strings, and added a migration for updating the `testPlanReports.metrics.__Formatted` in the database.

I stopped myself short of renaming variables and properties. I felt that the term `passed` was still useful for understanding the purpose of the variables. I do not have a strong take on this and I am open to expanding these changes to include variable names.

---
  - https://app.asana.com/0/0/1207778064599278


howard-e (commented):
@stalgiag thanks for getting this done so quickly! But looks like there was a misunderstanding where the terminology changes needed to happen.

According to #1145, it should only apply to "May" assertions. The "Must" and "Should" terminology should remain unchanged. https://github.com/w3c/aria-at/issues/1050#issuecomment-2206564252 also presents a table to show the distinction further for the results tables.

stalgiag:
Haha thanks 🤦‍♀️

Issue #1108, opened 2024-05-22
Manage Bot Run Dialog should  include "View Log" link


gnarf:
Currently you need to "Open run as" to see the view logs link, but it should be available in this dialog.

![Screenshot of "Manage VoiceOver Bot Run" dialog from aria-at.w3.org](https://github.com/w3c/aria-at-app/assets/549355/6103b5f3-87aa-450c-b147-ef9b8330b80c)

PR #1152, opened 2024-07-11
Add "View Logs" button to Manage Bot Run Dialog

stalgiag:
addresses #1108 

Opted for a button link with proper aria attributes in order to get disabled and enabled behavior that matches the rest of the buttons in the dialog.

---
  - https://app.asana.com/0/0/1207387087905472

### DIFF
--- a/client/components/ManageBotRunDialog/ViewLogsButton/index.jsx
+++ b/client/components/ManageBotRunDialog/ViewLogsButton/index.jsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from 'react-bootstrap';
+
+const ViewLogsButton = ({ externalLogsUrl }) => {
+  const onClick = () => {
+    window.open(externalLogsUrl, '_blank');
+  };
+
+  return (
+    <Button
+      disabled={!externalLogsUrl}
+      aria-disabled={!externalLogsUrl}
+      onClick={onClick}
+      role="link"
+      aria-label={`View logs at ${externalLogsUrl}`}
+      variant="secondary"
+    >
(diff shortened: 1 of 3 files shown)


client/components/ManageBotRunDialog/ViewLogsButton/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+    window.open(externalLogsUrl, '_blank');
+  };
+
+  return (
+    <Button
+      disabled={!externalLogsUrl}
+      aria-disabled={!externalLogsUrl}
+      onClick={onClick}
+      role="link"
+      aria-label={`View logs at ${externalLogsUrl}`}
```
howard-e, line 16:
```suggestion
      aria-label={`View logs at ${externalLogsUrl}`}
```

This will still say `View logs at null` until the url IS available to a screen reader. Do we want another defaulted statement while it isn't ready?

Alternatively, I'm wondering if announcing the link itself may be unnecessary and the button's description of just "View Log" is sufficient.

howard-e (commented):
@stalgiag looks good! Confirmed using data from the sandbox as well to make sure it links as expected and also confirmed the button is present locally (but disabled).

Code looks good, I've left just a small inline comment.

client/components/ManageBotRunDialog/ViewLogsButton/index.jsx
stalgiag: Good point! I removed it.

howard-e: Thanks for checking that out!

howard-e (approved):

PR #1153, opened 2024-07-15
Reset TestPlanVersion `nextval` during import if IDs aren't used

howard-e:
I was inspecting the production's data from July 10 2024 and took note of significantly high id numbers in the `TestPlanVersion` table. The latest row's id is `163615` but there is just 103 rows in the table, which uses auto-increment IDs. Running `select currval(pg_get_serial_sequence('"TestPlanVersion"', 'id'))` also gives `175465`.

This happens because the `import-tests` script is dependent on knowing the "next" id to create relevant data that is retrieved using [postgres' nextval (which also also updates the next possible ID value)](https://github.com/w3c/aria-at-app/blob/1f61c87fa0bad33e63e2ce145fe989942b332dd8/server/scripts/import-tests/index.js#L143:L148). The issue with this is that those ids could go unused if a potential TestPlanVersion being created is ["discarded" if non-unique data is found](https://github.com/w3c/aria-at-app/blob/1f61c87fa0bad33e63e2ce145fe989942b332dd8/server/scripts/import-tests/index.js#L183), but the incremented value is left as is.

The amount of test plan versions that could be imported is 35 and the cronjob responsible for doing that **should** run [every 15 minutes](https://github.com/w3c/aria-at-app/blob/76c49246b54ea22545e8b872a8838c20cdf82587/deploy/roles/application/tasks/cron.yml#L3:L8) . So potentially, the incremented value is unnecessarily bumped 140 times per hour (105 at a minimum) if there are no new test plan versions.

So this PR does the following:
* Resets `nextval` if it hasn't been used due to that non-unique data condition (continued using `nextval` for the evaluation of the reset value because `currval` or `lastval` may not successfully run if `nextval` hasn't been ran during the db's transaction session)
* Resets `nextval` to current latest ID + 1 in a migration

### DIFF
--- a/server/scripts/import-tests/index.js
+++ b/server/scripts/import-tests/index.js
@@ -141,13 +141,20 @@ const buildTestsAndCreateTestPlanVersions = async (commit, { transaction }) => {
     // Gets the next ID and increments the ID counter in Postgres
     // Needed to create the testIds - see LocationOfDataId.js for more info
     const [testPlanVersionIdResult] = await sequelize.query(
-      `SELECT nextval(
-                pg_get_serial_sequence('"TestPlanVersion"', 'id')
-            )`,
+      `SELECT nextval(pg_get_serial_sequence('"TestPlanVersion"', 'id'))`,
       { transaction }
     );
     const testPlanVersionIdResultRow = testPlanVersionIdResult[0];
     const testPlanVersionId = testPlanVersionIdResultRow.nextval;
+
+    // Get the currently set value to rollback the 'correct' nextval for
+    // subsequent runs
+    const [currentTestPlanVersionIdResult] = await sequelize.query(
+      `SELECT currval(pg_get_serial_sequence('"TestPlanVersion"', 'id'))`,
+      { transaction }
+    );
+    const currentTestPlanVersionId =
+      currentTestPlanVersionIdResult[0].currval - 1;
 
     // Target the specific /tests/<pattern> directory to determine when a pattern's folder was
     // actually last changed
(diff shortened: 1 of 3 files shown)


server/tests/integration/graphql.test.js
```diff
@@ -181,444 +181,444 @@ describe('graphql', () => {
...
+            __typename
+            id
+            status
+          }
+          collectionJobByTestPlanRunId(testPlanRunId: 1) {
+            __typename
+            id
+            status
+          }
+          v2TestPlanVersion: testPlanVersion(id: 80) {
```
howard-e, line 273: For reviewers: Formatting has messed with this diff. This id was changed to 80 from 133

server/tests/integration/graphql.test.js
```diff
@@ -181,444 +181,444 @@ describe('graphql', () => {
...
+            id
+            testPlanReports {
+              id
+            }
+            testPlan {
+              id
+              directory
+            }
+          }
+          recommendedTestPlanVersion: testPlanVersion(id: 67) {
```
howard-e, line 381: For reviewers: Formatting has messed with this diff. This id was changed to 67 from 69

server/scripts/import-tests/index.js
```diff
@@ -141,14 +141,21 @@ const buildTestsAndCreateTestPlanVersions = async (commit, { transaction }) => {
...
     const testPlanVersionId = testPlanVersionIdResultRow.nextval;
 
+    // Get the currently set value to rollback the 'correct' nextval for
+    // subsequent runs
+    const [currentTestPlanVersionIdResult] = await sequelize.query(
+      `SELECT currval(pg_get_serial_sequence('"TestPlanVersion"', 'id'))`,
+      { transaction }
+    );
+    const currentTestPlanVersionId =
+      currentTestPlanVersionIdResult[0].currval - 1;
```
stalgiag, line 157: Took me a second to understand why `currval -1` works here even with the comment helping explain but I eventually understood and the approach makes sense.

stalgiag (approved):
This fix makes sense and I tested rebuilding the DB and migrating a production dump without encountering any issues.

PR #1160, opened 2024-07-17
fix: onHide error thrown on `ESC` press when `<UpdateVersionModal />` is open

howard-e:
I was opening the `Manage Assistive Technology Versions` disclosure on the Test Queue page to add/edit AT verions. Found that if I hit `ESC` on my keyboard to close either modals, an error is thrown which reads `onHide is not a function` where the stacktrace points to a line inside `react-bootstrap`'s Modal component. Screenshot:

Screenshot showing onHide error

I confirmed this doesn't happen on production so my only guess is this error was introduced in #1148 when `react-bootstrap` was bumped from `2.7.0` to `2.10.4`. Although, nothing obvious jumps out at me when viewing the [release notes](https://github.com/react-bootstrap/react-bootstrap/releases).

This PR should restore the functionality that's on production (although I'm also wondering if the title refocus instead of closing the modal is intentional).

### DIFF
--- a/client/components/common/BasicModal/index.jsx
+++ b/client/components/common/BasicModal/index.jsx
@@ -79,8 +79,8 @@ const BasicModal = ({
         id={`focus-trapped-${id}`}
         centered={centered}
         animation={animation}
-        onHide={useOnHide ? handleHide || handleClose : null}
-        onExit={!useOnHide ? handleHide || handleClose : null}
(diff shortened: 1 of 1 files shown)


stalgiag (approved):
Odd! If that is true then there might be a regression in that library. Surely they aren't calling `null` props as functions?? Either way, this change is small on our end and seems less fragile to upstream problems. I am going to approve and merge so the problem doesn't wind its way to prod.

PR #1165, opened 2024-07-22
Decompose and organize `import-tests` script

stalgiag:
This pull request reorganizes and documents the code in the import-tests script. I tried to avoid changing functionality and only made minor code changes beyond breaking big functions up, grouping related functions into specific modules, and add JSDoc comments.

The database builds and the tests pass. I assume that this script is so foundational to DB construction that at least one of the tests would surface an errant change in functionality. If this is not the case, and there is room for this to introduce breaking changes, then I am okay with not merging out of an abundance of caution.

I also recognize that this reorganization is highly damaging to the git history. If we don't feel that the tradeoff in readability/maintainability is worth it then I would be happy to discard this PR.

---
  - https://app.asana.com/0/0/1207768136984532


server/scripts/import-tests/gitOperations.js
```diff
@@ -0,0 +1,83 @@
...
+  const gitMessage = gitRun(`log -1 --format=%s -- .`, directoryPath);
+  const gitCommitDate = new Date(
+    gitRun(`log -1 --format=%aI -- .`, directoryPath)
+  );
+
+  return { gitSha, gitMessage, gitCommitDate };
+}
+
+module.exports = {
+  gitRun,
```
howard-e, line 79: Doesn't look like this is getting used outside this function so no need to export (and perhaps for the best to keep the operations contained and true to the file's purpose)

server/scripts/import-tests/settings.js
```diff
@@ -0,0 +1,7 @@
+const path = require('path');
+
+const gitCloneDirectory = path.resolve(__dirname, 'tmp');
```
howard-e, line 3: This also reminds me, could you include `server/scripts/import-tests/tmp` in the `.gitignore`?

I've been guilty in the past of accidentally including it in a `git add .` (but caught it!) while an import was still happening. It's such a quick process that I suppose it's been missed all this time

server/scripts/import-tests/testParser.js
```diff
@@ -0,0 +1,423 @@
...
+  createAssertionId
+} = require('../../services/PopulatedData/locationOfDataId');
+const { convertAssertionPriority } = require('shared');
+const { getAppUrl } = require('./utils');
+const deepPickEqual = require('../../util/deepPickEqual');
+
+/**
+ * Strategies for different test format versions.
+ */
+const testFormatStrategies = {
```
howard-e, line 15: Neat!

server/scripts/import-tests/testParser.js
```diff
@@ -0,0 +1,423 @@
...
+  v1: {
+    createTest: createV1Test,
+    createRenderableContent,
+    createRenderedUrls,
+    createScenarios
```
howard-e, line 20: nit: continuining to use the v1 identifier in the function names could be pretty useful to not show any "default" preference or priority.

server/scripts/import-tests/testPlanVersionOperations.js
```diff
@@ -0,0 +1,481 @@
...
+
+  return `${versionStringBase}-${duplicateCount}`;
+}
+
+module.exports = {
+  buildTestsAndCreateTestPlanVersions,
+  importHarness,
+  updateJsons,
+  updateAtsJson,
+  getVersionString
```
howard-e, line 480: These 2 also aren't being called elsewhere

howard-e (commented):
@stalgiag this is such a good separation of concerns! Thanks much! I've left a few inline comments and suggestions but nothing blocking

howard-e (approved):
@stalgiag this looks good to me!

> DB construction that at least one of the tests would surface an errant change in functionality. If this is not the case, and there is room for this to introduce breaking changes

Yep a test should exist to confirm the state of the TestPlanVersion rows is the same before and after an import.

The other tests not breaking still serve as an indicator that all _should_ be well given they're built around the import script but an explicit test like above should be done. It can be done in this PR or in another, I don't think this has to be blocked on that.

stalgiag:
Adding a note to wait to merge this until after additional testing from #1169 and #1175 is merged in and confirmed

server/scripts/import-tests/settings.js
stalgiag: Done!

PR #1169, opened 2024-07-22
fix: Update and expand `/client` tests

howard-e:
This is a supporting PR for #1147 as well for expanding the current `/client` test suites. Changes here will contain a subset of of #1147.


stalgiag (approved):
Wow! This is great. This is a ton of coverage. 

It isn't in the scope for this PR because so much is already happening here and this lays a great foundation but I just want to make a note about adding more multistep, multicomponent interactions to some of the pages in the future. These kinds of tests become more important with snapshot testing checking basic things like content rendered and links etc. This frees us up to do more user-workflow imitation. For example, the Data Management page could test something like:

1. Advance to Recommend
2. Verifyy content of confirmation modal
3. Confirm with modal
4. Check updated content in row

You are doing some testing like this, especially in the TestRun view. I am down to merge as-is since this PR is a major accomplishment. This note is more of a message for both of us to keep an eye out for opportunities for user-workflow testing and as we add more of that, and snapshot testing solidifies, we might consider scaling back some of the content and link testing.

PR #1172, opened 2024-07-24
fix: Update metrics calculations and related UI components

howard-e:
See #1173 and #1174

This PR addresses several issues:
* Persist proper metrics values to database when assertionExceptions exist (this wasn't being fully handled during a server calculation which could lead to incorrect values when referencing the database stored metrics values -- but was already being recalculated and accounted for when displaying the values on the client so there should be no noticeable change in those cases)
* Historically, ARIA-AT App was presenting only the calculated value of the MUST assertions for the progress bars and the associated messages on the Reports and Candidate Review Page. I can only guess this was because of the previously used REQUIRED (now MUST) and OPTIONAL (now SHOULD) priorities. But now that there is MUST, SHOULD and MAY, the representation of that progress bar now additionally includes SHOULD priority assertions.
* Replaces the usages of `Math.round()` with `Math.floor()` in presenting metrics' `supportPercent`.
* Adds additional tests for metrics calculations.
* Revised use of aria-label on progressbar component.


server/resolvers/TestPlanReport/finalizedTestResultsResolver.js
```diff
@@ -8,6 +8,9 @@ const finalizedTestResultsResolver = async (testPlanReport, _, context) => {
   // Return the primary test plan run, otherwise pick the first TestPlanRun found.
   const testPlanRun =
     testPlanReport.testPlanRuns.find(({ isPrimary }) => isPrimary) ||
+    testPlanReport.testPlanRuns.find(testPlanRun =>
+      testPlanRun.testResults?.some(testResult => !!testResult.completedAt)
+    ) ||
```
howard-e, line 13: For reviewers: Found odd behaviors when doing the following:
* Assigning self to report row on Test Queue
* Assigning other tester to report row on Test Queue
* Opening the report as the other tester navigating and this resolver could silently throw errors when navigating so attempted to resolve that

stalgiag (approved):
Tested locally and this resolves the issues as listed. The code looks good and the unit tests are welcome since it is difficult to get the DB into the various related testing states. Looks good to me!

PR #1175, opened 2024-07-25
Basic snapshot testing

stalgiag:
This PR introduces basic snapshot testing for each major route.

Notes:
- I opted to use a limited set of routes after trying a version that [tested all available routes](https://github.com/w3c/aria-at-app/commit/b3ca12cb9ac31afe4523e182aac381d2860fe231). I felt that this would be too heavy a lift for testing and maintaining.
- A number of elements are removed due to unpredictability in generation. Some of these are intentionally dynamic and some have unexpected inconsistencies between my local and CI. I have clearly marked each case in `clients/tests/e2e/snapshots/util.js`. Some of these may be removed after the underlying inconsistency is investigated.
- Snapshot testing can present challenges with maintenance and upkeep. I'd invite reviewers to consider whether the benefit of having an easy-to-test basic content rendering is worth the tradeoff for occasionally ballooning our pull requests due to updated snapshots. I am of the opinion that it is worth it in this limited scope but I am very open to discussing the pros + cons.


client/tests/e2e/snapshots/utils.js
```diff
@@ -0,0 +1,113 @@
...
+  const content = await page.content();
+
+  // Strip out "Previously Viewed" badges and other dynamic content
+  let cleanedContent = content.replace(
+    /<span[^>]*class="[^"]*viewed-badge[^"]*"[^>]*>Previously Viewed<\/span>/g,
+    ''
+  );
+
+  // Remove href and version number on data management page
+  // Addresses inconsistent links and date -> version number
```
howard-e, line 62: A think a TODO should be included for ones like this so we can explore and remove this cleaner when necessary, explaining what the issue is.

There certainly shouldn't be a mismatch with the version dates being displayed no matter the environment.

client/tests/e2e/snapshots/utils.js
```diff
@@ -0,0 +1,113 @@
...
+  cleanedContent = cleanedContent.replace(
+    /<span[^>]*class="[^"]*full-width[^"]*"[^>]*>\s*<a href="\/test-review\/\d+">\s*<span>\s*<svg[\s\S]*?<\/svg>\s*<b>V\d+\.\d+\.\d+(?:-\d+)?<\/b>\s*<\/span>\s*<\/a>\s*<\/span>/g,
+    '<span class="full-width"><a><span><svg>...</svg><b>VERSION_REMOVED</b></span></a></span>'
+  );
+
+  // Strip out randomly generated IDs
```
howard-e, line 68: Makes sense!

howard-e (approved):
In acknowledging that this is an extra layer of maintenance and developing of new features, it's also a huge plus in ensuring the accuracy of such a data heavy system. Kudos for getting this going! Looks good to me!

I've left only a small inline comment but feel free to merge when ready.

PR #1177, opened 2024-07-25
Use UTC time for date strings

stalgiag:
Targeting the `snapshot-testing` branch since there are related changes here but broke this out into a separate PR since that PR has already been reviewed and the changes here will have effects outside of snapshot testing.

While writing snapshot tests, I noticed that dates were inconsistent between my system and CI. Based on @howard-e 's advice, I changed all of the calls to `moment` to be `moment.utc`. This fixes the problem.

The only remaining unexpected issue uncovered by the snapshot tests is related to the rollback of the 'Ready to Review', 'Review in Progress', and 'Previously Viewed' statuses after e2e testing. This can be addressed separately.

### DIFF
--- a/client/tests/e2e/snapshots/saved/_candidate-review.html
+++ b/client/tests/e2e/snapshots/saved/_candidate-review.html
@@ -124,8 +124,8 @@
                           >Modal Dialog Example V22.05.26 (18 Tests)</a
                         >
                       </th>
-                      <td class="css-niacka"></td>
-                      <td class="css-niacka"></td>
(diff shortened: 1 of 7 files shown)


howard-e (approved):
Looks good! Happy this was found and at least this is just a display concern. It doesn't seem like it _directly_ affected any database persisted data. I figured there could be an issue with how the app was searching created GH issues with the issuesResolver search but haven't observed any problems there.

Looking back at this and where the project currently is as well, think it may be best to eventually move these date utilities to the `shared` project and serve it from there. There is already a minor duplication happening here but it will make it easier to switch out if needed.

PR #1181, opened 2024-07-30
Ensure that "Target Days X Past" is not included in snapshot

stalgiag:
This is the first day after merging the snapshot testing and I realized that there is another date consideration, the "Target Days X Past" values will increment by one with each day. This means that if we test this value we would need to update the snapshots daily.

Since this is a small fix needed for dev today, it only affects testing without any implications for the app, and the only applicable reviewer is out today I am going to merge once the tests pass. I can revert if there are issues created by this.

cc for when you're back online @howard-e

### DIFF
--- a/client/tests/e2e/snapshots/utils.js
+++ b/client/tests/e2e/snapshots/utils.js
@@ -35,14 +35,17 @@ async function cleanAndNormalizeSnapshot(page) {
         const text = el.textContent.trim();
         if (
           text.includes('Ready for Review') ||
-          text.includes('Review in Progress')
+          text.includes('Review in Progress') ||
+          (text.includes('Days') && text.includes('Past'))
         ) {
           el.remove();
         }
       });
     }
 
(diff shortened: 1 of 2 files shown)


client/tests/e2e/snapshots/utils.js
```diff
@@ -35,14 +35,17 @@ async function cleanAndNormalizeSnapshot(page) {
         const text = el.textContent.trim();
         if (
           text.includes('Ready for Review') ||
-          text.includes('Review in Progress')
+          text.includes('Review in Progress') ||
+          (text.includes('Days') && text.includes('Past'))
```
howard-e, line 39: You'll also want to check for ['Days Away'](https://github.com/w3c/aria-at-app/blob/7f6d795306253035bd65aa574721f02a502bf80d/client/components/DataManagement/DataManagementRow/index.jsx#L940)

howard-e (commented):
@stalgiag good catch! You may also want to account for the 'Days Away' text

PR #1182, opened 2024-07-30
Improved `atVersion.supportedByAutomation` field handling

stalgiag:
This PR adds a custom resolver for `atVersion.supportedByAutomation` that uses caching.

### DIFF
--- a/client/components/AddTestToQueueWithConfirmation/index.jsx
+++ b/client/components/AddTestToQueueWithConfirmation/index.jsx
@@ -66,7 +66,13 @@ function AddTestToQueueWithConfirmation({
   );
 
   const existingTestPlanReports =
-    existingTestPlanReportsData?.existingTestPlanVersion?.testPlanReports;
+    existingTestPlanReportsData?.existingTestPlanVersion?.testPlanReports.filter(
+      tpr =>
+        tpr.at?.id === at?.id &&
+        tpr.browser?.id === browser?.id &&
+        tpr.minimumAtVersion?.id === minimumAtVersion?.id &&
+        tpr.exactAtVersion?.id === exactAtVersion?.id
+    );
 
   let latestOldVersion;
   let oldReportToCopyResultsFrom;
--- a/server/resolvers/AtVersion/index.js
+++ b/server/resolvers/AtVersion/index.js
@@ -1,7 +1,7 @@
-const supportedByAutomationResolver = require('./supportedByAutomationResolver');
+const supportedByAutomation = require('./supportedByAutomationResolver');
 
 const AtVersion = {
-  supportedByAutomation: supportedByAutomationResolver
+  supportedByAutomation
 };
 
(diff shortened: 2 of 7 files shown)


client/components/AddTestToQueueWithConfirmation/index.jsx
```diff
@@ -66,7 +66,13 @@ function AddTestToQueueWithConfirmation({
   );
 
   const existingTestPlanReports =
-    existingTestPlanReportsData?.existingTestPlanVersion?.testPlanReports;
+    existingTestPlanReportsData?.existingTestPlanVersion?.testPlanReports.filter(
+      tpr =>
+        tpr.at?.id === at?.id &&
+        tpr.browser?.id === browser?.id &&
+        tpr.minimumAtVersion?.id === minimumAtVersion?.id &&
+        tpr.exactAtVersion?.id === exactAtVersion?.id
```
howard-e, line 74: To the best of my understanding, only one of these may exist at a time so conditionally checking may be good, with `exactAtVersion` being the priority. Does that change your thoughts on this comparison?

I suppose it shouldn't matter though, since in the instance of the one (exact or minimum) that doesn't exist, `undefined === undefined` will be evaluated to true anyways anyways.

server/resolvers/AtVersion/index.js
```diff
@@ -0,0 +1,7 @@
+const supportedByAutomationResolver = require('./supportedByAutomationResolver');
+
+const AtVersion = {
+  supportedByAutomation: supportedByAutomationResolver
```
howard-e, line 4: nit: just for consistency, the other files in `/resolver` import the resolver as the expect attribute name, so `const supportedByAutomation = require('./supportedByAutomationResolver');`

howard-e (commented):
This is a very welcome addition! Had to do some special handling of this field in #1147 and this should make it a whole lot simpler.

I left small comments inline.

server/resolvers/AtVersion/index.js
stalgiag: Good call!

client/components/AddTestToQueueWithConfirmation/index.jsx
stalgiag: Yeah I think it still makes sense because we are ensuring the Reports are a match and so if one has an `undefined` minimum then the other would as well. One or the other is going to be a useless check since having only a minimum or a required is probably enforced elsewhere but I think it is a lower cognitive load to read than doing an `OR` between those two.

howard-e: Fair enough on the cognitive load, this is fine with me.

stalgiag: Thank you!

howard-e (approved):
Looks good, thanks for addressing the feedback

Issue #1167, opened 2024-07-22
Add Test plan versions to Candidate review pages


ccanash:
Add the versions from data management to the CR pages.

Issue #1166, opened 2024-07-22
Add Run history information to Candidate review pages


ccanash:
Bring over the run-history disclosure from the test-plan-report page (e.g. /report/2407/targets/201)

PR #1186, opened 2024-08-01
feat: Content updates to Candidate Review page

stalgiag:
This PR does the following:
- Adds At Version to heading in Candidate Review Page
- Adds Test Plan Version name and a link to the corresponding `test-review/` page to the Candidate Review page
- Adds Run History to the Candidate Review page

addresses #1166 
addresses #1167 
addresses #1168


client/components/Reports/getTestersRunHistory.js
stalgiag, line 1: I am not entirely satisfied by this placement but there wasn't a strong enough organization strategy in place for content-related utilities used by more than one component. Ideally, it would be a `useMemo` custom hook but that would've required some restructuring. Open to alternatives but this seemed acceptable for now.

howard-e (commented):
A detail for consideration was missed when describing what should be done here. Since the Candidate Test Plan Run page will list the results of several Browsers if found, how the Run History render function is being called here will hide results for all the n+1 reports, as shown in the screenshot.

screenshot showing only run history for testers that ran Firefox but could there should also be run history for Chrome

This may require some restructuring of the `getTestersRunHistory` function to get them into the single disclosure. But having separate Run History disclosures so as to not make extensive changes there also seem okay to me. ie.
- Test Instructions
- Test Results for Firefox
- Run History for Firefox
- Test Results for Chrome
- Run History for Chrome

stalgiag:
> A detail for consideration was missed when describing what should be done here. Since the Candidate Test Plan Run page will list the results of several Browsers if found, how the Run History render function is being called here will hide results for all the n+1 reports, as shown in the screenshot.
> 
> screenshot showing only run history for testers that ran Firefox but could there should also be run history for Chrome
> 
> This may require some restructuring of the `getTestersRunHistory` function to get them into the single disclosure. But having separate Run History disclosures so as to not make extensive changes there also seem okay to me. ie.
> 
> * Test Instructions
> * Test Results for Firefox
> * Run History for Firefox
> * Test Results for Chrome
> * Run History for Chrome

Ah good catch! I had to get my DB into a different state to see this case. I decided to go ahead and turn this into a dedicated component. I made the `RunHistory` component just the content rather than the disclosure since this content might be useful in other contexts. I made it support multiple `testPlanReport`s so that all of the browser + at variations can be seen in one place.

stalgiag:
> A detail for consideration was missed when describing what should be done here. Since the Candidate Test Plan Run page will list the results of several Browsers if found, how the Run History render function is being called here will hide results for all the n+1 reports, as shown in the screenshot.
> 
> screenshot showing only run history for testers that ran Firefox but could there should also be run history for Chrome
> 
> This may require some restructuring of the `getTestersRunHistory` function to get them into the single disclosure. But having separate Run History disclosures so as to not make extensive changes there also seem okay to me. ie.
> 
> * Test Instructions
> * Test Results for Firefox
> * Run History for Firefox
> * Test Results for Chrome
> * Run History for Chrome

Decided to go with the combined run history under one disclosure with a dedicated component. Thanks for pointing this out!

client/components/CandidateReview/CandidateTestPlanRun/index.jsx
```diff
@@ -391,9 +394,11 @@ const CandidateTestPlanRun = () => {
...
         <div className="info-label">
           <b>Candidate Test Plan:</b>{' '}
-          {`${
-            testPlanVersion.title || testPlanVersion.testPlan?.directory || ''
-          }`}
+          <a href={`/test-review/${testPlanVersion.id}`}>
+            {`${
+              testPlanVersion.title || testPlanVersion.testPlan?.directory || ''
+            } ${testPlanVersion.versionString}`}
+          </a>
```
howard-e, line 401: Note to ourselves to also bring this to the TestRun page as well!

howard-e (approved):
Looks good. Thanks for addressing the feedback. Getting that content into the single disclosure certainly looks better!

howard-e:
@stalgiag there is now a merge conflict here but feel free to merge once it's resolved

PR #1188, opened 2024-08-01
feat: Test navigation via URI fragment

stalgiag:
See title.

This adds specified test navigation via URI fragment to `TestRun` and `CandidateTestPlanRun`.

---
  - https://app.asana.com/0/0/1207890671031952

### DIFF
--- a/client/hooks/useUrlTestIndex.js
+++ b/client/hooks/useUrlTestIndex.js
@@ -1,21 +1,27 @@
 import { useLocation, useNavigate } from 'react-router-dom';
 import { useEffect, useState } from 'react';
 
-export const useUrlTestIndex = () => {
+export const useUrlTestIndex = maxTestIndex => {
   const location = useLocation();
   const navigate = useNavigate();
   const [currentTestIndex, setCurrentTestIndex] = useState(0);
 
   const getTestIndex = () => {
     // Remove the '#' character
-    const fragment = location.hash.slice(1);
+    const fragment = parseInt(location.hash.slice(1), 10) || 1;
+
+    if (!maxTestIndex || maxTestIndex < 1) {
+      // If the max test index is less than 1, return 0
+      return 0;
+    }
     // Subtract 1 to make it 0-indexed
-    return Math.max(parseInt(fragment || '1', 10) - 1, 0);
+    // and clamp to the max test index
+    return Math.max(0, Math.min(fragment - 1, maxTestIndex - 1));
   };
 
   useEffect(() => {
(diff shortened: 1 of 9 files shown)


stalgiag:
> The case where one navigates to #X where 'X' isn't a valid test number has to be handled though.

Oh thanks for thinking of this! I'll add the fallback behavior and a related test.

client/hooks/useUrlTestIndex.js
```diff
@@ -0,0 +1,33 @@
...
+export const useUrlTestIndex = maxTestIndex => {
+  const location = useLocation();
+  const navigate = useNavigate();
+  const [currentTestIndex, setCurrentTestIndex] = useState(0);
+
+  const getTestIndex = () => {
+    // Remove the '#' character
+    const fragment = parseInt(location.hash.slice(1), 10) || 1;
+
+    if (!maxTestIndex || maxTestIndex < 1) {
```
howard-e, line 13: Messing with this a bit further handling of something like `#-1` would still cause an issue at https://github.com/w3c/aria-at-app/blob/547497802032eab678a3e5bc4035dc2db6dfce19/client/utils/ScrollFixer.jsx#L47 unless the validity of the selector is checked (but I can't see something like that happening unless trying to force it)

This has also made me aware that additional routes need to be guarded against invalid values much better `eg. /report/<unknown_id>`. So this could be an unnecessary optimization for now and punted to a separate route validation and rerouting task if you agree on that?

howard-e (approved):
Looks good! Thanks for addressing my previous feedback. Left a small inline comment for your thoughts but will leave merging to you if you see fit

client/hooks/useUrlTestIndex.js
stalgiag: This conditional isn't for that purpose but is instead for handling situations where the page is loading and the number of tests is not known. The return statement below should clamp the actual number to zero-to-max. I wasn't sure where to draw the bounds on handling URL mistakes since as you said we have a lot of ways of generating console errors when an incorrect URL is entered. Some of these console errors are likely useful but you are right that this one probably doesn't provide value and if for whatever reason we end up with UI that allows navigation to negative numbers there is no harm in handling that now. I put a `try, catch` around the `querySelector` call in the `ScrollFixer`. This should help with a broad range of incorrect URI fragments.

howard-e: Got it!

howard-e (approved):
Looks good! Thanks for addressing my *many* points

PR #1189, opened 2024-08-01
Reusable proptypes

stalgiag:
This PR introduces reusable PropTypes. The proptypes are largely a mirror of the existing GraphQL query fragments. A few additional proptypes were added for situations with relatively complex `PropType.shape`s.

To ensure that the PropTypes are honored and that no additional console errors are introduced by changes, I added a console error checker to the first load of pages during e2e testing.


client/tests/e2e/util/index.js
```diff
@@ -11,7 +11,31 @@ const display = async (page, selector, { wait = true } = {}) => {
...
 };
 
+/**
+ * Checks for errors on a given page
+ * @param {Page} page - Puppeteer page object
+ * @param {Function} action - Async function to perform actions that might trigger PropType errors
+ * @param {number} [timeout=500] - Timeout in milliseconds to wait for errors
+ * @returns {Promise<string[]>} - Array of error messages, if any
+ */
+async function checkConsoleErrors(page, action, timeout = 500) {
```
howard-e, line 21: This is super useful! This definitely gives a lot of confidence!

client/tests/e2e/DataManagement.e2e.test.js
```diff
@@ -12,10 +12,15 @@ describe('Data Management common traits', () => {
   it('renders page with introduction', async () => {
     await getPage({ role: false, url: '/data-management' }, async page => {
       const introductionHeadingSelector = 'h2 ::-p-text(Introduction)';
-      const introductionHeadingText = await text(
-        page,
-        introductionHeadingSelector
-      );
+      const errors = await checkConsoleErrors(page, async () => {
```
howard-e, line 15: This is great but I think it has a chance to miss errors on certain element checks or with how the page element is being loaded. In this instance, errors could happen after the "Introduction" text has loaded which is what I'm guessing is happening in the following screenshot. There is a prop error not being caught on the DataManagement page:

screenshot showing error with prop being passed to DataManagementRow

I'm thinking the `page.on('console')` listener could be done in `getPage` and passed into the callback the same way the browser and baseUrl and checked in that way

client/components/common/proptypes/index.js
```diff
@@ -0,0 +1,322 @@
+import PropTypes from 'prop-types';
```
howard-e, line 1: Wow, huge plus for putting this together and making all those changes. No more having to figure out how to account for certain structures :).

Also appreciate the care taken for identifying what should be required per prop type.

howard-e (commented):
This is great and certainly a plus for how we create and hydrate components!

I've left an additional comment on the new console check utility.

stalgiag:
Thanks for the suggestion on how to handle the error tracking in e2e. It worked like a charm and is much cleaner!

Interestingly it also uncovered an error that comes up on the reports page, "Failed to load resource: the server responded with a status of 404 (Not Found)". This only happens on CI. This likely existed previously but wasn't being checked. I can try to figure out what resource it is referring to tomorrow.

stalgiag:
> Thanks for the suggestion on how to handle the error tracking in e2e. It worked like a charm and is much cleaner!
> 
> Interestingly it also uncovered an error that comes up on the reports page, "Failed to load resource: the server responded with a status of 404 (Not Found)". This only happens on CI. This likely existed previously but wasn't being checked. I can try to figure out what resource it is referring to tomorrow.

I ended up adding a tag for a placeholder favicon so that the browser doesn't try to find a favicon by default. Used [this solution](https://stackoverflow.com/questions/31075893/how-to-fix-favicon-ico-error-failed-to-load-resource-neterr-empty-response/40527353#40527353).

client/static/index.html
```diff
@@ -5,6 +5,7 @@
     <meta name="google" content="notranslate">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="stylesheet" href="/index.css">
+    <link rel="shortcut icon" href="data:image/x-icon;," type="image/x-icon" />
```
howard-e, line 8: Interesting find!

howard-e (approved):
Super excited about the increased organization and deduplication something like this change brings. Thanks for putting this together and addressing my feedback!

howard-e:
@stalgiag there's a new conflict here after merging #1188. Feel free to update and merge once resolved

PR #1190, opened 2024-08-05
Address sandbox deployment issue

howard-e:
The current sandbox environment doesn't user the root user to make permission required changes which then prevents it from deploying as expected. This PR accounts for that in `/application/tasks/main.yml`.

A separate issue is there appears to be a package registry issue on the current sandbox when installing npm (but it does already exist). The change to `/nodejs/tasks/main.yml` is a suggested workaround.


stalgiag (approved):
Looks good to me!

Is the `npm` install even needed or is the version that is installed with the `apt-get install nodejs` enough?

howard-e:
> Looks good to me!
> 
> Is the `npm` install even needed or is the version that is installed with the `apt-get install nodejs` enough?

Looking back at this, it really should have. Let me take another look at what has messed with the configuration here

howard-e:
> Looks good to me!
> 
> Is the `npm` install even needed or is the version that is installed with the `apt-get install nodejs` enough?

@stalgiag you were correct in this. The sandbox had outdated package registry links which would cause issues installing npm during the node package install. This should be resolved now and so I've removed the need in https://github.com/w3c/aria-at-app/pull/1190/commits/5223bbfb2b9c04aff84fd91e1b4a92ee61205fa9

stalgiag (approved):
Nice! Glad there was an opportunity to simplify. Thanks for looking into it!

PR #1191, opened 2024-08-05
Move date related functions to @shared/dates

howard-e:
This is a follow-up from my comment at https://github.com/w3c/aria-at-app/pull/1177#pullrequestreview-2205228277. This should make updating and testing these related functions easier in the future.


stalgiag (approved):
Making this shared and consistent is very welcome. Thank you!

PR #1194, opened 2024-08-15
Update testing snapshot cleaner to account for changing Run History date

howard-e:
Based on the [latest build's push to `development`](https://github.com/w3c/aria-at-app/actions/runs/10404400325/job/28812810708), the Candidate Test Run snapshot will fail on every new date when the populate script is ran, because of how the Run History is evaluated.

Here is an attempt at updating the cleaner to account for this dynamism.


stalgiag (approved):
Thank you! I didn't quite understand that one. I wasn't sure whether it was updating every day or if it was upstream changes. Good to know it is changing every day and can be filtered out. Thanks for handling it!

Issue #1176, opened 2024-07-25
AssignTester dropdown reading null


ccanash:
When the user focuses on the AssignTesters dropdown and hit any key they get the following: :
- Cannot read properties of null (reading 'children')
- TypeError: Cannot read properties of null (reading 'children')

PR #1196, opened 2024-08-19
fix: container null when using keyboard to navigate out of "Assign Testers" dropdown

stalgiag:
see title. 

Bug was caused by the Tab key passing the Alphanumeric Regexp check.

addresses #1176


howard-e (approved):
Good catch!

PR #1195, opened 2024-08-16
feat: Enhanced conflict review

stalgiag:
This PR adds a dedicated page for reviewing conflicts between test plan runs in a test plan report.

addresses #975 

The design and layout is a first attempt and will likely need a few rounds of fine-tuning.

<details>
  <summary>Screenshot of Conflicts page with disclosures closed</summary>
A screenshot of a new page titled 'Conflicts for Test Plan Report'. Below the title, a section with h2 titled "Introduction" says "This page displays conflicts identified in the current test plan report. Conflicts occur when different testers report different outcomes for the same test assertions or unexpected behaviors" . Another section titled 'Test Plan Report' provides details about the specific test plan, including the version 'Select Only Combobox Example V22.03.17,' the assistive technology used 'NVDA (2020.4 and above),' and the browser 'Firefox.' Further down, the 'Conflicts' section indicates that there are three conflicts in the test plan report. These conflicts are listed in collapsible disclosures with titles: 'Test 4: Navigate backwards to a collapsed select-only combobox in interaction mode.', 'Test 6: Read information about a collapsed select-only combobox in interaction mode.', 'Test 7: Open a collapsed select-only combobox in reading mode.'

  The screenshot described above

</details>

<details>
  <summary>Expanded View of Specific Conflict</summary>
A screenshot shows an expanded conflict in the "Conflicts" section. The conflict is for "Test 4: Navigate backwards to a collapsed select-only combobox in interaction mode." The table compares assertions like the role, name, and state of the combobox, showing the results from two testers. One tester passed all assertions, while the other failed the first assertion but passed the rest. The first column is the assertion description, the second two columns are for each of the two testers. The column with different assertion results has strong text. Below the table, related GitHub issues are linked with details about the Issue title, whether it is currently open, and who opened it. Below that there are two buttons one to raise a new issue and one to open the run as a specific tester.

  The screenshot described above

</details>

PR #1197, opened 2024-08-21
Enhanced conflict review [extended]

howard-e:
@stalgiag while doing additional checks on #1195, noticed a few usability concerns that I've addressed in this PR that I had originally missed:

* Include versionString in issue body and title when creating from `/conflicts`
* The `requiredAtVersionName` variable was incorrect. Changed to using `exactAtVersionName`
* Updated the disclosures to be unique by commands per test, otherwise the same conflict was being duplicated if tester1 and tester2 had different results for assertion1 and assertion2 in the same test (each difference counts as conflict which was being mapped over in the view)
* Added support for the markdown conflict being included from the Raise an Issue for Conflict button

### DIFF
--- a/client/components/TestQueue/AssignTesters.jsx
+++ b/client/components/TestQueue/AssignTesters.jsx
@@ -94,7 +94,7 @@ const AssignTesters = ({ me, testers, testPlanReport }) => {
 
   const onKeyDown = event => {
     const { key } = event;
-    if (key.match(/[0-9a-zA-Z]/)) {
+    if (key.length === 1 && key.match(/[a-zA-Z0-9]/)) {
       const container = event.target.closest('[role=menu]');
       const matchingMenuItem = Array.from(container.children).find(menuItem => {
         return menuItem.innerText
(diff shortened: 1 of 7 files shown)


howard-e:
> * Updated the disclosures to be unique by assertions per test, otherwise the same conflict was being duplicated if tester1 and tester2 had different results for assertion1 and assertion2 in the same test (each difference counts as conflict which was being mapped over in the view)

Thinking on this some more, to retain current behavior from the Test Run page, this should probably be unique to just the test

stalgiag (approved):
This all makes sense to me! Thanks for adding it. This is a case of me forgetting to check with a production dump and relying on the limited cases in the default DB.

I am curious if now that we are heavily suggesting the use of a generated template for the issue, we should only enable a 'Raise an Issue' button when there isn't an existing issue. This can be a different discussion and isn't an important point but it feels a bit odd to me to suggest raising an issue that will be largely identical to an existing one.

howard-e:
> I am curious if now that we are heavily suggesting the use of a generated template for the issue, we should only enable a 'Raise an Issue' button when there isn't an existing issue. This can be a different discussion and isn't an important point but it feels a bit odd to me to suggest raising an issue that will be largely identical to an existing one.

Very good point there! We can discuss a bit further and certainly a point to raise to others who may be reviewing.

Issue #1202, opened 2024-08-27
Modify message for Vendors when they mark as Not approved a test plan


ccanash:
The new message should read: Thank you for reviewing, if you haven’t opened any issues yet, please open an issue describing why this is not approved.

PR #1206, opened 2024-08-28
Modify vendor review message

stalgiag:
see #1202


client/stories/ThankYouModal.stories.jsx
stalgiag, line 1: I deleted this while renaming because it was just a stub without meaningful customization and we aren't using stories as far as I understand.

client/components/CandidateReview/CandidateTestPlanRun/index.jsx
```diff
@@ -654,21 +673,8 @@ const CandidateTestPlanRun = () => {
...
-          handleAction={async () => {
-            setThankYouModalShowing(false);
-            navigate('/candidate-review');
-          }}
-          githubUrl={generalFeedbackUrl}
-        />
-      ) : (
-        <></>
       )}
+      {confirmationModal && confirmationModal}
```
stalgiag, line 677: This is technically unnecessary but I thought it helped convey the possible of a no render for the modal.

client/stories/ThankYouModal.stories.jsx
howard-e: That seems fine.

Part of me has felt keeping the storybook setup to be unnecessary. Historically, it hasn't gotten much traffic during development time and that's also reflected currently with only 2 storybook components being defined. Some only being created after the fact, so lacking for prototyping and certainly not much created after component creation so not much done in the way of functionalliy documenting components' usages. I'm all for the concept but not when there isn't any strong support for it -- right now at least.

Another thing is this project's versions are severely outdated and should probably get some updates!

client/components/CandidateReview/CandidateTestPlanRun/index.jsx
howard-e: Yea, thinking you don't have to make conveying this clear but `{!!confirmationModal && confirmationModal}` could also display that?

client/components/CandidateReview/CandidateModals/NotApprovedModal/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+      actions={[
+        {
+          label: 'Close',
+          onClick: handleAction
+        }
+      ]}
+      title={
+        <div className="review-confirmation-title">
+          <FontAwesomeIcon
+            icon={faCheck}
```
howard-e, line 35: Thinking an orange info circle or so could quickly reflect that this is the "non-final" path. I still thought I had incorrectly selected "Approved" instead of "Not Approved" at the end.

Just a thought, what do you think?

howard-e (approved):
Looks good! Left a small inline comment on a design thought but nothing blocking here, thanks!

client/components/CandidateReview/CandidateModals/NotApprovedModal/index.jsx
stalgiag: The green check felt odd to me too but I wasn't confident enough about what sentiment we wanted to convey about 'not approved' reviews. Now that I know it feels odd to you as well, I'll swap it out.

stalgiag: This was challenging. I decided that the exclamation triangle felt most accurate after trying a few others including the info circle. That change is [in this commit](https://github.com/w3c/aria-at-app/pull/1206/commits/201e8208153675290b83c85b127c081d04281990) but I don't feel attached to that decision. That particular "i" just felt off for a modal that didn't have information. I felt that the exclamation mark conveyed the need for attention without being too severe since it is orange. Another option is a pause to indicate the process is on hold or waiting further approval. Here are some screenshots. Do you have a preference?

modal with orange info i in circle
modal with orange triangle exclamation
modal with info i with no circle

howard-e: I'm also not strongly tied to either (except the 'i' with no background). Moving forward with the triangle seems fine to me! But thanks so much for the thought you put into that

howard-e (approved):

Issue #1204, opened 2024-08-27
Allow current results to be publicly viewed


howard-e:
A main purpose of this project is to easily present results and relevant data to visitors on how assistive technologies behave today. This use case is served by the [reports page](https://aria-at.w3.org/reports) which displays "finalized" results. These results have been fully reviewed and then presented following the test result collection process. This page is available to all users regardless of their "logged in" status.

However, if one were curious of how the current results being collected on the Test Queue looks for individual testers, only the admin can view that. Another tester can only view their runs, and only have an idea of another's if they force a conflict between their results. A non signed in user can view the Test Queue, the testers that are assigned and also the number of tests they have completed but they do not have the option to view the testers' results.

The came up during a recent discussion with @mcking65 while discussing if conflicts should be visible to all visitors in #1195. The current proposal is to do the following:

1. Signed out user:
	1. In the actions column on the Test Queue page, if there are testers assigned, add a "View Results for ..." button which opens a menu of the assigned testers (this is similar to the admin's "Open run as ..." functionality).
	2. Selecting an option open's the tester's run with all their results populated on each test's form page, in a read-only mode.
2. Signed in user (not admin):
	1. In the actions column on the Test Queue page, if there are testers assigned which are not the currently signed in user, add a "View Results for ..." button which opens a menu of the assigned testers (this is similar to the admin's "Open run as ..." functionality).
	2. Selecting an option open's the tester's run with all their results populated on each test's form page, in a read-only mode.

PR #1209, opened 2024-09-10
Allows testers' results to be publicly viewable

howard-e:
Address #1204.


client/components/TestRenderer/UnexpectedBehaviorsFieldset/index.jsx
```diff
@@ -84,7 +88,10 @@ const UnexpectedBehaviorsFieldset = ({
           name={`problem-${commandIndex}`}
           autoFocus={isSubmitted && unexpectedBehaviors.failChoice.focus}
           defaultChecked={unexpectedBehaviors.failChoice.checked}
-          onClick={unexpectedBehaviors.failChoice.click}
+          onClick={e => {
+            if (readOnly) e.preventDefault();
```
stalgiag, line 92: It might be nice to wrap this behavior in a handler function

client/components/TestRun/index.jsx
```diff
@@ -993,9 +1009,7 @@ const TestRun = () => {
       </div>
     );
 
-    // we are ready enough to show the page and all the buttons when the above code is
-    // pageReady and we have an anon view, bot test run, or a test result to display
-    const completeRender = pageReady && (!isSignedIn || currentTest.testResult);
+    // const enableInteractions = isSignedIn || !isViewingRun;
```
stalgiag, line 1012: Left behind?

stalgiag (commented):
This is awesome!

I tested manually and tried to see if I could break it. I was unable to. It does feel a bit weird to get the form view when viewing an unsubmitted test. I get that it isn't actually editable but just feels strange. Maybe it would feel better if the icon on the left in the navigator didn't switch to the "in progress" icon. No strong feelings here though.

I left minor organizational comments inline.

client/components/TestRun/index.jsx
howard-e: Good catch!

client/components/TestQueue/Actions.jsx
```diff
@@ -255,7 +255,29 @@ const Actions = ({
...
+          </Dropdown>
+        )}
+        {!isAdmin && (
+          <Dropdown focusFirstItemOnShow>
+            <Dropdown.Toggle
+              variant="secondary"
+              disabled={!nonSelfAssignedRuns.length}
+            >
+              <FontAwesomeIcon icon={faFileImport} />
+              View Results for...
```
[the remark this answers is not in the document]
howard-e: Another good catch! Done in https://github.com/w3c/aria-at-app/pull/1209/commits/643c123ce4abf3908215010abef78eed8776605c

howard-e:
> It does feel a bit weird to get the form view when viewing an unsubmitted test. I get that it isn't actually editable but just feels strange. Maybe it would feel better if the icon on the left in the navigator didn't switch to the "in progress" icon

@stalgiag agreed with this, it does seem a little odd. Following this is a screenshot I've done to represent the case where there is "missing" results for that tester's test in the test navigator. Also in the screenshot, is a highlighted text to show some description around the current state of the test if viewing as a user that's not admin / signed out (although I'm more weary about adding new language). I was thinking it could be something like:
* Viewing "submitted | in progress | unopened(?)" tests of tester in read-only mode ...

What do you think?

screenshot showing what's described above

stalgiag:
> @stalgiag agreed with this ...
> 
> What do you think?
> 

That looks great! I think 'unopened' works well enough.

howard-e:
> > @stalgiag agreed with this ...
> > What do you think?
> 
> That looks great! I think 'unopened' works well enough.

Thanks for your thoughts there! I've pushed https://github.com/w3c/aria-at-app/pull/1209/commits/38ef9529350e75e17fc8dddbc5bab0f692abcec7 for you to check out.

stalgiag (approved):
Looks great! Thanks for going above and beyond in addressing the feedback.

PR #1212, opened 2024-09-12
Address incorrect aria-at harness path being updated during API updates on deployment environments

howard-e:
Fixes issue where the import script API (and cron job) would put the updated harness files into `<PROJECT_ROOT>/../client/resources` instead of `PROJECT_ROOT/client/resources`. The resources were properly placed during a deployment, and remain unaffected.

Also had to make adjustments to support the resources being updated on the current sandbox environment.


stalgiag (approved):
LGTM!

PR #1213, opened 2024-09-17
Allow public viewing for Conflicts Page

howard-e:
After additional discussions with the CG, this should be supported and I was wrong in https://github.com/w3c/aria-at-app/pull/1195#discussion_r1722498164 so this change mainly reverts that commit.

Also noticed a small bug with the indexing not being done when opening a test from the Conflicts page.


stalgiag (approved):
Thanks for handling this!

PR #1214, opened 2024-09-18
fix: Add tests for retry cancelled collection jobs

gnarf:


howard-e (approved):
LGTM! Confirmed by testing manually on my local environment. Also reviewed and ran the new tests locally and all seems fine to me.

Issue #1226, opened 2024-09-27
Bot Run Assignment Failed


IsaDC:
#### Steps to Reproduce:
1. Navigate to the Test Queue and select the "Navigation Menu" button (V24.09.19).
2. For either NVDA or VoiceOver, press the "Assign Testers" button and select the bot.

#### Expected Behavior:
The bot should be assigned to the selected test run.

#### Actual Behavior:
After the "Assign Testers" dialog is displayed, the bot is not assigned to any test run.

#### Alternate Procedure:
Tried adding the test plan from the "Data Management" page and selecting "Add Bot Run" for both NVDA and VoiceOver. This method also fails.

#### Additional Information:
Both bots run successfully in the sandbox environment, indicating that the issue seems specific to the test queue.

#### Suggested Solution:
Fixing this issue as soon as possible would be greatly appreciated, as it significantly speeds up the testing process.

IsaDC:
### Update:
The bots are now being assigned correctly; however, the test run is immediately canceled after assignment.

howard-e:
This should now be addressed in the [latest release](https://github.com/w3c/aria-at-app/blob/releases/CHANGELOG.md#190-2024-10-09)

cc @mcking65 @IsaDC @ccanash

PR #1208, opened 2024-09-04
Vendor verification

stalgiag:
see #1201 

More info soon

PR #1227, opened 2024-10-01
Additional checks for AtVersion automation run support

howard-e:
Address #1226.

This PR explicitly checks for the AtVersion automation run support in `server/util/getAtVersionWithRequirements` with the already defined valid AtVersion constants instead of using the `AtVersion.supportedByAutomation` graphql attribute.

This is a quick patch due to the `supportedByAutomation` gql attribute not being returned as expected in this instance, while investigations into what may have changed continues.

Confirmed tests with this on the sandbox environment.


server/resources/ats.json
```diff
@@ -55,10 +28,40 @@
...
+                    "If the virtual cursor is active, press &lt;kbd&gt;Insert&lt;/kbd&gt;+&lt;kbd&gt;z&lt;/kbd&gt; to disable the virtual cursor."
+                ]
+            }
+        }
+    },
     {
         "id": 3,
         "name": "VoiceOver for macOS",
         "key": "voiceover_macos",
+        "vendorId": 4,
```
jugglinmike, line 64: Why do we need to define the `vendorId` property for this patch? Would you mind if I moved JAWS back in order to make the diff clearer?

howard-e: Odd, just realized this happened. Looks to be a change included by #1208 that wasn't pushed. That can definitely be done separately

howard-e: Removed that change

jugglinmike: Awesome! I'll merge once CI passes

jugglinmike (approved):

Issue #1228, opened 2024-10-03
Submit failure with Navigation Menu Button V24.09.19


howard-e:
Posted by Dean Hamack on October 2, 2024 to the public-aria-at@w3.org mailing list:

> I started doing testing on the Navigation Menu Button, but for some reason it says “0 of 11 test completed”, even though I’ve completed 2 of them. Curious if [others are] having the same issue.

PR #1229, opened 2024-10-03
fix: Assertion with EXCLUDE priority causing test result submit errors

howard-e:
Address #1228.

This seems to be a special case that happens when the assertion is excluded in tests the tests.csv and then included in *-commands.csv's `assertionExceptions` with a non-excluding priority. For example:
* [Navigation Menu Button's `navForwardsToMenuButton` test](https://github.com/w3c/aria-at/blob/bab79e2321ae203e8cfdd464871809247fddd3c3/tests/menu-button-navigation/data/tests.csv) which excludes `interactionModeEnabled` by default and any of [it's `*-commands.csv` which then includes `interactionModeEnabled` using `assertionExceptions`](https://github.com/w3c/aria-at/blob/bab79e2321ae203e8cfdd464871809247fddd3c3/tests/menu-button-navigation/data/nvda-commands.csv).
* Do note an example of this exclusion pattern has already been been pushed through successfully with the latest Color Viewer Slider. Based on my checking, the issue started to happen when [v1.7.0](https://github.com/w3c/aria-at-app/blob/development/CHANGELOG.md#170-2024-08-26) was released on August 26, 2024 and the reports for Color Viewer Slider were completed by the latest on [July 25, 2024](https://aria-at.w3.org/report/163615).
* I've also confirmed this doesn't affect the other method of assertion exclusion where the `*-commands.csv` excludes an assertion through the `assertionExceptions` instead, like in the case of [modal-dialog](https://github.com/w3c/aria-at/blob/c1a3b08d12454779f053cae6e7df98ec7f6d011f/tests/modal-dialog/data/nvda-commands.csv).

### DIFF
--- a/client/tests/e2e/TestRun.e2e.test.js
+++ b/client/tests/e2e/TestRun.e2e.test.js
@@ -108,22 +108,24 @@ describe('Test Run when not signed in', () => {
 });
 
 describe('Test Run when signed in as tester', () => {
-  const assignSelfAndNavigateToRun = async page => {
-    const modalDialogSectionButtonSelector =
-      'button#disclosure-btn-modal-dialog-0';
-    const modalDialogTableSelector =
-      'table[aria-label="Reports for Modal Dialog Example V24.06.07 in draft phase"]';
+  const assignSelfAndNavigateToRun = async (
+    page,
+    {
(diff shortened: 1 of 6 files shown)


jugglinmike (approved):
I ran the `development` branch locally, confirmed the erroneous behavior, ran this patch locally, and verified the fix. Nice going, @howard-e!

PR #1230, opened 2024-10-03
Update `server/resources/ats.json` following #1208 merge

howard-e:
Update `server/resources/ats.json` which now has different attributes after #1208 merge. This change would happen locally when running:

```bash
yarn db-import-tests:dev;
```


stalgiag (approved):
Looks good to me!

PR #1237, opened 2024-10-09
fix: Only include CollectionJob metrics in Bot Status calculation on Test Queue page when `tester.isBot` is true

howard-e:
This PR is to address the case in which a bot run is assigned to a manual user to be evaluated, and then a bot is assigned a run on the same report again. It would cause an aggregation of the assigned bot's and the previously assigned tester's results which would create confusing metrics in the Bot Status calculation, on the Test Queue page.

### DIFF
--- a/client/components/BotRunTestStatusList/index.js
+++ b/client/components/BotRunTestStatusList/index.js
@@ -60,8 +60,11 @@ const BotRunTestStatusList = ({ testPlanReportId }) => {
     };
     let anyPossibleUpdates = false;
     if (testPlanRunsQueryResult?.testPlanRuns) {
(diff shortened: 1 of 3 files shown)


jugglinmike (approved):
Hey @howard-e, this looks good to me, and it works on my machine. I recognize that we're not convinced that this particular situation (a human user with a `collectionJob` attribute) ought to be allowed. [I added a test anyway](https://github.com/w3c/aria-at-app/pull/1237/commits/b62d38b12e97fbd6502851db047eb862579fd8cf) because it strengthens the evidence that your patch has the intended effect. Because that makes me the last one to touch this branch, though, I'm going to leave merging to you.

jugglinmike:
Oh, nice, thanks!

Issue #1260, opened 2024-10-23
Toggle button report data was not preserved when advancing v2024.10.22 to draft


mcking65:
I just advanced V2024.10.22 of toggle button to draft and the report status is that none of the required reports are started. None of the prior report data was copied in. I was expecting all of it.

This is in contrast with v2024.10.22 of the radio group test plan. Just before advancing the toggle plan, I advanced the radio plan to draft review, and all the data that should have been preserved was preserved. There were 4 tests for NVDA where the changes caused data to be dropped, and that was expected. the JAWS and VoiceOver reports for radio group were fully preserved.

howard-e:
> This is in contrast with v2024.10.22 of the radio group test plan. Just before advancing the toggle plan, I advanced the radio plan to draft review, and all the data that should have been preserved was preserved. There were 4 tests for NVDA where the changes caused data to be dropped, and that was expected. the JAWS and VoiceOver reports for radio group were fully preserved.

The difference between the 2 is that the `refIds` were updated for toggle button but the change wasn't being ignored during the results copy process. #1261 should address it.

When this lands production, the results can still be immediately copied by attempting to create the test plan reports matching the combinations from the old toggle button test plan version. A modal will pop up asking if you'd like to create an empty report or copy over the old results.

mcking65:
Thank you @howard-e for quickly deploying a fix!! I successfully advanced the latest version of the toggle button plan into candidate review.

PR #1261, opened 2024-10-24
fix: Ignore `renderableContent.info.references` changes when doing test results copy

howard-e:
Address #1260.


stalgiag (approved):
Makes sense. Thanks Howard!

PR #1262, opened 2024-10-24
docs: Update `database.md` to clarify how live environments database dumps should be used locally

howard-e:


stalgiag (approved):
Small but helpful changes. Thanks!

Issue #1248, opened 2024-10-21
Do not count unsupported assertions as failures in report headings


mcking65:
Vispero has provided the feedback that we are incorrectly calling out failures that are not present.

When a test plan contains assertions with priority 3 (optional), the unsupported priority 3 assertions should not be counted as failures in the headings.

Change the heading text when optional assertions are present to count the number of unsupported assertions separately. To keep the heading text as brief as possible, The supported optional assertions can continue to be counted as passes in the heading text.

Example:  In the [report for JAWS and Chrome for Radio Group Example Using aria-activedescendant](https://aria-at.w3.org/report/75322/targets/242), look at test 5.

The heading text is:
"Test 5: Navigate out of the start of a radio group (17 passed, 3 failed)"
![screenshot](https://github.com/user-attachments/assets/33d35e4e-c205-4614-86f7-2b15882782ee)

Change the heading text to:
"Test 5: Navigate out of the start of a radio group (17 passed, 0 failed, 3 unsupported)"

Ensure the change covers:
* Test headings
* command headings
* on report pages
*On candidate review pages

PR #1265, opened 2024-10-28
fix: Exclude unsupported assertion results from failure count in report headings; use separate "unsupported" metric

howard-e:
Address #1248.


stalgiag (approved):
This fits the bill. Thanks!

PR #1267, opened 2024-11-04
feat: Track vendor approvals across version updates through result copy process

howard-e:
Address #1232


howard-e:
@stalgiag this should be ready again for review after updating the tests

stalgiag (approved):
This is great. Thanks so much for the attention to detail in updating async and comparison code where you encountered it, adding in some utility testing methods, and writing an integration test that covers this feature. It was challenging to recreate the behavior manually but I managed with your instructions. Everything works as expected and the vendor approval status was carried over to the newer candidate version about promoting it. Great work!

PR #1279, opened 2024-11-20
fix: Allow AT and Browser versions to be edited when bot run is assigned to manual tester for evaluation

howard-e:
Address https://github.com/w3c/aria-at/issues/1163


stalgiag (approved):
Looks good to me! Thanks for the quick update Howard!

Issue #1241, opened 2024-10-10
Improve the experience of GH issue creation across the app.


ccanash:
Define how to improve the experience of GH issue creation across the app. Some examples of improvement can include: 

1. Add the possibility to create a GH issue from additional pages in the app. For example: about a report, about 1 test, about a test plan as a whole,  about a specific command within a test)
2. Improve how to display existing issues. It currently it mixes open and closed issues. 
3. Explore having existing issues appear elsewhere than → in the same places where we are adding the possibility to submit an issue for example in candidate review (have a link so that user know that there are existing issues and can check what they are)

howard-e:
> 1. Add the possibility to create a GH issue from additional pages in the app. For example: about a report, about 1 test, about a test plan as a whole,  about a specific command within a test)

We can re-use the “Raise an Issue” button to instead present a secondary set of radio options in a dialog (or similar) depending on the context:
* On Test Run:
    1. Raise an Issue -> Select “… about the test” (current default)
    2. Raise an Issue -> Select “… about a specific command” (selecting this now shows another list for which command (or commands) are problematic
    3. Raise an Issue -> Select “… about the test plan”
    4. In addition to no.3, the Test Plan info shown at the top of the Test Run should also show and link to the correct test plan version’s review page and there should also be a “Raise an Issue” button on that page

* On Test Review Page:
    1. Include “Raise an Issue” button under GitHub Issues table and given the context, that covers the entire Test Plan (this will require a template addition/change to highlight this is for the entire test plan)

* On Test Plan’s Report Page:
    1. Include “Raise an Issue” button under the Metadata section. We can allow the user to select a specific combination of the listed AT + Browser reports that are present, otherwise we can assume it covers the entire report (this will require a template addition/change to highlight the issue is for a published report)

* On Test Plan’s AT + Browser Report Page:
    1. We can re-use the existing “Raise an Issue” buttons here in the same way they are being used on the Test Run page (this will require a template addition/change to highlight the issue is for a published report)

> 2. Improve how to display existing issues. It currently it mixes open and closed issues.

This would have been recently covered by #1266 and #1269.

> 3. Explore having existing issues appear elsewhere than → in the same places where we are adding the possibility to submit an issue for example in candidate review (have a link so that user know that there are existing issues and can check what they are)

Can do!

PR #1281, opened 2024-11-21
Include additional "Raise an Issue" links

howard-e:
Address #1241, particularly the comments in https://github.com/w3c/aria-at-app/issues/1241#issuecomment-2491778334.

### DIFF
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -39,7 +39,6 @@ import { evaluateAuth } from '../../utils/evaluateAuth';
 import './TestRun.css';
 import ReviewConflicts from '../ReviewConflicts';
 import createIssueLink from '../../utils/createIssueLink';
-import { dates } from 'shared';
 import { Provider as CollectionJobContextProvider } from './CollectionJobContext';
 import { useUrlTestIndex } from '../../hooks/useUrlTestIndex';
 
--- a/client/components/TestRun/index.jsx
+++ b/client/components/TestRun/index.jsx
@@ -366,10 +366,7 @@ const TestRun = () => {
     issueLink = createIssueLink({
       testPlanTitle: testPlanVersion.title,
       testPlanDirectory: testPlanVersion.testPlan.directory,
-      versionString: `V${dates.convertDateToString(
-        testPlanVersion.updatedAt,
-        'YY.MM.DD'
-      )}`,
+      versionString: testPlanVersion.versionString,
       testTitle: currentTest.title,
       testRowNumber: currentTest.rowNumber,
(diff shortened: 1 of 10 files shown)


client/components/TestRun/index.jsx
```diff
@@ -967,9 +985,62 @@ const TestRun = () => {
       <div role="complementary">
         <h2 id="test-options-heading">Test Options</h2>
         <ul className="options-wrapper" aria-labelledby="test-options-heading">
-          <li>
+          <li className="raise-issue-container">
```
stalgiag, line 988: I know this matches a definition in the issue but I can't help but feel like this would make for a challenging user experience, especially on mobile. I almost feel like we want to move this radio into a modal or have a separate link for each command in a future design. For now this works well.

stalgiag (approved):
Great work! I tested the TestRun and IssuesTable links. Both work well. The code is straightforward and easy to read. I left two small comments, one is a layout nit and the other is a musing on user flow. Neither are blocking. Thanks for this work! I am going to approve now and leave to you to merge.

client/components/TestRun/TestRun.css
```diff
@@ -523,3 +523,35 @@ a.btn-options:hover {
...
+.raise-issue-container .options.type {
+  justify-content: space-between;
+  margin-bottom: 0.5rem;
+}
+
+.raise-issue-container .options.command {
+  flex-direction: column;
+  align-items: flex-start;
+  text-align: left;
+  margin-bottom: 0.5rem;
```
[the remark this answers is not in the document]
howard-e: No issue with this. I increased by a somewhat small amount because I felt it shouldn't go _too_ far away from the related 'Raise an Issue' button. But that may also be my preference slipping in as well ha. Any thoughts on an exact value you may prefer here?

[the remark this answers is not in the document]
stalgiag: That works! Thanks for the compromise between our preferences haha.

client/components/TestRun/index.jsx
stalgiag: That is a good point. I do like the thought of keeping consistency in a single-click approach. The separate link might have some navigational advantages in that it would reduce the distance between the part of the interface with information about a command and the inputs to create the issue.

stalgiag (approved):
I definitely prefer the links on each test. I also think it will be easier to use. Thanks for trying out different options and following up on feedback even when it wasn't mandatory!

client/components/TestRun/index.jsx
stalgiag: Love it! Gave an approving review but generally find this easier to parse and navigate.

PR #1288, opened 2024-12-18
feat: Add report section summarizing failing assertions from all tests in a plan

stalgiag:
see #1253 

This PR adds a report table summarizing all failing assertions to the Candidate Review and on Report pages.

There were a number of design and user experience decisions made here that I especially welcome feedback on. There were also a few decisions for which I was on the fence. Namely:
- Whether to keep the FailingAssertionsSummaryTable in a Disclosure. Sometimes the table can be quite large.
- Whether the Candidate Review page should start on the summary page or if it should continue to use the first test as a starting view. I opted for the latter but there is value in both.


client/components/TestRun/TestNavigator.jsx
```diff
@@ -61,6 +61,25 @@ const TestNavigator = ({
           aria-labelledby="test-navigator-heading"
           className="test-navigator-list"
         >
+          {isVendor && (
```
howard-e, line 64: Should probably conditionally show the Summary page or provide some default content (maybe some mashup of results table from the report page?) when there are no failing assertions.

Otherwise the following screenshot happens, which is just a blank content area and the heading being "Summary of Failing Assertions (0 must, 0 should)":

screenshot showing blank content area with the heading Summary of Failing Assertions (0 must, 0 should)

My preference is to hide if none exists because while I like the utility of the new "Begin Review" button, it would make more sense if this was an overall summary page rather than a page that just says "No failures exist, begin your review!". The overall summary isn't the request here and it may happen in the future but not something I'd push to have included right now. At least not before we make proper component based changes. Do you have any thoughts on that?

client/components/CandidateReview/CandidateTestPlanRun/index.jsx
```diff
@@ -557,40 +616,55 @@ const CandidateTestPlanRun = () => {
...
+                      )}
+                      {isSummaryView ? (
+                        <li className="begin-review-button-container">
+                          <Button
+                            ref={nextButtonRef}
+                            variant="secondary"
+                            onClick={handleNextTestClick}
+                            disabled={isLastTest}
+                          >
+                            Begin Review
```
howard-e, line 653: Nice, I like the inclusion of clear entry points!

client/components/common/proptypes/index.js
```diff
@@ -208,6 +208,39 @@ export const TestPlanReportConflictPropType = PropTypes.shape({
   ).isRequired
 });
 
+export const TestPlanReportMetricsPropType = PropTypes.shape({
```
howard-e, line 211: Well needed!

howard-e (changes requested):
Thanks @stalgiag! Very impressive getting all this added and in a pretty maintainable way as well given where it has to be referenced.

The main issue I've come across is the handling of the priority 0 assertions but I left some inline suggestions on that. Seems this is all very close!

client/components/TestRun/TestNavigator.jsx
stalgiag: I agree. I think just leaving it out entirely is sensible. Will do!

client/hooks/useFailingAssertions.js
```diff
@@ -0,0 +1,35 @@
...
+
+    return testPlanReport.finalizedTestResults.flatMap(
+      (testResult, testIndex) => {
+        return testResult.scenarioResults.flatMap(scenarioResult => {
+          return (
+            scenarioResult.assertionResults
+              .filter(assertionResult => !assertionResult.passed)
+              // We only want to show MUST and SHOULD assertions
+              .filter(
+                assertionResult => assertionResult.assertion.priority !== 'MAY'
```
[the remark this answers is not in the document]
stalgiag: Thank you!

client/components/FailingAssertionsSummary/Table.jsx
```diff
@@ -0,0 +1,64 @@
...
+  const failingAssertions = useFailingAssertions(testPlanReport);
+  const { metrics } = testPlanReport;
+
+  if (failingAssertions.length === 0) return null;
+
+  return (
+    <>
+      <p>
+        {metrics.assertionsFailedCount} assertions failed for{' '}
+        {metrics.commandsCount} commands in {metrics.testsFailedCount} tests.
```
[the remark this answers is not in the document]
stalgiag: Yeah, I am unsure on this one. I did think of it the way you are suggesting but with the template of `X assertions failed for Y commands in Z tests`, I believe that X and Y will always have the same value. The way it is currently assumes that the reader's goal is to understand the ratio of failures to successes - `X failures of Y possible failures across Z total tests". After reading this do you still think I should take the `X assertions failed Y commands with failing assertions Z tests with commands that have failing assertions`? I am still on the fence.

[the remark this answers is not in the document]
howard-e: Ahh I now get the other side when reading through this. Was also on the fence, but going back through #1253 does make it feel like it is correct how it's currently coded. I can move on from this particular one

client/components/TestRun/TestNavigator.jsx
howard-e: Sounds good!

client/components/FailingAssertionsSummary/Table.jsx
```diff
@@ -0,0 +1,64 @@
...
+}) => {
+  const failingAssertions = useFailingAssertions(testPlanReport);
+  const { metrics } = testPlanReport;
+
+  if (failingAssertions.length === 0) return null;
+
+  return (
+    <>
+      <p>
+        {metrics.assertionsFailedCount} assertions failed for{' '}
```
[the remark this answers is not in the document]
stalgiag: I believe this is addressed by [79adb6d](https://github.com/w3c/aria-at-app/pull/1288/commits/79adb6d9e401f86320c2ae6e705674d9599b72dc)

client/components/TestRun/TestNavigator.jsx
stalgiag: I believe this was addressed in [043e55b](https://github.com/w3c/aria-at-app/pull/1288/commits/043e55bff2b5deb3733ec9095519d4a027ac910b)

client/components/FailingAssertionsSummary/Table.jsx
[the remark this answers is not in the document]
howard-e: > I believe this is addressed by [79adb6d](https://github.com/w3c/aria-at-app/pull/1288/commits/79adb6d9e401f86320c2ae6e705674d9599b72dc)

@stalgiag the underlying calculation is correct but the result of it isn't being used here. This suggestion should still be done because the `MAY` assertions from the overall `getMetrics` calculation is still being counted in it.

It makes it confusing when the result table only has 14 rows but the text notes `24 assertions failed for x commands in y tests`

client/components/TestRun/TestNavigator.jsx
howard-e: Looks good!

howard-e (commented):
@stalgiag the changes look good! I still have one pressing concern I left inline: https://github.com/w3c/aria-at-app/pull/1288#discussion_r1932242722

Looks good to go for me after.

Also, I'm realizing that my previous concern in our discussion from [this thread](https://github.com/w3c/aria-at-app/pull/1288#discussion_r1907855493) may just have to do with the structure of the sentence. `x assertions failed ACROSS y commands in z tests` _feels_ better to me. But once again, maybe it's just my understanding being slightly off. But something to think on if it's raised in the future

client/components/FailingAssertionsSummary/Table.jsx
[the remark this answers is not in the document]
stalgiag: Ah yes! Apologies. Brain fog must've been in charge that day. Thanks for sticking to your point 😅

stalgiag:
> @stalgiag the changes look good! I still have one pressing concern I left inline: [#1288 (comment)](https://github.com/w3c/aria-at-app/pull/1288#discussion_r1932242722)
> 
> Looks good to go for me after.
> 
> Also, I'm realizing that my previous concern in our discussion from [this thread](https://github.com/w3c/aria-at-app/pull/1288#discussion_r1907855493) may just have to do with the structure of the sentence. `x assertions failed ACROSS y commands in z tests` _feels_ better to me. But once again, maybe it's just my understanding being slightly off. But something to think on if it's raised in the future

I agree with "across" feeling better. I took the liberty of changing it 😄  [72ba042](https://github.com/w3c/aria-at-app/pull/1288/commits/72ba042a69104b61fdfe9d4b1da93700d3164b50)

client/components/FailingAssertionsSummary/Table.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+  if (failingAssertions.length === 0) return null;
+
+  return (
+    <>
+      <p>
+        {failingAssertions.length} assertions failed across{' '}
+        {metrics.mustAssertionsFailedCount +
+          metrics.shouldAssertionsFailedCount}{' '}
+        commands in {metrics.testsFailedCount} tests.
```
howard-e, line 26:
```suggestion
        {metrics.commandsCount} commands in {metrics.testsFailedCount} tests.
```

This should still be `metrics.commandsCount` based on our discussions. I definitely introduced some confusion here with this [suggestion](https://github.com/w3c/aria-at-app/pull/1288#discussion_r1907848679) so my apologies!

(The `metrics.mustAssertionsFailedCount + metrics.shouldAssertionsFailedCount` in the comment was meant to be an alternative to the `failingAssertions.length` if you preferred, to show how I was thinking of it)

howard-e (commented):
@stalgiag one more inline change because of some confusion I introduced! 😅

Should be good to go afterwards

howard-e (approved):
Changes look good and very exciting! Thanks for working through all the feedback items!

PR #1290, opened 2024-12-18
feat: Include Settings page for admin and UI to manually import latest test plan versions

howard-e:
Coming from a recent discussion with Matt + CG:

>  When we merge a plan to the main” branch, I know we have a “cron” job running–it would be nice if we had an admin button to just refresh the data on the data management page. Like “Initiate GitHub Import”.

This PR does the following:
* ~Adds an 'Import Latest Test Plan Versions' button to the Data Management page for an admin user~
* Allows navigation to the `/settings` page for the admin which will now also include the button mentioned above. It will also display the date of the latest imported test plan version

Note: Supersedes #1289 to avoid issues with commit history


client/components/DataManagement/index.jsx
```diff
@@ -94,143 +115,154 @@ const DataManagement = () => {
...
+          filterOptions={filterLabels}
+          activeFilter={filter}
+          onFilterChange={setFilter}
+        />
+
+        {isAdmin && (
+          <Button variant="primary" onClick={handleImportTests}>
+            Import Latest Test Plan Versions
+          </Button>
+        )}
```
stalgiag, line 188: I prefer only having this on the /settings page. If you disagree then perhaps the button can be somewhere else? I feel it is confusing to have it between the filter buttons and the table that they filter.

client/components/UserSettings/AdminSettings.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from 'react-bootstrap';
+import { LoadingStatus, useTriggerLoad } from '../common/LoadingStatus';
+import { useThemedModal, THEMES } from '../../hooks/useThemedModal';
+import { dates } from 'shared';
+
+const AdminSettings = ({ latestTestPlanVersion, refetch }) => {
```
stalgiag, line 8: Really love this use of the settings page.

client/components/UserSettings/AdminSettings.jsx
```diff
@@ -0,0 +1,79 @@
...
+        setShowThemedModal(true);
+      }
+    }, 'Importing latest Test Plan Versions');
+  };
+
+  return (
+    <LoadingStatus message={loadingMessage}>
+      <section>
+        <h2>Actions</h2>
+        <p>Use the following actions below to perform a certain function:</p>
```
stalgiag, line 54: This text feels somewhat superfluous as it is now. Perhaps it could be worthwhile to indicate that these are actions that can be performed by admins only?

client/components/UserSettings/AdminSettings.jsx
```diff
@@ -0,0 +1,79 @@
...
+    setThemedModalType
+  } = useThemedModal({
+    type: THEMES.SUCCESS,
+    title: 'Success!'
+  });
+
+  const handleImportTests = async () => {
+    await triggerLoad(async () => {
+      try {
+        const response = await fetch('/api/test/import', { method: 'POST' });
```
stalgiag, line 25: I know that it would probably be a relatively significant lift to get a progress bar on this  but it is difficult to see the infinite spinner with such a long running operation. My system took a few minutes to complete the task and I was concerned that something had broken. If the progress bar would be too much work then perhaps we could add a note that "this could take a few minutes"?

client/components/common/BasicThemedModal/index.jsx
```diff
@@ -80,7 +86,13 @@ const BasicThemedModal = ({
               <FontAwesomeIcon
                 icon={faExclamationTriangle}
                 size="lg"
-                color={theme === 'danger' ? '#ce1b4c' : '#fab700'}
+                color={
+                  theme === 'danger'
+                    ? '#ce1b4c'
+                    : theme === 'warning'
+                    ? '#fab700'
+                    : '#2ba51c' // success
```
stalgiag, line 94: Wonder if the nested ternary `getThemeColor()` could become a function since it is used twice? Would give an opportunity to use a switch or something a bit easier to read.

server/util/constants.js
```diff
@@ -6,7 +6,7 @@ const AT_VERSIONS_SUPPORTED_BY_COLLECTION_JOBS = {
   // https://github.com/bocoup/aria-at-gh-actions-helper/blob/main/.github/workflows/voiceover-test.yml#L39
   'VoiceOver for macOS': ['13.0', '14.0'],
   // These are tracked with the https://github.com/bocoup/aria-at-automation-nvda-builds/releases
-  NVDA: ['2024.4.1', '2024.1', '2023.3.3', '2023.3']
```
stalgiag, line 9: History issue?

stalgiag (changes requested):
This is exciting! Really love this use of the settings page. Left some comments inline.

client/components/DataManagement/index.jsx
howard-e: I'm not a big fan of it here either. I tried to plug it in mainly because of the original request being to have the button on the Data Management page but I feel it more fitting by itself on the Settings page. So removing here isn't an issue to me!

client/components/UserSettings/AdminSettings.jsx
howard-e: Seems fair!

server/util/constants.js
howard-e: Ahh perhaps. Good catch!

howard-e: https://github.com/w3c/aria-at-app/pull/1290/commits/6d906689da6e7c67b75cf75bb1a6685c27d26f91

client/components/UserSettings/AdminSettings.jsx
howard-e: Opted to add a note instead in https://github.com/w3c/aria-at-app/pull/1290/commits/a9dae3ab42039199b8be4906613f43df803f4012. Let me know if this feels better to you

client/components/UserSettings/AdminSettings.jsx
howard-e: Should be addressed in https://github.com/w3c/aria-at-app/pull/1290/commits/f5193e99d415fd6276f79e7d40c333522fabcc4f

client/components/UserSettings/AdminSettings.jsx
stalgiag: Yep this feels much better. Thanks!

client/components/common/BasicThemedModal/index.jsx
stalgiag: Nice!

stalgiag (approved):
This is great now! Thanks for addressing the feedback.

PR #1299, opened 2024-12-19
Maintain `aria-bot` sudo permission to run the `import-tests` script in deployed environments

howard-e:
This PR ensures a `/etc/sudoers.d/10_aria-bot` file exists to execute the import-tests script on deployed environments as a sudo user.

This should have no change on how the application currently runs as the same was already set up on the production, staging and sandbox environments separately. This is merely a maintenance task to make sure if the application is spun up again on another server that this file is automatically included.


stalgiag (approved):
Makes sense! Thanks for covering this base.

Issue #1270, opened 2024-11-06
Bot Run assignment issues with VoiceOver


howard-e:
Bot run oddities observed during [today's CG meeting](https://www.w3.org/2024/11/06-aria-at-minutes.html#9f76):

1. Admin @IsaDC assigned `VoiceOver Bot` to "Action Menu Button Example Using element.focus() - V24.10.31 for VoiceOver + Safari". The bot reported 1 error and 10 tests cancelled, for the 11 tests in the test plan. [[View logs on that failed run]](https://github.com/bocoup/aria-at-gh-actions-helper/actions/runs/11618336761).
2. Afterwards, the results of that bot run was assigned to Isa which expectedly had no results populated, to be evaluated.
3. Isa's run was then removed and a new run created. It was then immediately assigned to IsaDC to be evaluated. This run was successful but it was able to be assigned before the run was completed. That doesn't impact the results being populated, but this is a particularly confusing case when the expectation is that a bot's results can only be assigned and evaluated AFTER the bot run has been completed [[View logs on that successful run]](https://github.com/bocoup/aria-at-gh-actions-helper/actions/runs/11709768478/job/32614530609)

TODO:
1. Investigate why the VoiceOver bot failed (logs above in no.1).
2. Investigate @mcking65's report on why the 'Manage Bot Run' button didn't appear to get to the option to re-run the bot's results collection.
3. Prevent the bot's run from being assigned until it is fully completed.

ccanash:
Closing this issue as it addressed items 2 and 3.  Opened https://github.com/w3c/aria-at-app/issues/1305 to track item 1 (cc @IsaDC )

PR #1300, opened 2024-12-20
fix: prevent reassignment of bot run before status is completed, cancelled, or errored

stalgiag:
see title

partially addresses #1270 

We don't currently have a good way to write e2e tests for these delayed simulated bot run ui flows. I think this is out of scope for this particular task but is worth considering in the future.


howard-e (approved):
Looks good to me! Agreed this would benefit from some future tests but as it is right now, this is nothing to block on

Issue #759, opened 2023-08-30
Configure staging and sandbox to pull tests from a aria-at-data-sandbox repo instead of aria-at


mcking65:
We can't fully test the process in staging and sandbox because there is no way to:
1.  Test how the system responds to new test plan versions.
2. Test how the system responds to edge conditions where test plans are in every phase of the working mode.
3. Test creation of issues from the app UI.
4. Test tracking of issues by the app.

To conduct this kind of testing, we need an equivalent to the aria-at repo where we can change the test plan data and github issues without consequences to the production aria-at data.

howard-e:
> 1. Test how the system responds to new test plan versions.
> 2. Test how the system responds to edge conditions where test plans are in every phase of the working mode.

Agreed. The idea when mirroring the production environment in this instance was to make it easier for the differences to be understood between the prod environment, and to also check for any unexpected changes to existing behavior when using the same data. Ideally, the sandbox environment will always have sample data to support such testing like this in the future.

> 3. Test creation of issues from the app UI.

When being ran against the sandbox/staging environments, we can point this [Bocoup's aria-at fork](https://github.com/bocoup/aria-at/issues) for testing this feature.

> 4. Test tracking of issues by the app.

Same as above

howard-e:
> > 3. Test creation of issues from the app UI.
> 
> When being ran against the sandbox/staging environments, we can point this [Bocoup's aria-at fork](https://github.com/bocoup/aria-at/issues) for testing this feature.
> 
> > 4. Test tracking of issues by the app.
> 
> Same as above

When not being ran on production, `bocoup/aria-at/issues` will now be used as the source

mcking65:
Moving to the ops project

ccanash:
Closing as released on October 9th

ccanash:
Reopening to address additional elements

ccanash:
Most of this was addressed in https://github.com/w3c/aria-at-app/pull/1301 . What is pending here is to write documentation on how how to get a dump of the production database to restore it on staging.

PR #1301, opened 2025-01-08
Add database management API calls and update crons

howard-e:
Partially address #759 in making it easier to get a database dump from `production` and restoring it in `staging` or `sandbox`.

This PR adds API functions:
* Dumping postgres database
* Restoring postgres database
* Cleaning up the database dumps directory so it doesn't grow too large (defaulted to max of 28 days)

This PR adds the following cron jobs:
* Dumping postgres database every day at 00:00
* Cleaning up the dumps directory if necessary every day at 00:05

### DIFF
--- a/server/controllers/DatabaseController.js
+++ b/server/controllers/DatabaseController.js
@@ -1,0 +1,240 @@
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+const { Client } = require('pg');
+const { exec } = require('child_process');
+const { dates } = require('shared');
+
+// Database configuration
+const DB_NAME = process.env.PGDATABASE;
+const DB_USER = process.env.PGUSER;
+const DB_PASSWORD = process.env.PGPASSWORD;
+const DB_HOST = process.env.PGHOST;
+
+// To avoid unintentionally exposing environment variables in error messages
+const sanitizeError = message => {
+  const redacted = '#redacted#';
+
+  const dbNameRegex = new RegExp(DB_NAME, 'g');
+  const dbUserRegex = new RegExp(DB_USER, 'g');
+  const dbPasswordRegex = new RegExp(DB_PASSWORD, 'g');
+  const dbHostRegex = new RegExp(DB_HOST, 'g');
+
+  return message
+    .replace(dbNameRegex, redacted)
+    .replace(dbUserRegex, redacted)
+    .replace(dbPasswordRegex, redacted)
+    .replace(dbHostRegex, redacted);
+};
+
+const removeDatabaseTablesAndFunctions = async res => {
+  // Database connection configuration
+  const client = new Client({
+    user: DB_USER,
+    host: DB_HOST,
+    database: DB_NAME,
+    password: DB_PASSWORD
+  });
+
+  try {
+    // Init database connection
+    await client.connect();
+    res.write('Connected to the database.\n');
+
+    // [Tables DROP // START]
+    // Get all tables in the public schema
+    const tablesQuery = `
+      SELECT tablename
+      FROM pg_tables
+      WHERE schemaname = 'public';
+    `;
+    const tablesResult = await client.query(tablesQuery);
+
+    // Build and execute DROP TABLE commands
+    const tableNames = tablesResult.rows
+      .filter(row => row.tablename !== 'session')
+      .map(row => `"${row.tablename}"`);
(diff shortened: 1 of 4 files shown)


server/controllers/DatabaseController.js
```diff
@@ -0,0 +1,242 @@
...
+    res
+      .status(400)
+      .write(`Unable to import database: ${sanitizeError(error.message)}`);
+    return res.end();
+  }
+};
+
+// Function to clean up the dump folder
+const cleanFolder = (req, res) => {
+  const { dumpOutputDirectory, maxFiles } = req.body;
```
stalgiag, line 181: I think some safety checks are needed here. If I am not mistaken a 
```
{
  "dumpOutputDirectory": "../../../etc",
  "maxFiles": 1
}
``` 
could wreak havoc. I know that user permissions will probably prevent this but with deletion I like to have multiple safeguards.

Maybe this should be exclusively allowed with the default path? Alternatively, some path validation could be used if you want to allow a few different locations. I also think this route shouldn't be accessible on prod either.

server/controllers/DatabaseController.js
```diff
@@ -0,0 +1,242 @@
...
+      .send(`Unable to dump database: ${sanitizeError(error.message)}`);
+  }
+};
+
+const restorePostgresDatabase = async (req, res) => {
+  // Prevent unintentionally dropping or restoring database tables if ran on
+  // production unless manually done
+  if (process.env.ENVIRONMENT === 'production') {
+    return res.status(405).send('This request is not permitted');
+  }
```
stalgiag, line 144: I know, intellectually, that this is probably enough to prevent this being run on prod but the string matching as the only safeguard leaves me a bit uneasy

server/controllers/DatabaseController.js
```diff
@@ -0,0 +1,242 @@
...
+
+  const { pathToFile } = req.body;
+  if (!pathToFile)
+    return res.status(400).send("'pathToFile' is missing. Please provide.");
+
+  // Purge the database's tables and functions to make restoring with the
+  // pg import easier
+  await removeDatabaseTablesAndFunctions(res);
```
stalgiag, line 152: A dump for backup could be made first?

server/controllers/DatabaseController.js
```diff
@@ -0,0 +1,242 @@
...
+    const files = fs.readdirSync(OUTPUT_DIR).map(file => {
+      const filePath = path.join(OUTPUT_DIR, file);
+      return {
+        name: file,
+        path: filePath,
+        time: fs.statSync(filePath).mtime.getTime() // Get last modified time
+      };
+    });
+
+    // Sort files by modification time (oldest first)
```
stalgiag, line 209: Some of these comments feel unnecessary when the code is self-documenting or straightforward.

server/controllers/DatabaseController.js
```diff
@@ -0,0 +1,242 @@
...
+const { dates } = require('shared');
+
+// Database configuration
+const DB_NAME = process.env.PGDATABASE;
+const DB_USER = process.env.PGUSER;
+const DB_PASSWORD = process.env.PGPASSWORD;
+const DB_HOST = process.env.PGHOST;
+
+// To avoid unintentionally exposing environment variables in error messages
+const sanitizeError = message => {
```
stalgiag, line 15: Nice!

stalgiag (changes requested):
I see how this could make several tasks much easier. I left a few concerns inline. I may be overly cautious since I don't know the ansible environment well but if my understanding is correct here then I think we should add a couple of additional safeguards just to be careful.

server/controllers/DatabaseController.js
howard-e: Good point! Going over this again, I feel it unnecessary to even provide these custom paths for the clean and dump functions

server/controllers/DatabaseController.js
howard-e: Makes sense! Can also confirm the host as well

server/controllers/DatabaseController.js
howard-e: Definitely

stalgiag (approved):
Looks good! Thanks for addressing the feedback.

I'll leave to you to merge when you are ready @howard-e

PR #1302, opened 2025-01-09
Add husky and lint-staged

howard-e:


stalgiag (approved):
Tested locally and confirmed that this lints/prettifies prior to commits. It does not block commits that have linting errors that cannot be automatically fixed which is the preferred behavior. Thanks!

PR #1303, opened 2025-01-13
Bump `ansible-core` to address package support issues

howard-e:
Thursday's latest deploy to the staging environment caused an [error](https://github.com/w3c/aria-at-app/actions/runs/12698011352/job/35398847198) with the following message:

```bash
ModuleNotFoundError: No module named 'ansible.module_utils.six.moves'
```

`ubuntu-latest` was recently updated to Ubuntu 24.0.1, updated from the recent 22 LTS. This suggests that the previously packaged python [six module](https://pypi.org/project/six/) wasn't included by default in 24 as it was in 22.

Running the following would also address this error:

```bash
pip install six
```

But after checking the [ansible-core release versions](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix), we've long since been at EOL (Nov 22). So I'd like to take the chance to bump ansible-core from `2.11` to `2.16`. I've also confirmed this [builds and deploys](https://github.com/w3c/aria-at-app/actions/runs/12750303888/job/35534727666) as expected with minimal changes to the tasks (using `include_tasks` or `import_tasks` in place of `include`, introduced in [`ansible-core 2.12`](https://github.com/ansible/ansible/blob/stable-2.12/changelogs/CHANGELOG-v2.12.rst#deprecated-features)).

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -22,7 +22,7 @@ jobs:
           python-version: '3.10'
       - name: Install ansible and deploy to production
         run: |
-          python -m pip install --user ansible-core==2.11.1
+          python -m pip install --user ansible-core==2.16.14
           cd deploy
           echo ${{ secrets.ANSIBLE_VAULT_PASSWORD }} > ansible-vault-password.txt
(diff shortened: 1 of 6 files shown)


jugglinmike (approved):
Wonderful! Nice work, @howard-e :)

PR #1308, opened 2025-01-28
fix: restrict AT release dates to after 1999

outofambit:
Updates validation for the date entry field used in the add and update AT releases UI to validate that the entered date is after the year 1999.

related to https://github.com/w3c/aria-at/issues/1175

---
  - https://app.asana.com/0/0/1209264961572458


howard-e (approved):
Looks good. Thanks for patching this up!

@outofambit a more appropriate error message may be better received for the [modal](https://github.com/w3c/aria-at-app/blob/80f00261d369273e9dbda9164c17e88cad04e0ac/client/components/common/UpdateVersionModal/index.jsx#L155) now? Like `Please enter a valid date (after December 31, 1999)` or similar. I don't feel strongly on that phrasing but any thoughts?

outofambit:
Thanks @howard-e! Good point. I paraphrased something to make the message a little more descriptive.

howard-e:
> Thanks @howard-e! Good point. I paraphrased something to make the message a little more descriptive.

Looks good, thanks!

PR #1309, opened 2025-01-28
Only use push event for running tests

outofambit:
We're getting duplicate runs in some situations but "push" should cover all the cases we need on its own.


stalgiag:
I always figured this was intended. Perhaps to surface flakiness? I would prefer to only use `push` though.

outofambit:
@howard-e what do you think?

howard-e (approved):
Thanks! I'm fine with making this change

PR #1312, opened 2025-02-04
docs: Update `deploy/README.md` on how to use API functions to backup and restore database

### LINKED ISSUE #759: Configure staging and sandbox to pull tests from a aria-at-data-sandbox repo instead of aria-at

howard-e:
Close #759

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -52,38 +52,38 @@ can view logging from ansible with `sudo -i cat /var/log/messages`.
 To deploy this project to server:
 
 1. Merge from development to the releases branch as detailed in [docs/release.md](../docs/release.md).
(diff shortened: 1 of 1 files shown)


stalgiag (approved):
This is helpful. Thank you!

PR #1314, opened 2025-02-10
feat: Support multiple settings

howard-e:
Support multiple settings, being introduced in https://github.com/w3c/aria-at/pull/1125.

**Note: Do not merge until https://github.com/w3c/aria-at/pull/1125 is merged**

### DIFF
--- a/server/scripts/import-tests/gitOperations.js
+++ b/server/scripts/import-tests/gitOperations.js
@@ -3,7 +3,7 @@
 const spawn = require('cross-spawn');
 
 const ariaAtRepo = 'https://github.com/w3c/aria-at.git';
-const ariaAtDefaultBranch = 'master';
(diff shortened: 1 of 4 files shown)


server/scripts/import-tests/gitOperations.js
```diff
@@ -3,7 +3,7 @@
 const spawn = require('cross-spawn');
 
 const ariaAtRepo = 'https://github.com/w3c/aria-at.git';
-const ariaAtDefaultBranch = 'master';
+const ariaAtDefaultBranch = 'fix-1002-test'; // TODO: Revert to master before merging; based on test PR https://github.com/w3c/aria-at/pull/1126
```
howard-e, line 6: > [!WARNING]
> This branch was set to test an example scenario with the "Alert" plan and should be reverted before merging

PR #1317, opened 2025-02-10
fix: Include result of undesirable behaviors assertions in summary of failing assertions

howard-e:
Noticed that the description shared on the Candidate Review page for what failed wasn't matching the result when opening the Candidate Run page.

Example:

* Color Viewer Slider stating "16 assertions failed across 3 tests run with 1 browser" on the Candidate Review Page
screenshot of the above description from the Candidate Review Page

* but yet the Candidate Run Page states "10 assertions failed across 20 commands in 3 tests" so there was a discrepancy when it came to the number of assertions
screenshot of the above description from the Candidate Run page

Found this to be coming from no handling happening for the undesirable behaviors which are [derived failures](https://github.com/w3c/aria-at-app/blob/9c54e2d12e471c69f3bb5e79a11193df33c7fbe9/client/components/common/TestPlanResultsTable/index.jsx#L91:L120). This was also being missed in the overall getMetrics calculation so tests where only an undesirable behavior was listed, the test wasn't defined as "failed". So we could end up in situations where the summary would read "12 assertions failed across 0 tests run with 2 browsers"

 This PR now adjusts and recalculates for that.

### DIFF
--- a/client/components/TestRun/TestNavigator.jsx
+++ b/client/components/TestRun/TestNavigator.jsx
@@ -34,7 +34,11 @@ const TestNavigator = ({
   );
 
   const shouldShowFailingAssertionsSummary = useMemo(() => {
-    return isVendor && testPlanReport.metrics.assertionsFailedCount > 0;
+    return (
+      isVendor &&
+      (testPlanReport.metrics.mustAssertionsFailedCount > 0 ||
+        testPlanReport.metrics.shouldAssertionsFailedCount > 0)
+    );
   }, [isVendor, testPlanReport]);
 
   return (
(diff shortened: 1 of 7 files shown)


howard-e:
This mainly stemmed from the derived nature of the undesirable behavior assertion. It does raise a concern on where the source of that should really come from, whether generated as a "real" assertion at import time, continuing to be plugged as needed but ensuring there are proper utility functions to support it or otherwise.

stalgiag (approved):
This makes sense to me. Thanks for catching it and addressing it!

PR #1320, opened 2025-02-13
feat: Backend support and tests for refresh runs from previous at version

stalgiag:
This pull request adds backend support and testing for the refresh feature.

To keep this in a semi-reviewable state, I am going to open additional PRs targeting this one. This should be reviewed first but I will target this branch with subsequent related work.

### Remaining related tasks (follow-up)
- [ ] UI for initiating refresh runs
- [ ] Historical test assertion verdict copying currently assumes that the comparisons are made with a finalized run in the **same** test plan report. With this feature, this is no longer the case. (#1325 )
- [ ] GQL resolver for getting or counting refreshable reports

### Notes
- In order to untangle circular dependencies, I broke up `AtService` into `AtService` and `AtVersionService`. Apologies for the noise in the PR
- I opted to extend the testing db dump to provide the necessary data for testing

Can be reviewed but **should not be merged** since I want to aggregate additional changes here.


server/models/services/AtVersionService.js
howard-e, line 1: A welcome change!

server/models/services/AtVersionService.js
```diff
@@ -0,0 +1,460 @@
...
+  getAtVersionById,
+  getAtVersionByQuery,
+  createAtVersion,
+  updateAtVersionById,
+  updateAtVersionByQuery,
+  removeAtVersionByQuery,
+  removeAtVersionById,
+  findOrCreateAtVersion,
+  getUniqueAtVersionsForReport,
+  getRefreshableTestPlanReports: getRefreshableTestPlanReportsForVersion
```
howard-e, line 459: nit: Is there a reason for not exporting with the function name. Or should the function name be renamed to `getRefreshableTestPlanReports`?

Just wondering if it's because of the sole import location and that context why the rename happened here. If so, I personally would rather the rename happen where the function is being imported because this just causes a tiny bit of initial confusion when looking at these exports.

server/models/services/CollectionJobService.js
```diff
@@ -585,13 +592,6 @@ const scheduleCollectionJob = async (
     transaction
   });
 
-  const atVersion = await getAtVersionWithRequirements(
```
howard-e, line 588: Given this removal, doesn't the scheduleCollectionJobResolver#L17 now need to also pass `atVersion`? Assigning a bot now immediately fails or is that intended as part of the follow-up?

server/scripts/populate-test-data/pg_dump_test_data.sql
```diff
@@ -27,11 +27,28 @@ BEGIN
 END;
 $$;
 
+CREATE OR REPLACE FUNCTION get_at_version_id(atId integer, p_name text) RETURNS integer
```
howard-e, line 30: Neat!

server/models/services/CollectionJobService.js
stalgiag: I didn't finish addressing the changes on the frontend and that is part of why I didn't want to merge this work until more had aggregated here. So, yes, it is meant to be a follow-up but the work shouldn't be merged without it

But, the point you have here raises a really good question that makes me wonder why the `scheduleCollectionJob` gql mutation in the integration test is still working. Will look into it

server/models/services/AtVersionService.js
stalgiag: Ah good callout! No this is just due to me renaming the function using my IDE's "auto-rename" feature and the feature making a choice to keep the name the same in the export. Can fix

server/models/services/CollectionJobService.js
stalgiag: This was still working because of the way that the test version of a collection job is initialized. Luckily no eerie mysteries this time.

I took the resolver to a slightly more logical stopping point for the frontend work with [5c01031](https://github.com/w3c/aria-at-app/pull/1320/commits/5c01031a3701f238cbbe6c3155f86c227577b926)

stalgiag:
superceded by #1348

PR #1325, opened 2025-02-20
feat: Copy historical results between reports when run with previous automatable version

stalgiag:
This is an extension of #1320 and adds historical assertion verdict copying between reports for "refresh" runs

### DIFF
--- a/server/controllers/AutomationController.js
+++ b/server/controllers/AutomationController.js
@@ -29,7 +29,8 @@ const runnableTestsResolver = require('../resolvers/TestPlanReport/runnableTests
 const getGraphQLContext = require('../graphql-context');
 const populateData = require('../services/PopulatedData/populateData');
 const {
-  getTestPlanReportById
+  getTestPlanReportById,
+  updateTestPlanReportById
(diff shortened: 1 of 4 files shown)


server/controllers/AutomationController.js
```diff
@@ -250,89 +282,182 @@ const updateOrCreateTestResultWithResponses = async ({
...
+    }
 
-    const processedResponses = convertEmptyStringsToNoOutputMessages(responses);
+    res.json({ success: true });
+  } catch (error) {
+    const statusCode = error.statusCode || 500;
+    const errorResponse = {
+      error: error.message,
+      details: error.details || {},
+      stack: process.env.NODE_ENV !== 'production' ? error.stack : undefined
```
howard-e, line 390: Neat! Reasonable function here to make into a reusable error handler

howard-e (approved):
Thanks @stalgiag! Makes sense to me. The updated test really helped in following some of the changes here.

Nothing additional to call out from me except that wasn't already discussed in #1320.

PR #1329, opened 2025-03-04
Fix build issue and address date issue with snapshots

howard-e:
This PR does the following:

**Build Issue**
The [Postgres 16.8](https://github.com/actions/runner-images/blob/ubuntu24/20250302.1/images/ubuntu/Ubuntu2404-Readme.md#postgresql) bundled in the latest `ubuntu-24` image is unexpectedly unable to be ignored during the build process. It cannot be overwritten or have another postgres version installed alongside it. The last successful build's image used [Postgres 16.6](https://github.com/actions/runner-images/blob/ubuntu24/20250209.1/images/ubuntu/Ubuntu2404-Readme.md#postgresql) bundled so I can only assume it's because of the version or some permissions issue but not completely sure.

So this PR removes any instances of pre-bundled postgres pacakges first then installs our instance. It is a less than ideal but it eliminates this current blocker. Will continue to investigate with a lower priority.

**Snapshots Timestamp Issue**
Removes an outdated utility in favor of our standard UTC-based date utilities.

### DIFF
--- a/shared/dates.js
+++ b/shared/dates.js
@@ -1,19 +1,24 @@
+// TODO: momentjs is no longer supported. Replace with another date utility/practice. See https://momentjs.com/docs/#/-project-status/.
 const moment = require('moment');
 
-const convertDateToString = (date, format = 'DD-MM-YYYY') => {
+const convertDateToString = (date, format = 'DD-MM-YYYY', { locale } = {}) => {
   if (!date) return '';
+  if (locale) setLocale(locale);
   return moment.utc(date).format(format);
 };
 
-const convertStringToDate = (date, format = 'DD-MM-YYYY') => {
+const convertStringToDate = (date, format = 'DD-MM-YYYY', { locale } = {}) => {
+  if (locale) setLocale(locale);
   return moment.utc(date, format).toDate();
 };
 
 const convertStringFormatToAnotherFormat = (
   date,
   fromFormat = 'DD-MM-YYYY',
-  toFormat = 'MM-DD-YYYY'
+  toFormat = 'MM-DD-YYYY',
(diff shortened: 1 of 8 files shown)


shared/tests/dates.test.js
```diff
@@ -1,28 +1,64 @@
...
+    const formattedDate = convertDateToString(date, outputFormat, {
+      locale: 'en'
+    });
     expect(formattedDate).toBe('Nov 30, 2021 at 2:51:28 pm UTC');
   });
 
   it('returns a formatted string when using 24-hour notation in a different locale (French)', () => {
     const date = '2021-11-30T14:51:28.000Z';
-    const formattedDate = gitUpdatedDateToString(date, 'fr-FR');
+    const formattedDate = convertDateToString(date, outputFormat, {
```
howard-e, line 50: This project has yet to officially support ii8n so it felt a bit odd to preserve these locale based tests. Curious if others feel these necessary at the moment.

shared/dates.js
```diff
@@ -1,19 +1,24 @@
+// TODO: momentjs is no longer supported. Replace with another date utility/practice. See https://momentjs.com/docs/#/-project-status/.
```
howard-e, line 1: Very low priority item I had done several years ago in #483 and willing to redo if others feel strongly for this, in a separate PR.

shared/tests/dates.test.js
stalgiag: I love the ambition but yeah these seem superfluous for the foreseeable future.

stalgiag (approved):
Makes sense to me! Thanks for diving in and figuring out the issues with prebundled postgres

Issue #1334, opened 2025-03-06
Revise length of generated GitHub issue link


howard-e:
We are hitting the GitHub URI length limits in special cases due to the amount of content included in the link. We should revise the generated content to check if we can appropriately trim the included content.

In shortening the URI length in this way, it only slightly mitigates the solution and other alternatives are also welcome.

The latest instance has happened when attempting to raise an issue for the conflict for `Radio Group Example Using Roving tabindex V25.02.12, VoiceOver for macOS, Test 13`. The following URI was being generated:

<details>

<summary>the generated URI length is 10089</summary>

[...]

PR #1336, opened 2025-03-12
fix: Revise length of generated GitHub issue link to avoid hitting URI max length

stalgiag:
This PR truncates conflict markdown when a "Raise an issue" link has more than 8k chars.

Assumptions:
- 8k was tested in Chrome, Firefox, and Safari. It works on all. There is conflicting information about the limits for Edge but the most up-to-date info suggests that it shares Chrome's limit which would make sense.
- I couldn't create a testing scenario in which anything other than conflict markdown creates an issue link that exceeds this limit. If a reviewer believes otherwise please let me know.

This PR isn't entirely satisfying as it does result in truncated data in the issue. I tried to mitigate the worst effects of this by including a link back to the originating location. I looked into other options but the tradeoffs didn't seem worth it.

Alternatives
- Creating a GH issue via post to GH API
  - pros: no limitation on content
  - cons: would come from a central account rather than users' personals
- Replacing body with a link to a central view of the relevant info in the app
  - pros: no limitations on content, freedom with presentation
  - cons: requires navigating away from GH
  
related to #1334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+## [1.13.0](https://github.com/w3c/aria-at-app/compare/v1.12.0...v1.13.0) (2025-03-17)
+
+
+### Features
+
+* Group required reports in test plan report status dialog ([#1337](https://github.com/w3c/aria-at-app/issues/1337)) ([a705888](https://github.com/w3c/aria-at-app/commit/a705888df6a53abb69f011a94ffd641678949769))
+
+
+### Bug Fixes
+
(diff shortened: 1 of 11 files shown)


howard-e (commented):
Thanks for addressing. This seems like a safe enough solution to move forward with!

On the alternatives, GH API would definitely be a solid and long lasting solution. The in-app view sounds like a larger buildout of the `/conflicts` page. And with it just being considered an alternative (but always available) if GH cannot take all the content doesn't really place it as too much of a con for me! At a glance, the latter feels more favorable to me right now.

There may also be very minor changes and removal of some unnecessary white space in the template builder but not enough savings for me to require that be done right now.

Other than that, left an inline comment on the truncated text message.

howard-e (approved):
Confirmed internally that the changes work as expected. Thanks for sorting this out @stalgiag!

Issue #1332, opened 2025-03-05
Improve overall mobile experience


ccanash:
### /report/<id>/targets/<id>**

- [ ] Going beyond width bounds for h1 level title
- [ ] Breadcrumbs UI component not mobile friendly
- [ ] Wrapping of lines overlays on top of each other
- [ ] Test title section has buttons going beyond the width bounds

### /data-management
- [ ] Filter components going beyond width bounds
- [ ] Data Management page going beyond width bounds (the background being gray gives that impression, although we could assume it should allow that width to flex beyond)

### /test-queue
- [ ] ManageTestQueue component has several oddities not properly designed for mobile views
- [ ] TestPlanReport rows going beyond width bounds of page
- [ ] #1366

### /run/<id>
- [ ] AtBrowserDetailsDialog. Horizontally placed <select>s need to be vertically aligned
- [ ] AtBrowserDetailsDialog. Oddities with text wrapping

### /candidate-test-plan/<id>/<id>

- [ ] Headings going beyond width bounds

    

- [ ]  with some tests, if there is a dropdown it will show up under the Run Test Setup button

ccanash:
Closing as done and creating a new issue for the one item not addressed: https://github.com/w3c/aria-at-app/issues/1366

Issue #1330, opened 2025-03-05
Update (and modernize) styling approach for mobile


ccanash:
In order to modernize the style in mobile we will:

- [ ] Remove old styles
- [ ] Update breakpoints
- [ ] Standardize colors and dimensions (this definitely got away a bit. I'm seeing way too many grays)
- [ ] Prioritize responsive design

PR #1338, opened 2025-03-14
feat: declare css variables; move away from styled components

howard-e:
This PR addresses multiple items raised through:

* #1330 
* #1332 (except the manage bot dialog; that dialog was never brought up to par with the Test Queue update and I think should be handled separately)

This PR also:

* Switches from `@emotion/*` based styled components to introducing CSS modules (although much more optimizations can be done in that approach, open to additional feedback)
* Defines and trims down css variables that were being used throughout the application (still more purging could be done)
* Prioritizes responsive design in cases where it was lacking for smaller screens

NB. The file changes here are quite sizable (sorry!). But no functionality has been changed here outside of styling evaluations. Would recommended leaning on snapshot differences here initially.


client/components/BotRunTestStatusList/BotRunTestStatusList.module.css
```diff
@@ -0,0 +1,10 @@
+.bot-run-test-container {
+  font-size: 0.875rem !important;
```
stalgiag, line 2:
```suggestion
  font-size: var(--14px) !important;
```

client/components/BotRunTestStatusList/BotRunTestStatusList.module.css
```diff
@@ -0,0 +1,10 @@
...
+  padding: 0.5rem 0;
+  margin: 0.5rem 0;
```
stalgiag, line 4: It could be useful to have gap variables as well. A future improvement consideration

client/components/ManageTestQueue/ManageTestQueue.module.css
```diff
@@ -0,0 +1,103 @@
...
+
+  .form-group-at-version {
+    display: flex;
+    flex-wrap: wrap;
+    column-gap: 1rem;
+    row-gap: 0.75rem;
+
+    select {
+      width: inherit;
+      @media (max-width: 767px) {
```
stalgiag, line 93: Could be useful to have variables for breakpoints as well

client/components/common/None/index.jsx
stalgiag, line 1: very grateful for these small common components!

client/components/common/FilterButtons/FilterButtons.module.css
```diff
@@ -0,0 +1,18 @@
...
+
+  &.active,
+  &:active {
+    background: var(--bg-light-blue) !important;
+    border: var(--border-dark-blue) 2px solid !important;
+    box-shadow: none !important;
+
+    &:hover,
+    :focus {
+      box-shadow: 0 0 0 0.2rem rgba(103, 171, 197, 0.5) !important;
```
stalgiag, line 15:
```suggestion
      box-shadow: 0 0 0 0.2rem var(--button-shadow) !important;
```

client/components/common/styles.module.css
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+/* Colors */
+.dark-gray {
+  color: var(--bg-dark-gray);
+}
+
+span.required-indicator {
+  color: var(--bs-form-invalid-color);
```
stalgiag, line 51: We might want to get rid of these bootstrap variables since we are maintaining our own variables now.

client/webpack.config.js
```diff
@@ -30,6 +31,21 @@ module.exports = {
           'css-loader'
```
stalgiag, line 31: could be useful to use some kind of postcss-loader optimization. Not a requirement for right now but it could be useful for prod at least in the future.

stalgiag (changes requested):
Thank you for this truly heroic effort. I left a few items inline with specific requests. Some that should be done now, some that are optional and/or for later.

One bigger question is whether this strategy where all of the module are loaded rather than imported at a component level and applied with `styles.class` potentially using a conditional class utility like [clsx](https://www.npmjs.com/package/clsx). I think this can be a future improvement but I can see this path leading to difficulty naming classes and tracking where styles come from.

client/components/ManageTestQueue/ManageTestQueue.module.css
howard-e: Seems to still not be possible today without some other pre/post-processor. What I have done though is to use more general breakpoints across the app.

(I've still defined those variables because this aligns with your thoughts below on potentially including postcss-loader which gets a +1 from me)

client/components/BotRunTestStatusList/BotRunTestStatusList.module.css
howard-e: Certainly! I've started defining them in https://github.com/w3c/aria-at-app/pull/1338/commits/f9dff83a2a6b5600bd444e97c4b27901ed3b16ab but it feels like an even bigger effort to go through and change those now, though I wish I considered it at the time 😅. So I think we can certainly have this as a future improvement

client/components/common/styles.module.css
howard-e: Agreed! Done in https://github.com/w3c/aria-at-app/pull/1338/commits/83353dedef9dceb0a31872c847d9e11903c109d2

client/webpack.config.js
howard-e: Certainly! I'd love for us to set that up in a future improvement task

howard-e:
> One bigger question is whether this strategy where all of the module are loaded rather than imported at a component level and applied with `styles.class` potentially using a conditional class utility like [clsx](https://www.npmjs.com/package/clsx). I think this can be a future improvement but I can see this path leading to difficulty naming classes and tracking where styles come from.

Thanks for raising that. I agree with that being the optimal way forward. It was a bit less clear for me on how to do that here without a bit more refactoring because of how some of the components are interconnected. Let's create a task on that and resolve in a future improvement?

stalgiag:
> Thanks for raising that. I agree with that being the optimal way forward. It was a bit less clear for me on how to do that here without a bit more refactoring because of how some of the components are connected. Let's create a task on that and resolve in a future improvement?

Sounds good!

stalgiag (approved):

PR #1339, opened 2025-03-14
Avoid empty `issue.body` throwing an error when querying GH API for issues' content

howard-e:
Refer to title


stalgiag (approved):
Thanks!

PR #1340, opened 2025-03-17
fix: Account for `referencePage` update during copy process

howard-e:
In reviewing the newly imported result of https://github.com/w3c/aria-at/pull/1208, the referencePage name update was unnecessarily tossing away old results which shouldn't happen.

This also doesn't create new bot assignments if they were assigned to an old test plan run (and were not yet assigned to be evaluated). The bot assignemnt orchestration is done by a separate system which may cause unexpected issues during the copy (today, every new run undergoes a bot assignment as well)


stalgiag (approved):
This makes sense to me. Thanks for the quick fix!

Issue #1353, opened 2025-03-24
Data management page shows incorrect number of issues for a test plan


mcking65:
The number of open issues showing in the test plan status table is incorrect for some test plans. For example,
[Action Menu Button Example Using aria-activedescendant Test Plan Versions | ARIA-AT](https://aria-at.w3.org/data-management/menu-button-actions-active-descendant)
has 6 open issues. However the candidate review column of the test plan status table on the data management page shows "0 open issues".

![Image](https://github.com/user-attachments/assets/427f150d-f684-4944-90aa-177c58b792ad)

It would be helpful if the number of open issues, if >0, were a link to the issues table section on the test plan versions page.

PR #1357, opened 2025-03-26
fix: Incorrect GitHub sourced-issues count and displays

howard-e:
Address #1353.

While investigating #1353, I came up on several other issues which include:
* Fix confusing calls of `Issue.testNumberFilteredByAt`. There were mismatches between the row number vs the sequence number in practice which was leading to uncaught errors. eg. Test 1's issue showing up for Test 5 for example. This pins the usages to the more secure testRowNumber
* The Candidate Review Pages not accounting for "general feedback" as the original design didn't account for such and wasn't revisted so cases of `"x people left feedback for this test"` when it was feedback for the entire test plan
* Now outdated GitHub searches and copy when clicking `"x people left feedback for this test"` to view the linked GitHub issues:
  * "x" was being used to represent the number of issues so it got confusing clicking that and there were 4 issues tied to the same person for example
  * Various other copy updates

For reviewers, if you'd rather I split the original issue described above into it's own PR, I can also do that. I isolated that work to https://github.com/w3c/aria-at-app/commit/8935c476591284fa474bb0b439b7c32f3a560581


stalgiag (approved):
Thanks for quick fix and broader improvements! I would love for these values to exist on the `Test` at some point and then have the `Issue` optionally reference a test. I say this because there are still [decentralized calculations](https://github.com/w3c/aria-at-app/blob/fix-displayed-issues/client/components/TestQueue/Conflicts/index.jsx#L209) of the `testSequenceNumber`. It would be much easier to hand over a test id and trust that the test always knows what its row and at-to-display sequence is.

That said, I don't think that has to make it into this PR since this is already a huge improvement and needed fix. If you agree that this could be an improvement, I can open an issue for a follow-up.

Approving! Feel free to merged once you've read this.

howard-e:
> Thanks for quick fix and broader improvements! I would love for these values to exist on the `Test` at some point and then have the `Issue` optionally reference a test. I say this because there are still [decentralized calculations](https://github.com/w3c/aria-at-app/blob/fix-displayed-issues/client/components/TestQueue/Conflicts/index.jsx#L209) of the `testSequenceNumber`. It would be much easier to hand over a test id and trust that the test always knows what its row and at-to-display sequence is.

That would be excellent, really. I'm onboard for something like that but certainly a bit more to add in there.

> That said, I don't think that has to make it into this PR since this is already a huge improvement and needed fix. If you agree that this could be an improvement, I can open an issue for a follow-up.

I'd rather take you up on creating a follow up issue to make that change.

howard-e:
Will merge after https://github.com/w3c/aria-at-app/pull/1357/commits/0bad853fde7bee09b36b9940f56143a92d4d59fa builds

Issue #1351, opened 2025-03-24
Summary of failing assertions miscalculates number of commands


mcking65:
In the section of reports that summarizes failing assertions, the number of commands is incorrect. For example, in
[VoiceOver for macOS and Safari for Command Button Example | ARIA-AT Reports](https://aria-at.w3.org/report/74533/targets/247),
the sentence that summarizes the failures after the heading is:

>3 assertions failed across 10 commands in 1 tests.

There is only one command, not ten commands, that failed, "shift+j".

![Image](https://github.com/user-attachments/assets/2061d54c-7fba-4fdc-9f66-5d6b237c4ab3)

howard-e:
@stalgiag had a sync discussion with @mcking65 where we went over my above [comment](https://github.com/w3c/aria-at-app/issues/1351#issuecomment-2758828936). Found that several variations of reworking that text may potentially be confusing to viewers so it made sense to isolate the metrics (and revisit if any feedback arises from that). So the new format would be:

```
x of y total assertions fail across:
* z1 tests
* z2 unique assertion statements
* z3 unique commands
```

So for the linked report in the [top comment (Command Button Example V24.03.12 with VoiceOver for macOS and Safari)](https://aria-at.w3.org/report/74533/targets/247), this would read:

```
3 of 40 total assertions fail across:
* 1 test
* 3 unique assertion statements
* 1 unique command
```

stalgiag:
Is your understanding that this subheading should now be a subheading followed by an unordered list? This could have broader UI consequences.

mcking65:
@stalgiag 

This part of the text:

>x of y total assertions fail across:

would not be a heading, just a paragraph element followed by a list of 3 items.

PR #1358, opened 2025-03-26
fix: calculate unique commands for failing assertions in failing assertions summary table subheader

stalgiag:
see title

addresses #1351 

The previous number for commands used in this element was [the source of discussion during initial review](https://github.com/w3c/aria-at-app/pull/1288#discussion_r1907855493). There was uncertainty on the intention. It is now understood that this number should represent unique commands for which there is a failing assertion.

### DIFF
--- a/client/components/FailingAssertionsSummary/Table.jsx
+++ b/client/components/FailingAssertionsSummary/Table.jsx
@@ -21,7 +21,8 @@ const FailingAssertionsSummaryTable = ({
     <>
       <p>
         {failingAssertions.length} assertions failed across{' '}
-        {metrics.commandsCount} commands in {metrics.testsFailedCount} tests.
+        {failingAssertions.uniqueCommandsCount} commands in{' '}
+        {metrics.testsFailedCount} tests.
       </p>
 
       <Table bordered responsive aria-labelledby="failing-assertions-heading">
--- a/client/hooks/useFailingAssertions.js
+++ b/client/hooks/useFailingAssertions.js
@@ -6,7 +6,7 @@ export const useFailingAssertions = testPlanReport => {
       return [];
     }
 
-    return testPlanReport.finalizedTestResults.flatMap(
+    const failingAssertions = testPlanReport.finalizedTestResults.flatMap(
       (testResult, testIndex) => {
         return testResult.scenarioResults.flatMap(scenarioResult => {
           const commonResult = {
@@ -27,7 +27,10 @@ export const useFailingAssertions = testPlanReport => {
                 // Some revision of how that key combination + setting is rendered may be useful
                 return cmd.text.split(' (')[0];
               })
-              .join(' then ')
+              .join(' then '),
+            commandId: `${
+              scenarioResult.scenario.id
+            }_${scenarioResult.scenario.commands.map(cmd => cmd.id).join('_')}`
           };
 
           /**
@@ -72,5 +75,12 @@ export const useFailingAssertions = testPlanReport => {
         });
       }
     );
+
+    const uniqueCommandsWithFailures = new Set(
+      failingAssertions.map(a => a.commandId)
+    );
+    failingAssertions.uniqueCommandsCount = uniqueCommandsWithFailures.size;
+
+    return failingAssertions;
   }, [testPlanReport]);
 };
--- a/client/components/common/fragments/ScenarioResult.js
+++ b/client/components/common/fragments/ScenarioResult.js
@@ -29,6 +29,7 @@ const SCENARIO_RESULT_FIELDS = (type = 'simple') => {
           id
           output
           scenario {
+            id
             commands {
               id
               text
--- a/client/tests/e2e/snapshots/saved/_candidate-test-plan_24_1#summary.html
+++ b/client/tests/e2e/snapshots/saved/_candidate-test-plan_24_1#summary.html
@@ -300,7 +300,7 @@
                           <div
                             class="failing-assertions-summary-table-container">
                             <p>
-                              3 assertions failed across 26 commands in 3 tests.
+                              3 assertions failed across 3 commands in 3 tests.
                             </p>
                             <div class="table-responsive">
                               <table
--- a/client/tests/e2e/snapshots/saved/_report_67_targets_20.html
+++ b/client/tests/e2e/snapshots/saved/_report_67_targets_20.html
@@ -370,7 +370,7 @@
           <h2 id="failing-assertions-heading">
             Summary of Failing Assertions (0 must, 1 should)
           </h2>
-          <p>1 assertions failed across 28 commands in 1 tests.</p>
+          <p>1 assertions failed across 1 commands in 1 tests.</p>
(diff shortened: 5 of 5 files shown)


client/components/FailingAssertionsSummary/Table.jsx
```diff
@@ -21,7 +21,8 @@ const FailingAssertionsSummaryTable = ({
     <>
       <p>
         {failingAssertions.length} assertions failed across{' '}
-        {metrics.commandsCount} commands in {metrics.testsFailedCount} tests.
+        {failingAssertions.uniqueCommandsCount} commands in{' '}
```
howard-e, line 24: This needs some conditional copy here for single/plurals handling or it could read:

> 1 assertions failed across 1 commands in 1 tests.

Also thinking the "across" may not read well in a situation above, so examples could be:
> 1 assertion failed for 1 command in 1 test.
> 2 assertions failed for 1 command in 1 test.
> 1 assertion failed across 2 commands in 1 test.
> 1 assertion failed for 1 command in 2 tests. (this one isn't possible with the current set up so asking a follow-up below)

I don't feel strongly on the conditional "across" -> "for" though so I'm willing to go along without that particular change.

client/hooks/useFailingAssertions.js
```diff
@@ -72,5 +75,12 @@ export const useFailingAssertions = testPlanReport => {
         });
       }
     );
+
+    const uniqueCommandsWithFailures = new Set(
+      failingAssertions.map(a => a.commandId)
+    );
+    failingAssertions.uniqueCommandsCount = uniqueCommandsWithFailures.size;
```
howard-e, line 82: Neat! This works but it does now raise another question that I wasn't able to gather from the previous discussions, so wondering if the answer was somewhere and I missed it. If not, we can ask it in the latest bug report to further remove any ambiguity.

Q: Should the _unique_ command increment be based on the individual test or on the entire group of tests? Consider the following summary of failing assertions table:

| Test Name                          | Command                            | Assertion Priority | Assertion                                   | JAWS Response                                                                                      |
|------------------------------------|------------------------------------|--------------------|---------------------------------------------|---------------------------------------------------------------------------------------------------|
| Navigate backwards to a button | Shift+j (virtual cursor active)       | MUST               | assertionStatement1              | response1            |
| Navigate backwards to a button | Shift+j (virtual cursor active)            | SHOULD               | assertionStatement2              | response2            |
| Navigate forwards to a button| Shift+j (virtual cursor active) | MUST               | assertionStatement3              | response3            |

So, should the table state:
a) "3 assertions failed across 2 commands in 2 tests" or
b) "3 assertions failed for 1 command in 2 tests"

a) is what happens now.

howard-e (commented):
@stalgiag thanks for addressing this so quickly!

I've only left a small comment on the copy but there is another question I've left inline to further remove any ambiguity I may still have.

client/hooks/useFailingAssertions.js
stalgiag: It is a good idea to ask to further remove ambiguity. I would be very surprised (bordering on disagreeing with the design) if it was b) since those three rows have the same keychord+setting combo but theoretically 2 different context/starting states.

Would you like to pose the question as you've outlined it here or should I copy your words and do it?

client/components/FailingAssertionsSummary/Table.jsx
stalgiag: Ah yes, thanks for catching the plurals!  I also agree with "for" replacing "across". That word made more sense when the number was for the total number of commands included in the report.

Completed in [8c2e38c](https://github.com/w3c/aria-at-app/pull/1358/commits/8c2e38cf3a72ec68cac0e765fcc8f43c8d5f8f08)

client/hooks/useFailingAssertions.js
howard-e: > Would you like to pose the question as you've outlined it here or should I copy your words and do it?

I can, shared the question on the issue, https://github.com/w3c/aria-at-app/issues/1351#issuecomment-2758828936

client/components/FailingAssertionsSummary/Table.jsx
howard-e: Looks good! Thanks for addressing!

stalgiag:
@howard-e I believe this is ready for review again. I *believe* that the fields are now what is desired. I was able to verify with the example given, though the total assertions are 20 and not 40 but that tracks with my manual count of the total assertions for the test plan version on VO.
<details>
<summary>Screenshot of failing assertions summary table in local development environment</summary>
A screenshot showing the failing assertions summary table, the content of the p element reads: 3 of 20 total assertions fail across:, underneath that is an unordered list with: • 1 test • 3 unique assertion statements • 1 unique command
</details>

howard-e:
@stalgiag looks good! But for the following statement:

> 20 and not 40 but that tracks with my manual count of the total assertions for the test plan version on VO

I added this [commit](https://github.com/w3c/aria-at-app/pull/1358/commits/d931d241527bdb741ccece26e3cd32549995ed54) to help move this review along 

The 40 total comes from the 2 additional severe and moderate behavior-related assertions that get conditionally included per command. It also matches up with the `MUST` and `SHOULD` counts on the report page’s results table. 

Those assertions were included at a time when the assumptions around them were less clear, so got added in a less stable way. Seems this should be provided in a more stable way now especially as the metrics usage expands. To avoid further confusion on values. I’ve left notes to reflect that.

A note for more clarity: `failingAssertions.totalAssertionsCount` could have been `testPlanReport.metrics.mustAssertionsCount + testPlanReport.metrics.shouldAssertionsCount` (but looking at how this hook is being used with only references to `testPlanReport.finalizedTestResults`, seems like preserving less dependency on another attribute is preferred)

If this seems good with you, I’m happy to merge.

stalgiag:
> @stalgiag looks good, but for the following statement:
> 
> > 20 and not 40 but that tracks with my manual count of the total assertions for the test plan version on VO
> 
> I added this [commit](https://github.com/w3c/aria-at-app/pull/1358/commits/d931d241527bdb741ccece26e3cd32549995ed54) to help move this review along 
> 
> The 40 total comes from the 2 additional severe and moderate behavior-related assertions that get conditionally included per command. It also matches up with the `MUST` and `SHOULD` counts on the report page’s results table.
> 
> Those assertions were included at a time when the assumptions around them were less clear, so got added in a less stable way. Seems this should be provided in a more stable way now especially as the metrics usage expands. To avoid further confusion on values. I’ve left notes to reflect that.
> 
> A note for more clarity: `failingAssertions.totalAssertionsCount` could have been `testPlanReport.metrics.mustAssertionsCount + testPlanReport.metrics.shouldAssertionsCount` (but looking at how this hook is being used with only references to `testPlanReport.finalizedTestResults`, seems like preserving less dependency on another attribute is preferred)
> 
> If this seems good with you, I’m happy to merge.

Thanks for these changes! I didn't have a memory of these additional conditional assertions. Everything looks good to me.

Issue #1319, opened 2025-02-12
Include snapshot updating in GH Actions


howard-e:
We frequently end up having to re-run `yarn update-snapshots` should an update be pushed to [aria-at's default branch](https://github.com/w3c/aria-at).

We can automate updating the snapshots via an action on push, so contributors can avoid understanding the nuance around that.

cc @stalgiag

stalgiag:
This would be great. Thanks for suggesting it!

PR #1359, opened 2025-03-26
chore: add workflow for automatically updating snapshots

stalgiag:
This PR adds a workflow that can be triggered by the aria-at repo. The workflow sets up a test environment, updates the snapshots, and pushes to development. The test environment setup was created as a shared action for easier maintenance.

Leaving in `DRAFT` until the broader approach is agreed upon.

addresses #1319


howard-e (approved):
Looks good! Thanks for optimizing this process. Will definitely make things easier especically for contributors outside ourselves.

Also approved the related PR at https://github.com/w3c/aria-at/pull/1223.

PR #1361, opened 2025-03-31
Align `ManageBotDialog` with Test Queue V2 visual changes

howard-e:
Follow up from #1338:

> (except the manage bot dialog; that dialog was never brought up to par with the Test Queue update and I think should be handled separately)


client/components/ManageBotRunDialog/StopRunningCollectionButton/index.jsx
```diff
@@ -7,18 +7,19 @@ import { LoadingStatus, useTriggerLoad } from '../../common/LoadingStatus';
...
       collectionJobId: collectionJob.id
-    }
+    },
+    skip: !collectionJob
   });
 
+  if (!collectionJob) {
+    return null;
+  }
+
```
stalgiag, line 22: Nice!

stalgiag (approved):
Looks good to me! Definitely an improvement

PR #1363, opened 2025-04-08
feat: QA updates & reset db button to reset to build dump during testing

stalgiag:
I've added onto this PR as it became the most convenient location for QA changes. Obviously not ideal and I am willing to extract the reset button changes if that is desired. Either way, it useful to have a separate PR for the latest changes to the Report Rerun feature due to the significant but isolated UI changes recommended after QA.

[See this doc for the QA feedback](https://docs.google.com/document/d/1eGhwq3cqryaAikvIKnLgaAA6iTLZ_MWa28ZxmmtC3pA/edit?tab=t.0#heading=h.kghksqgdkv63) that led to the UI changes.

One item of feedback was skipped for this round, namely:

> "9. Instead of a separate section for events, when reports are running, consider adding a run status column to the table and show a status for each report. Later, if actions are available, you might add an action column. Alternatively, make an event history table to which we could later add sort and filter functions ."

This could potentially be a future improvement but as it stands now it would introduce too many changes to the way that the view state as available update "cards" are removed when updates aren't available (as per another item of feedback), therefore a separate element tracking is necessary since updates can happen long after the initial rerun start.

Old note:
This PR adds a Reset DB button that can be used for testing in sandbox. This feature allows one to reset the DB to the state it was at during the previous build. This can be used for manual QA for interactive features with permanent linear-forward effects.


client/tests/e2e/snapshots/saved/_candidate-test-plan_24_1.html
```diff
@@ -697,7 +697,7 @@ <h1 class="disclosure-heading">
                                       >esmeralda-baggins</a
                                     ></b
                                   >
-                                  on April 01, 2025.
+                                  on April 08, 2025.
```
howard-e, line 700: Realized this was accidentally reintroduced as part of the snapshots with the css re-work. https://github.com/w3c/aria-at-app/pull/1361/commits/a637884800157e9d4aeb9b0deba6d6264a2ac226 in #1361 should fix it. But let me know if you'd like me to just pull that into it's own PR instead

howard-e (approved):
A very useful utility function! Thanks for thinking about this.

It could serve more needs outside just this feature. Maybe we put it on the admin settings page? No need to do because it's such a small thing but just a thought.

client/tests/e2e/snapshots/saved/_candidate-test-plan_24_1.html
stalgiag: Fine with it being fixed over there. Thanks!

client/components/ReportRerun/index.jsx
```diff
@@ -91,55 +118,111 @@ const ReportRerun = ({ onQueueUpdate }) => {
...
   };
 
   return (
-    <div className={styles.rerunSection}>
-      {isAdmin && (
-        <RerunDashboard
-          activeRuns={activeRuns}
-          onRerunClick={handleRerunClick}
+    <LoadingStatus message={loadingMessage}>
+      <div className={styles.rerunSection}>
```
howard-e, line 198: Can't seem to find `.rerun-section` in case a style has been wrongly chosen.

howard-e (approved):
Changes look good to me after reviewing Isabel's and Matt's comments in the [QA doc](https://docs.google.com/document/d/1eGhwq3cqryaAikvIKnLgaAA6iTLZ_MWa28ZxmmtC3pA/edit?tab=t.0).

I also agree with your thoughts on:

> One item of feedback was skipped for this round, namely:
> 
> > "9. Instead of a separate section for events, when reports are running, consider adding a run status column to the table and show a status for each report. Later, if actions are available, you might add an action column. Alternatively, make an event history table to which we could later add sort and filter functions ."
> 
> This could potentially be a future improvement but as it stands now it would introduce too many changes to the way that the view state as available update "cards" are removed when updates aren't available (as per another item of feedback), therefore a separate element tracking is necessary since updates can happen long after the initial rerun start.

I've left another point of feedback in the parent PR #1348, https://github.com/w3c/aria-at-app/pull/1348#discussion_r2054172204 which may be inconsequential.

client/components/ReportRerun/index.jsx
stalgiag: Thank you! [2bd346d](https://github.com/w3c/aria-at-app/pull/1363/commits/2bd346d2cd8b061a282a1c0f9c69df9765046d85)

Issue #1371, opened 2025-04-10
Data management page showing: Response not successful: Received status code 500


ccanash:
The Data management page is not loading for logged in or logged out users.

Message received is: **Response not successful: Received status code 500.** 

cc @IsaDC 

Error:

{
    "errors": [
        {
            "message": "Converting circular structure to JSON\n    --> starting at object with constructor 'ClientRequest'\n    |     property 'socket' -> object with constructor 'TLSSocket'\n    --- property '_httpMessage' closes the circle",
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR",
                "exception": {
                    "stacktrace": [
                        "TypeError: Converting circular structure to JSON",
                        "    --> starting at object with constructor 'ClientRequest'",
                        "    |     property 'socket' -> object with constructor 'TLSSocket'",
                        "    --- property '_httpMessage' closes the circle",
                        "    at JSON.stringify (<anonymous>)",
                        "    at prettyJSONStringify (/home/aria-bot/aria-at-report/node_modules/apollo-server-core/dist/runHttpQuery.js:315:17)",
                        "    at processHTTPRequest (/home/aria-bot/aria-at-report/node_modules/apollo-server-core/dist/runHttpQuery.js:234:20)",
                        "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"
                    ]
                }
            }
        }
    ]
}

PR #1372, opened 2025-04-10
fix: Use cursor-based pagination for navigating GitHub issues

howard-e:
Addresses #1371

More notes at https://github.com/w3c/aria-at-app/issues/1371#issuecomment-2794604825

* Use cursor based pagination for navigating GitHub issues instead
* Handle issues query error


stalgiag (approved):
Confirming that the change to pagination works as expected. The `try/catch` is also welcome. Thanks for the quick fix!

PR #1375, opened 2025-04-15
Correct exit code for development script

jugglinmike:
Prior to this patch, when the "populate test data" script failed, it simply printed the error object and exited with a successful exit code. This made recognizing the failure difficult because the error object includes hundreds of lines of context which do not obviously indicate a problem.

Signify an error has taken place using a non-zero exit code as per UNIX convention. This allows the caller to recognize the failure (and in particular, it causes the `npm` utility--which invokes this script on the user's behalf--to prominently describe the failure).


howard-e (approved):

howard-e:
Thanks for catching!

PR #1381, opened 2025-04-22
fix: Preserve automation initiated status for runs across results copy

howard-e:
Address #1380.

The main issue arose when results during a version copy was being done.

ie. If the tester A's results in TestPlanVersion 1 were initially created using automation, that flag wasn't brought across for tester A's results in TestPlanVersion 2, which caused the UI to render the status message as `x tests completed` instead of `x tests evaluated`


client/components/TestQueue/CompletionStatusListItem/index.jsx
howard-e, line 1: For reviewers: This file was being used but from the wrong location. So this is what was moved to `/components/TestQueueCompletionStatusListItem/index.jsx` so wanted to take the chance to address it here because I only noticed when investigating this issue.

client/components/TestQueueCompletionStatusListItem/index.js
howard-e, line 1: For reviewers: This file wasn't being used. The contents of it should have been what `/components/TestQueue/CompletionStatusListItem/index.js` were so that's been moved here with no functional changes.

jugglinmike (approved):
Works on my end! Thanks, @howard-e!

Issue #1368, opened 2025-04-09
Add functionality to set the minimum version of the AT when adding a test plan to the test queue from the report status dialog


ccanash:
As requested in [Policy for minimum AT versions when adding plans to Test Queue](https://github.com/w3c/aria-at/issues/1213): 

When an admin adds a test to the test queue from the test report status dialog, please add functionality to select the minimum AT version to use. 

![Image](https://github.com/user-attachments/assets/efa6f5e4-db42-4480-8530-84ecf3d9dea9)

This functionality should be similar to the ability to specify a minimum AT version when adding a run to the queue from the "Add Test Plans to the Test Queue" disclosure:

![Image](https://github.com/user-attachments/assets/fcab6f57-6ee4-42e2-9ce5-2f82de8cbea8)

mcking65:
I updated the description with more specifics. This should be fully actionable now.

PR #1386, opened 2025-04-29
feat: Allow setting minimum AT version from test plan report status dialog

howard-e:
Address #1368

Testing guidelines:

1. Sign in as an admin
2. Open the Test Plan Report Status Dialog via the Required Reports button via the interactions on the Test Queue and Data Management pages
   1. From the Test Queue page, expand any of the test plan disclosures and it is available at the top of the expanded view
   2. On Data Management, it is on any of the active phase cells other than R&D
3. If a row is available to be added to the Test Queue, change the minimum version from the inline dropdown then add to the test queue [screenshot attached]
4. Observe that the row is added with the selected minimum version saying '{{AT}} {{SELECTED_AT_VERSION}} or later'
5. Observe that the report with that specified version is now a part of the Test Queue.

screenshot illustrating number 3 from the above list


client/components/TestPlanReportStatusDialog/index.jsx
```diff
@@ -42,16 +42,35 @@ const TestPlanReportStatusDialog = ({
...
+    );
+  };
+
+  const tableRows = sortedStatuses.map((status, index) => {
     const {
       isRequired,
       at,
       browser,
-      minimumAtVersion,
+      // minimumAtVersion,
```
ChrisC, line 67: You can probably just delete this?

client/tests/e2e/TestQueue.e2e.test.js
```diff
@@ -236,6 +254,136 @@ describe('Test Queue admin traits when reports exist', () => {
...
+      // Get required VoiceOver row
+      const voRowBeforeSelectChange =
+        await getVoRowFromTestPlanReportStatusDialog(page);
+      const atVersionSelectHandle = await voRowBeforeSelectChange
+        .asElement()
+        .$('select.minimum-at-version-select');
+
+      await page.evaluate(select => {
+        // Set to VoiceOver 14.0 (At.id is 5 from sample data). Default At.id was 3 (VoiceOver 11.6 (20G165))
+        select.value = '5';
```
ChrisC, line 347: This is the reason this test is failing for me locally. My local db is probably not in the state it needs to be in order for this test to pass, but I'm also wondering why we're not just setting the value to the id of the AT version (for example, for 14.0)... Syncing the values might make for more durable tests and maybe just makes more senes? Is there a reason we're using a separate numbering scheme for the values here? 

Screenshot from a test run on my local at the moment where this test fails for me:
Screenshot 2025-05-22 at 12 06 52 PM

ChrisC (commented):
I have one question re: failing e2e tests on my local machine, but otherwise this feature is working for me and implementation looks ok! 

One UI note that I don't think we need to fix now: I couldn't actually figure out where to pop open the dialog with the updated table because the buttons are not visually distinct or styled as buttons! I basically didn't even know those "Required Reports" buttons were even clickable until you added the steps to test! This might be something we'll want to revisit later?

client/tests/e2e/TestQueue.e2e.test.js
howard-e: Odd! Is it possible you had a conflicting db change here? There doesn't seem to be any seeders that adds VoiceOver 13.0 at least on this branch. The following screenshot of just the v11.6 and v14.0 are what I get from a fresh hydration of the test db, with value `"5"` still being for 14.0:

screenshot describing above

howard-e: @ChrisC merged in `main` to fix the expected value in https://github.com/w3c/aria-at-app/pull/1386/commits/4470f1ae4836d1c035a5b6a3c655188467a001ca

ChrisC: Ok now the test is passing locally for me now with this change! Tho I'm still curious about why we're using separate incrementing numbers for the values here as that seems a little brittle. But if it's a local test data sync issue, then maybe the bigger problem is how were setting up and tearing down test data across multiple environments.

PR #1396, opened 2025-05-06
fix: Replace with better understood live data example for `getMetrics.test.js`

howard-e:
Coming from discussions with @jugglinmike, there was a need for the mock data used in the `getMetrics` to not be as minimalistic as it is, given the absence of some clearer documentation around this function


jugglinmike (approved):
Nice! Thanks!

PR #1397, opened 2025-05-06
Add more tests for metrics calculation

jugglinmike:
The existing deterministic test uses fixture data which is large enough to challenge comprehension. Write additional tests which are based on smaller datasets, produced from a concise form which is easier to read at a glance.


howard-e (approved):
Looks good to me and indeed simpler to read

PR #1398, opened 2025-05-07
Refactor code into dedicated component

jugglinmike:
This factoring reduces the responsibilities of an existing component by relocating logic to a new component. That new component will serve as a basis for further extension in a forthcoming feature patch.

---

@howard-e Just trying to lighten the review load a bit by factoring out some non-substantive changes that can meaningfully stand alone.


howard-e (approved):
Looks good, these changes are always a huge plus

jugglinmike:
Thanks, @howard-e!

Issue #1352, opened 2025-03-24
Change unexpected behaviors to negative side effects and add a "is test blocked" question to test runner


mcking65:
This issue implements changes documented in w3c/aria-at#1043 and proposes a solution for w3c/aria-at#1216.

#### Change unexpected behavior wording in reports, candidate review, and test runner results

for w3c/aria-at#1043, make the following changes to assertion wording for unexpected behaviors, now to be called, negative side effects.

Assertion Form|Old Wording|New Wording
--- | --- | ---
Assertion Statement|Other behaviors that create moderate negative impacts are not exhibited|Moderate negative side effects do not occur
Assertion Statement|Other behaviors that create severe negative impacts are not exhibited|Severe negative side effects do not occur
 Assertion Phrase|Other behaviors that create moderate negative impacts are not exhibited|not cause moderate negative side effects
Assertion Phrase|Other behaviors that create severe negative impacts are not exhibited|not cause severe negative side effects

Note: It is intentional that the assertion phrases start with lower case "not". This is part of the convention for the test format V2.

In reports for commands with unexpected behaviors, change the phrase that appears above the table of unexpected behaviors from:

>Other behaviors that create negative impact:

to:

>Negative side effects of COMMAND_NAME

and mark it up as a level 4 heading (same level as the command results heading)

Change the caption on the table from:

>Undesirable behaviors for test TEST_NAME

to:

>Negative side effects of COMMAND_NAME

In the table, change the title of the "Behavior" column to "Side Effect".

#### Change to test runner wording for unexpected behaviors

for w3c/aria-at#1043, in the test runner, change wording as follows:

Old Wording|New Wording
--- | ---
Were there additional undesirable behaviors?|Did negative side effects occur?
No, there were no additional undesirable behaviors.|No, negative side effects did not occur.
Yes, there were additional undesirable behaviors.|Yes, negative side effects occured.

In the list of possible side effects, remove the words "behavior occurred" from each option. For example, change:

>Reading cursor position changed in an unexpected manner behavior occurred

to:

>Reading cursor position changed in an unexpected manner

#### In test runner, add checkbox indicating whether assertions are testable

To resolve w3c/aria-at#1216, in the test runner, after the output and before the first assertion, add a checkbox with the following label. Default the state to "not checked". 

>The response to COMMAND_NAME created conditions that made assertions untestable, e.g., focused the wrong element.

If the checkbox is checked:

1. Disable assertion verdict radio buttons.
[...]

PR #1403, opened 2025-05-13
fix: Correct type of metrics

jugglinmike:
Ensure that the `testsCount` and `testsPassedCount` properties of the "metrics" object are always numeric values.

---

@howard-e The current behavior is causing propType failures on [my feature branch](https://github.com/w3c/aria-at-app/compare/development...bocoup:aria-at-app:gh-1352-test-blocked) for gh-1352.


howard-e (approved):
Good catch!

jugglinmike:
Thanks!

PR #1406, opened 2025-05-14
fixes candidate test plan run page titles

### LINKED ISSUE #1388: Candidate review page title elements have inaccurate values
This is not the only `<title>` element problem in the app. Unless it is somehow more efficient to work on a more global fix, I suggest we continue prioritizing improvement to the candidate review experience.

When you navigate to the candidate review page from the main nav, the title tag has the correct value: "Candidate Review | ARIA-AT".

However, after opening a specific test plan, the page title tag has the value "Candidate Test Run Page | ARIA-AT". This title is misleading because it does not accurately reflect the page or its purpose. It doesn't indicate which test, which test plan, and it implies the user is running the test, instead of reviewing it.

There are separate problems with heading levels and content. I will raise those as separate issues.

I would suggest the form of the page title should be:

>TEST_TITLE | TEST_PLAN_TITLE | Candidate Review | ARIA-AT

Example:

>Navigate forwards to a menu button | Action Menu Button Example Using aria-activedescendant | Candidate Review | ARIA-AT

Alternatively, since we have the tests numbered in the navigation, it might be more accurate and helpful to include the number:

>1. Navigate forwards to a menu button | Action Menu Button Example Using aria-activedescendant | Candidate Review | ARIA-AT

ChrisC:
Page titles for candidate test plans runs now have the format:
> TEST_TITLE | TEST_PLAN_TITLE | Candidate Review | ARIA-AT

Specific example: 
> 1. Navigate forwards to a menu button | Action Menu Button Example Using aria-activedescendant | Candidate Review | ARIA-AT

If user navigates to the failed assertion summary, the title reads: 
> Summary of Failing Assertions | TEST_PLAN_TITLE | Candidate Review | ARIA-AT

If I have a little more time this week I may modify the candidate review e2e tests to reflect this change, but they are captured in updated e2e snapshots so that may be good enough...

Closes #1388

### DIFF
--- a/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
@@ -312,6 +312,10 @@ const CandidateTestPlanRun = () => {
   const { testPlanVersion } = testPlanReport;
   const { recommendedPhaseTargetDate } = testPlanVersion;
 
+  const pageTitle = `${currentTestIndex + 1}. ${currentTest?.title} | ${
+    testPlanVersion.title
+  } | Candidate Review | ARIA-AT`;
+
   const reviewStatusText = vendorReviewStatusMap[reviewStatus];
 
   const targetCompletionDate = dates.convertDateToString(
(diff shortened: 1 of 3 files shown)


howard-e (approved):
This change looks good to me, thanks @ChrisC!

PR #1408, opened 2025-05-21
Improve accuracy of rendered form

jugglinmike:
Prior to this commit, if the user indicated that there were unexpected behaviors without specifying any particular behaviors, their subsequent submission would be rejected, and the form would be rendered as though they had not indicated the presence of unexpected behaviors.

Update the rendering logic to more accurately describe the user's input following a failed submission.

---

@howard-e I think the current behavior is counter-intuitive in its own right (see above), but it will become more confusing when we implement the "untestable" checkbox. [That feature is spec'd to set the "unexpected behaviors" checkbox when enabled.](https://github.com/w3c/aria-at-app/issues/1352) If the user sets "untestable" without changing anything further and submits, then (with the current behavior) the form will return to them in an invalid state: "untestable" checked and "unexpected behaviors" unset.

### DIFF
--- a/client/components/TestQueue/queries.js
+++ b/client/components/TestQueue/queries.js
@@ -138,6 +138,7 @@ export const TEST_QUEUE_CONFLICTS_PAGE_QUERY = gql`
           }
           scenarioResult {
             output
+            hasUnexpected
             unexpectedBehaviors {
               text
               details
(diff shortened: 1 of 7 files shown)


jugglinmike:
This isn't quite right...

jugglinmike:
I think I got it

jugglinmike:
The tests aren't running. It might be because I opened this pull request from a branch on a fork. I'll try pushing upstream and opening a new pull request from there...

PR #1424, opened 2025-06-05
Support handling of unknown key commands that don't exist in commands.json

howard-e:
It was found that there may be a chance for newly added keys in aria-at's support.json file to not be captured in the deployed app's refresh which could then cause an error state if a test plan exists which uses any of those keys.

This helps prevent that error state from fully blocking the render of specific pages, such as the Test Queue.


stalgiag (approved):
Makes sense. Thanks for the fix!

PR #1428, opened 2025-06-11
Allow admins to approve Candidate Reports

howard-e:
Address #1410

This comes from a recent discussion which highlighted there may be an eventual need for an admin to approve a candidate test plan report on behalf of a vendor.


stalgiag (changes requested):
This looks great. I especially appreciate the updated "in-progress" calculation.

One small request, can we update the `vendor` property in the `ReviewerStatus` type in the GraphQL schema to be optionally null now that it is sometimes explicitly set to that value?

stalgiag (approved):

Issue #1429, opened 2025-06-12
Implement revisions to reporting requirements for untestable assertions that were specified in issue 1352


mcking65:
After assessing the implementation of the ability to generate reports for untestable assertions specified by #1352, I think we need to modify the following reporting requirements:

>#### Update reporting code to support untestable assertions
>
>When assertions have verdicts of "Untestable", change the reports to:
>4. When calculating the percentage of assertions that pass, leave untestable assertions out of both numerator and denominator.
>7. Even though untestable assertions are not technically "filures", add rows for them to the summary of failures.

I generated this sample report in staging: [JAWS and Chrome for Alert Example | ARIA-AT Reports](https://aria-at-staging.w3.org/report/109806/targets/298). As I assessed the understandability of the report, I concluded these requirements make the reports difficult to understand. I believe the following revisions to the requirements will resolve the difficulties.

#### Change to issue 1352 reporting requirement 4

>4. When calculating the percentage of assertions that pass, leave untestable assertions out of both numerator and denominator.

This requirement applies any place that we would show a percentage of passing must+should assertions. For example, the [top-level interop reports page](https://aria-at.w3.org/reports#main) shows a percentage of passing assertions.

[insert image]

An untestable assertion does not pass or fail. So, requirement 4 suggests that we would reduce the total number of assertions counted in the denominator. However, this would make it difficult for readers to understand how the percentage of passing assertions was calculated because it would be difficult to discern why some assertions are excluded. While there are valid arguments for not penalizing the AT for an untestable assertion (at least in some circumstances), this is a condition where simplicity and clarity is likely far more important than "fairness" in some edge cases.

Instead of implementing requirement 4 as written in #1352, I propose that we keep the calculation of percentage passing assertions unchanged from its prior formula:

>(count of passing p1 assertions + count of passing p2 assertions) / (count of p1 assertions + count of p2 assertions)

#### Change to issue 1352 reporting requirement 7

>7. Even though untestable assertions are not technically "filures", add rows for them to the summary of failures.

After reading the sample report, I now see two reasons for why this requirement is problematic:
1. Any untestable p3 assertions will be excluded.
2. The reader cannot distinguish between untestable and failing assertions, which is critical information.

The summary of failing assertions serves a very specific set of purposes, for which we may be providing more extensive support in the future. To maintain clarity and utility, it is likely best that we do not cloud the summary of failing assertions with untestable assertions.

Instead of implementing requirement 7 as specified in #1352, I propose the following.

[...]

PR #1440, opened 2025-06-19
Revise design of untestable assertions

jugglinmike:
Resolves gh-1429.

I'll leave this in "draft" until [we receive confirmation about the assumption I've made in this implementation](https://github.com/w3c/aria-at-app/issues/1429#issuecomment-2986286136).

### DIFF
--- a/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
+++ b/client/components/CandidateReview/CandidateTestPlanRun/index.jsx
@@ -35,6 +35,8 @@ import FailingAssertionsSummaryTable from '../../FailingAssertionsSummary/Table'
 import FailingAssertionsSummaryHeading from '../../FailingAssertionsSummary/Heading';
 import NegativeSideEffectsSummaryTable from '../../NegativeSideEffectsSummary/Table';
 import NegativeSideEffectsSummaryHeading from '../../NegativeSideEffectsSummary/Heading';
+import UntestableAssertionsSummaryTable from '../../UntestableAssertionsSummary/Table';
+import UntestableAssertionsSummaryHeading from '../../UntestableAssertionsSummary/Heading';
 import styles from './CandidateTestPlanRun.module.css';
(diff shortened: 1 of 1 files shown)


jugglinmike:
@stalgiag & @ChrisC: this is ready for review!

I've rebased this branch over the latest commit in `development` (since there are no reviews in progress) and updated according to the latest design feedback in gh-1429. ([The previous version of this branch is available here](https://github.com/w3c/aria-at-app/tree/gh-1429-revise-untestable-orig), in case that's of interest to anyone.) There is still an open question there, but because it reflects only a small cosmetic detail, I'm proposing we address it with a follow-up.

This changeset includes a revision to [an assumption we made in an earlier patch](https://github.com/w3c/aria-at-app/pull/1405#discussion_r2110296391). It would be cleaner to address that revision in a separate pull request, but I've taken it on here so as not to further delay this work. (As a middle ground, [I've structured those changes into a dedicated commit](https://github.com/w3c/aria-at-app/pull/1440/commits/8560b5e820a623e98212325f191028881ceae2b1).)

client/utils/summarizeScenario.js
stalgiag, line 1: Welcome improvement!

client/hooks/useFailingAssertions.js
```diff
@@ -48,10 +31,12 @@ export const useFailingAssertions = testPlanReport => {
               }));
           };
 
-          const assertionResults = [
-            ...processPriorityAssertionResults('MUST'),
-            ...processPriorityAssertionResults('SHOULD')
-          ];
+          const assertionResults = scenarioResult.untestable
```
stalgiag, line 34: elegant!

stalgiag (approved):
Tested locally and this satisfies my understanding of the requirements for now. The code is great and clean. Great job finding all of the necessary related changes.

I agree with handling the remaining design unknown in a follow-up.

Issue #1417, opened 2025-06-02
Improvements to the Automated Reports Update Test Queue


ccanash:
These are a series of improvements to the feature implemented in #1162: 

**How to manage the reports**
- [ ] After an automated reports is generated from the Automated Reports Update page the test reports that have the same responses as previous version will be automatically published.
- [ ] After an automated reports is generated from the Automated Reports Update page the test reports that have one or more AT responses that differ from the response in the previous report will be moved to the Manual Test Queue page. This is because humans will need to review the test answers. 

**How to display the reports that need human review in the manual test queue**

- [ ] In the status column show “Automated update 100% complete with X different responses generating Y conflicts”. 
- [ ] Display the manage bot run button so we that the report it can be assigned to a human.
- [ ] (lower priority) Add a link for the  “y conflicts”  that is similar to the existing conflicts feature. Here the conflict is between the tester who generated the prior report with the older AT version and the bot that generated the newer report with the newer AT version. For this to be understandable, the feature in #1400 to add an AT version column to the response table on the conflicts page should have already been implemented

ccanash:
After testing in staging, we have agreed that the next steps are: 

1. Adding a filtered view to the test queue for all users. The filtering options will be:  
     - All test runs (number)
     - Manual test runs (number) – this would be the default
     - Automated updates with run failures (number). This would only be visible if there are automated updates results available for review. 
     
2. The bot will assign an "Unknown" value to verdicts corresponding to test outputs that don't match the most recent finalized runs. This value will not be visible in the front end. 

4. In the test queue page. When there are missing assertion verdicts after an automated run:      
    - Add a labeled error icon to the start of a heading for commands where the output differed from prior output. It could be labeled “Conflicting Output”. So, a heading like “After 'Space' (browse mode on)” would be announced as “Conflicting Output image, After 'Space' (browse mode on)”:”
    - Add an error state to the text box for the output with an error message referenced by aria-error message. Error message: “This output differs from output reported for AT_NAME Version_NUMBER”

4. Adding the ability to compare the output between the bot run and the most recent finalized run prior to assigning assertion verdicts by: 
    - Giving reports a unique “rerun” state
    - Displaying the previous output in a a blockquote element like this: Output recorded for AT_NAME AT_VERSION on REPORT_DATE: < blockquote > output < /blockquote >. This text will be placed above the output recorded by the bot.

PR #1441, opened 2025-06-20
fix: improve feedback on results of automated report update

stalgiag:
addresses #1417 

## Overview

This PR addresses issues with the automated report update that made it difficult to determine what deviations existed with previous runs. The previous implementation did not "submit" results unless all outputs matched all historical output. This meant that there was no quick way to determine which results differed. This PR changes a few features to make this feedback more clear:
1. Each individual test result that has matching output with the previous report will be submitted
2. Any test that has differing output in a scenario will not be submitted and the test navigator for automated jobs now distinguishes between jobs that have been submitted and jobs that have not.
3. Add update to the status table on the automated report updates tab that communicates whether an update was completed with or without differences.
4. Keep the "Manage Bot" dialog available after a job has completed to make reassigning options more clear

There is one requirement from the linked issue that I did not address, namely the request to list conflicts for update runs. In cases where the output does not match, no verdicts are assigned, therefore there are not yet "conflicts" as they are understood in the app. A human reviewer must review the new, different output and complete the assertion verdicts. At this point, the current conflicts implementation should understand any deviation as conflicts.

## Testing

Testing this can be tricky. I have included a guide below to help reviewers. Obviously varied testing is helpful so feel free to disregard this guide once you understand the DB state needed for testing.

### Testing scenarios with deviation

For this I recommend running the actual automated runs. This requires [setting up a ngrok FQDN as per the deploy README.](https://github.com/microsoft/Consent-Package/pull/14)

Start by restoring the prod dump from the automated runs done previously. This is called `production_dump_20250603.sql` and shared internally.

Note that you will need to temporarily [delete this migration](https://github.com/w3c/aria-at-app/blob/development/server/migrations/20250603160126-removeAutomatedReportRuns.js) while running the migration setup step, otherwise the migration will remove the relevant reports.

Once that is complete, you can go to the Test Queue and delete the latest automated "Action Menu Button Example Using aria-activedescendant" report. Refresh and navigate to the "Automated Report Updates" tab. You should see the option to "Start Generating Reports" under NVDA Bot 2024.4.1.

After the job finishes running (~15 minutes), you should see an updated status in the Update Events table that indicates there was 1 difference in the run. When you view the test plan run, the test result for test 3 will show as unsubmitted. This is due to a difference in the third output.

### Testing scenarios without deviation

For this, I recommend starting with a standard preseeded DB via the population scripts and *not* connecting to GH Actions. This will result in the local development simulator automatically putting all of the same output for each test. You can then create a test run with NVDA 2024.1, add NVDA 2024.4.1 to the AT Versions, rerun and see the report automatically finalized.

### Screenshots

As mentioned previously, there was no visual or auditory indication when an update rerun had test outputs that did not match the previous report. Changes were made to address this. I am including screenshots below:

<details>
<summary>Before Screenshots</summary>
Test run view, test 3 is selected and shows as unsubmitted in the main content panel with the output and radio buttons for assertion verdicts. Note that all other tests would show as unsubmitted as well. The nav along the lefthand side has a green checkmark for all tests.

</details>

<details>
<summary>After Screenshots</summary>
Test run view, test 1 is selected and shows as submitted in the main content panel. The nav along the left hand side has a greencheck mark next to the nav link for test 1.

Test run view, test 3 is selected and shows as unsubmitted in the main content panel with the output and radio buttons for assertion verdicts. The nav along the left hand side has a white circle in a slightly paler green circle next to the nav link for test 3.

</details>

### DIFF
--- a/server/controllers/AutomationController.js
+++ b/server/controllers/AutomationController.js
@@ -483,8 +483,6 @@ const finalizeTestPlanReportIfAllTestsMatchHistoricalResults = async ({
         tr => String(tr.testId) === String(test.id)
       );
 
-      if (!currResult) continue;
-
       const histResult = historicalResults.find(
         hr => String(hr.testId) === String(test.id)
       );


ChrisC:
@stalgiag It would be helpful if you could include some screenshots of the UI updates whenever you get a chance!

stalgiag:
> @stalgiag It would be helpful if you could include some screenshots of the UI updates whenever you get a chance!

Good suggestion! Comment updated.

server/controllers/AutomationController.js
```diff
@@ -448,43 +474,74 @@ const finalizeTestPlanReportIfAllTestsMatchHistoricalResults = async ({
...
+    let differentResponsesCount = 0;
+    let allOutputsMatch = true;
+
     // Validate each applicable test's results against historical results
     for (const test of applicableTests) {
       const currResult = testPlanRun.testResults.find(
         tr => String(tr.testId) === String(test.id)
       );
+
+      if (!currResult) continue;
```
jugglinmike, line 486: Don't we want to set `allOutputsMatch = false` in this case? And if so maybe we can consider this branch an early optimization and just rely on the condition that follows the calculation of `histResult`.

jugglinmike (approved):
Code looks good and functions as intended. Thanks for the help in getting this running on my machine, Stalgia!

server/controllers/AutomationController.js
stalgiag: Good callout! Addressed in [e173a1f](https://github.com/w3c/aria-at-app/pull/1441/commits/e173a1febd328013a7c63765071404bff875ae68)

PR #1443, opened 2025-06-25
fix: tableRowMinimumAtVersions[index] is undefined

### LINKED ISSUE #1442: Bug: tableRowMinimumAtVersions[index] is undefined
When creating new reports and adding AT Versions, The `tableRowMinimumAtVersions` override and the `minimumAtVersions` stored in the `sortedStatuses` can get out of sync causing the `tableRowMinimumAtVersions[index] is undefined` error.

stalgiag:
resolves #1442 

Previously the overrides and the status properties were not being synced appropriately.


jugglinmike (approved):
This works for me! Thanks for the expedited diagnosis and fix!

Issue #1430, opened 2025-06-12
Improvement to the  “unexpected behaviors” to “negative side effects” feature


ccanash:
These are improvements related to #1352. 

- [ ] Add instruction to ensure the testers know that they should check at least one side effect is set to 'severe' for the command.
- [ ] Modify text in the checkbox. Currently for a screen reader user "The response to 'b'..." can sound like "the response to be...". To avoid this phonetic confusion. Please replace with "The [insert name of the screen reader, like JAWS] response to the command 'b' created conditions that made assertions untestable...". 
- [ ] When the submission error for side effect severity appears, the focus should be programmatically moved to the impact severity list itself, not the checkbox. This would help the user immediately locate and correct the issue. If only one side effect is checked, we should move focus to the impact picker, and it needs to have  an associated error message. If two side effects are checked it should go to the first side effects impact list, like it does for "Speech output" fields.

ccanash:
Confirming I have seen this and will incorporate it into our planning.

PR #1446, opened 2025-07-01
fix: Improve form UI for untestable assertions

jugglinmike:
Addresses gh-1430. Depend on https://github.com/w3c/aria-at/pull/1264

I built my own tooltip. This statement ought to concern any reviewer, so here's my justification:

1. The three most popular React-compatible tooltip libraries do not suit our purposes:
   - [Tooltip from the "React Component" library](https://react-component.github.io/tooltip/) is not accessible
   - [Tooltip in Material UI](https://mui.com/material-ui/react-tooltip/) appears to be inseparable from Material UI itself, which is a very large dependency
   - [React Tooltip](https://react-tooltip.com/) is not accessible (it is documented as such, but [the implementation has a flaw which subverts the user experience](https://github.com/react-component/tooltip/issues/496))
2. The application is minimal (it entailing the display of a single element whose only content is text nodes)
3. The implementation is small and requires no JavaScript

But of course, I'm still open to trying a more robust solution!

### DIFF
--- a/client/components/TestRenderer/TestRenderer.module.css
+++ b/client/components/TestRenderer/TestRenderer.module.css
@@ -172,21 +172,6 @@ label {
     input {
       margin-right: var(--4px);
     }
-
-    &:hover [role='tooltip'],
-    input:focus + [role='tooltip'] {
-      visibility: visible;
-    }
-
-    [role='tooltip'] {
-      position: absolute;
-      top: 100%;
-
-      background-color: rgba(255, 255, 255, 0.9);
-      padding: 1em;
-      border-color: var(--border-gray);
-      visibility: hidden;
-    }
   }
 
   &.assertions-label {
--- a/client/components/TestRenderer/TestRenderer.module.css
+++ b/client/components/TestRenderer/TestRenderer.module.css
@@ -184,7 +184,7 @@ label {
 
       background-color: rgba(255, 255, 255, 0.9);
       padding: 1em;
-      border-color: #999;
+      border-color: var(--border-gray);
       visibility: hidden;
     }
   }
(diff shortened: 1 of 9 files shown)


client/components/TestRenderer/TestRenderer.module.css
```diff
@@ -167,10 +167,26 @@ label {
...
+      visibility: visible;
+    }
+
+    [role='tooltip'] {
+      position: absolute;
+      top: 100%;
+
+      background-color: rgba(255, 255, 255, 0.9);
+      padding: 1em;
+      border-color: #999;
```
howard-e, line 187:
```suggestion
      border-color: var(--border-gray);
```
^ for consistency or if you'd really prefer that color, you could define it in the `index.css`.

Missing a border style here as well to show this

howard-e (commented):
Looks good and I understand the motivation behind this custom approach!

> 3. The implementation is small and **requires no JavaScript**

We may want to add a tiny bit of JS to support <kbd>Escape</kbd> closing the tooltip

howard-e (approved):
> > > 3. The implementation is small and **requires no JavaScript**
> > 
> > 
> > You may want to add a tiny bit of JS to support Escape closing the tooltip
> 
> @howard-e I think adding a tiny bit of JS will unfortunately complicate this solution beyond the point we want to maintain ourselves.
> 
> As suggested by the team, I've introduced [Shadcn's Toolip](https://ui.shadcn.com/docs/components/tooltip). I'm not thrilled by the amount of glue code here--even when (mostly!) copied from the documentation--but it seems necessary given that we're not interested in importing an entire UI framework.

Changes look good to me and the change of direction is fine. Ideally, we don't want to plug these things but this works as is and should continue to without much development overhead

> It may be easier to review the latest commit with Git's `-w`/`--ignore-all-space`, which can be enabled in the GitHub review UI by [appending `?w=1` to the relevant URL](https://github.com/w3c/aria-at-app/pull/1446/commits/45d831a1ab5d750af79944b63faed8f28095220b?w=1).

Had no clue, thanks for sharing!

Will merge as soon as the build passes.

PR #1451, opened 2025-07-08
ADB proxies

stalgiag:
### Overview 

This PR adds an adb proxy build system that outputs self-contained adb proxies for Windows, Linux, and macOS. Each portable output contains everything needed to run adb without installing Node or Android Platform Tools. This includes but not limited to using:
- Using `localtunnel` to expose the user's local proxy to the rest of the internet for the websocket to use. (optionally, can build using `ngrok` as well (could remove in future iterations, left in as proof-of-concept)

### Changes
- ADB command whitelisting
- Replaced direct shell script execution with HTTP calls to adb-proxy
- Added server sent events endpoint
- `EventEmitter`-based utterance capturer
- Adding a "copy to clipboard" button to the Capture Utterances Panel on the Test Run page when user is capturing utterances from their Android device


client/static/index.css
```diff
@@ -97,6 +97,14 @@
...
   --candidate-approved: #309d08;
 
+  --success-green: #155724;
+  --success-green-bg: #d4edda;
+  --success-green-border-color: #c3e6cb;
+
+  --error-red: #721c24;
+  --error-red-bg: #f8d7da;
+  --error-red-border-color: #f5c6cb;
+
```
stalgiag, line 107: These feels repetitive. I wonder if existing variables could be the baseline for green and red? I just want to promote a limited palette.

client/static/index.css
```diff
@@ -97,6 +97,14 @@
   --candidate-in-progress: #2560ab;
   --candidate-approved: #309d08;
 
+  --success-green: #155724;
+  --success-green-bg: #d4edda;
+  --success-green-border-color: #c3e6cb;
+
+  --error-red: #721c24;
+  --error-red-bg: #f8d7da;
+  --error-red-border-color: #f5c6cb;
```
stalgiag, line 106: These feels repetitive. I wonder if existing variables could be the baseline for green and red? I just want to promote a limited palette.

scripts/get-network-ip.js
stalgiag, line 1: I think this can be deleted.

server/controllers/ScriptController.js
```diff
@@ -1,81 +1,269 @@
...
+      return res.status(400).json({
+        error:
+          'No Android device connected. Please connect a device with USB debugging enabled.'
+      });
     }
-    if (stderr) console.error(`Script stderr: ${stderr}`);
 
-    res.json({ success: true, output: stdout });
-  });
+    await runAdbCommand('shell input keyevent KEYCODE_WAKEUP', req);
```
stalgiag, line 172: Nice!

server/websocket.js
```diff
@@ -61,98 +24,202 @@ const setupWebSocketServer = server => {
...
+                Accept: 'text/event-stream',
+                'Cache-Control': 'no-cache'
+              }
+            };
+
+            // Use https for ngrok or similar URLs, http for localhost
+            const requestModule = proxyUrl.protocol === 'https:' ? https : http;
+
+            const req = requestModule.request(options, response => {
+              console.info(
```
stalgiag, line 89: We probably want to make a lot of logging added in this PR configurable. This could be a follow up

stalgiag (commented):
This is great! Some small comments inline

server/controllers/ScriptController.js
howard-e: Ha, having to tap the screen repeatedly

server/websocket.js
howard-e: Good thought!

jugglinmike:
On the sandbox server today, the section titled "Android Device Testing" is rendered for every Test Plan  Run--not just the ones which designate Talkback as the AT under test.

howard-e:
> On the sandbox server today, the section titled "Android Device Testing" is rendered for every Test Plan Run--not just the ones which designate Talkback as the AT under test.

Good catch! Done in https://github.com/w3c/aria-at-app/pull/1451/commits/6ae4450d9d3268538ae56a798f335b87f9e11fa5

howard-e:
Will seek to handle remainder of items in `talkback-prototyping` branch

PR #1455, opened 2025-07-10
Specially handle which workflow is selected when running VoiceOver Bot on macOS 15

howard-e:
See title

### DIFF
--- a/server/services/GithubWorkflowService.js
+++ b/server/services/GithubWorkflowService.js
@@ -135,13 +135,10 @@ const getMacOSVersionName = atVersion => {
  * @param {object} atVersion
  * @param {string} atVersion.name
  */
-const uniqueWorkflowByAtVersion = (atKey, atVersion) => {
-  if (atKey === 'voiceover_macos') {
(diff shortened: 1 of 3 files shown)


server/services/GithubWorkflowService.js
```diff
@@ -118,6 +118,32 @@ const fetchInstallationAccessToken = async (jsonWebToken, installationID) => {
...
+  return atVersion?.name?.split('.')[0];
+};
+
+/**
+ * @param {string} atKey
+ * @param {object} atVersion
+ * @param {string} atVersion.name
+ */
+const uniqueWorkflowByAtVersion = (atKey, atVersion) => {
+  if (atKey === 'voiceover_macos') {
```
ChrisC, line 139: Just to simplify the logic a bit here, we probably don't need to handle any other `atKeys` at the moment... but I can see how this might be useful in the future!

howard-e: Yep that was the thought but I can simplify!

ChrisC (approved):

PR #1456, opened 2025-07-10
feat: Add support for VoiceOver on macOS 15 with VoiceOver Bot

howard-e:


server/services/GithubWorkflowService.js
```diff
@@ -174,7 +194,7 @@ const createGithubWorkflow = async ({ job, directory, gitSha, atVersion }) => {
       'X-GitHub-Api-Version': '2022-11-28'
     },
     data: JSON.stringify({
-      ref: 'main',
+      ref: 'self-hosted',
```
howard-e, line 197: Do you want to revert this before pushing to staging @ChrisC?

ChrisC: We can keep this on `staging` but should def revert in a separate PR before pushing to prod!

howard-e: Cool submitted #1457 to track that

PR #1459, opened 2025-07-15
feat: Link rerun reports to updated historical report

stalgiag:
This PR adds a new property in `TestPlanReport` that connects any report created with the rerun feature to the historical report that it updates. This enables additional UI feedback on differences in output between a rerun report and the report it is updating. 

Notes:
- Testing is easiest if you used the actual workflows and a prod dump of the database (available in internal channels)
- Apologies for the rearrangement in the `TestQueue`. It was necessary to memoize
- There is also a regrettable amount of prop drilling with `historicalReport` values. Open to alternatives but this seemed easiest for the time being.

### DIFF
--- a/client/components/ReportRerun/RerunDashboard.jsx
+++ b/client/components/ReportRerun/RerunDashboard.jsx
@@ -79,11 +79,6 @@ const RerunDashboard = ({ activeRuns, onRerunClick }) => {
                         : 'earlier versions'}{' '}
                       of {run.botName} can be automatically updated with the
                       &quot;Start Generating Reports&quot; button below.
-                      {/* TODO: Remove below when https://github.com/w3c/aria-at-app/issues/1417 is addressed */}
-                      <br />
-                      <br />
-                      Note: The button is temporarily disabled while we work on
-                      upcoming feature updates.
                     </p>
                   </div>
 
@@ -91,7 +86,7 @@ const RerunDashboard = ({ activeRuns, onRerunClick }) => {
                     {/* TODO: Re-enable when https://github.com/w3c/aria-at-app/issues/1417 is addressed */}
                     <button
                       className={styles.rerunButton}
-                      disabled={true /* totalReports === 0 */}
+                      disabled={totalReports === 0}
                       onClick={() => onRerunClick(run)}
                       aria-label={buttonAriaLabel}
                     >
--- a/client/components/TestQueue/index.jsx
+++ b/client/components/TestQueue/index.jsx
@@ -1,4 +1,4 @@
-import React, { Fragment, useRef, useState } from 'react';
+import React, { Fragment, useRef, useState, useMemo } from 'react';
 import { useApolloClient, useQuery } from '@apollo/client';
 import PageStatus from '../common/PageStatus';
 import { TEST_QUEUE_PAGE_QUERY } from './queries';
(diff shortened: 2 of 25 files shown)


client/components/ReportRerun/RerunDashboard.jsx
```diff
@@ -79,19 +79,14 @@ const RerunDashboard = ({ activeRuns, onRerunClick }) => {
...
-                      {/* TODO: Remove below when https://github.com/w3c/aria-at-app/issues/1417 is addressed */}
-                      <br />
-                      <br />
-                      Note: The button is temporarily disabled while we work on
-                      upcoming feature updates.
                     </p>
                   </div>
 
                   <div className={styles.actionHeader}>
                     {/* TODO: Re-enable when https://github.com/w3c/aria-at-app/issues/1417 is addressed */}
```
jugglinmike, line 86: This comment can go, too.

client/components/TestQueue/index.jsx
```diff
@@ -22,19 +22,165 @@ import Actions from './Actions';
...
+
+    const testers = data.users
+      .filter(user => user.roles.includes('TESTER'))
+      .sort((a, b) => a.username.localeCompare(b.username));
+
+    return { testPlans, testers };
+  }, [data]);
+
+  // Filter counts calculation
+  const { allCount, manualCount, automatedCount } = useMemo(() => {
```
jugglinmike, line 125: These values are incidental to the derivation of `filterOptions`, so they don't need to be memoized separately.

client/components/TestQueue/index.jsx
```diff
@@ -22,19 +22,165 @@ import Actions from './Actions';
...
+            testPlanReports: testPlanVersion.testPlanReports.filter(
+              testPlanReport => {
+                const hasBot = hasBotRun(testPlanReport);
+                const isRerun = !!testPlanReport.historicalReport;
+
+                if (activeFilter === FILTER_KEYS.MANUAL)
+                  return !hasBot || !isRerun;
+                if (activeFilter === FILTER_KEYS.AUTOMATED)
+                  return hasBot && isRerun;
+                return true;
```
jugglinmike, line 175: This statement is unreachable.

client/components/TestRenderer/CommandResults/index.jsx
```diff
@@ -31,14 +37,41 @@ const CommandResults = ({
...
   return (
     <>
-      <h3>{header}</h3>
+      <h3>
+        {hasConflictingOutput && (
+          <FontAwesomeIcon
+            icon={faExclamationTriangle}
+            style={{ color: '#ce1b4c', marginRight: '8px' }}
+            aria-label="Conflicting Output"
+          />
```
jugglinmike, line 61: Oddly, your application of `aria-label` is *not* disabling the library's built-in application of `aria-hidden="true"` (bug report...?), so we need to do that explicitly.

server/util/outputNormalization.test.js
```diff
@@ -80,7 +127,36 @@ describe('outputsMatch', () => {
...
+    expect(outputsMatch('HeLLo WoRLd', 'heLLo woRLd')).toBe(true);
+  });
+
+  it('returns true for strings with JAWS separator tokens', () => {
+    expect(
+      outputsMatch('Text\u001dSeparator\u001eMore', 'Text Separator More')
+    ).toBe(true);
+    expect(outputsMatch('Start\u001dMiddle\u001eEnd', 'Start Middle End')).toBe(
+      true
+    );
```
jugglinmike, line 159: Could we add one test case for `"Start \u001d Middle \u001e End"`? That string more closely resembles the behavior we've observed, and the test case will prevent future revisions from collapsing white space before making the substitution.

jugglinmike (changes requested):
So far, so good! Here's a bit of low-level feedback while we solve the mystery we identified privately.

client/components/ReportRerun/RerunDashboard.jsx
stalgiag: [99030f7](https://github.com/w3c/aria-at-app/pull/1459/commits/99030f7dfcd35e6815074ff1ca8ee1ef67bd5d98)

client/components/TestQueue/index.jsx
stalgiag: Good catch! I think this was a leftover from when I was using those values elsewhere for some reason. Thanks! [190af56](https://github.com/w3c/aria-at-app/pull/1459/commits/190af56386b2f4e394c9a976657b699b82471388)

server/util/outputNormalization.test.js
stalgiag: Good call! [6b85938](https://github.com/w3c/aria-at-app/pull/1459/commits/6b859389810eb3b28800376eec78afdacf99c4ea)

client/components/TestQueue/index.jsx
jugglinmike: Oh, right on!

PR #1460, opened 2025-07-17
Make staging use prod repo for self-hosted actions

stalgiag:
see title

This is needed as temporary workaround while the dev runner repo does not support the self-hosted runner.


stalgiag:
Closing in favor of fix in upstream service

PR #1461, opened 2025-07-21
count unsubmitted assertion verdicts as "assigned"

stalgiag:
see title

With assertion verdict copying, currently test results are only being submitted when all outputs match. This PR changes the calculation for "X of Y assertions assigned" to include verdicts that have been assigned but not submitted.

---
  - https://app.asana.com/0/0/1210859735312703


howard-e (approved):
Looks good to me!

Issue #1464, opened 2025-07-22
Improve how to communicate missing veredict assignments


ccanash:
In test plans, the app is not communicating effectively when there are veredict assignments missing. This is specially true for screen reader users. 

For other fields the app signals with a (required) when there is missing data required for submitting a result.

stalgiag:
Testing on VoiceOver 14.5, when one clicks on "Submit Results" on a test with missing assertion verdicts, the page moves my focus to "No" radio button in the "Did negative side effects occur?" radio group for the missing scenario verdicts. A red "required" element appears above this radio button but in order to hear it announced a screenreader user would need to navigate backwards. My sense is that there are 3 issues with the screenreader UX:

1. It isn't clear why focus was moved and the test was not submitted when attempting to submit a result without assertion verdicts.
2. It isn't clear where focus was moved when attempting submit a result without assertion verdicts.
3. Focus is moved to the "Did negative side effects occur?" section for a scenario when it should be moved to the first missing verdict.

Options for improving UX:
1. Add `aria-required=true` to the assertion fieldsets
2. Add `aria-invalid="true"` to the assertion fieldsets after attempted submission with missing verdicts.
3. Add `aria-describedby` to the assertion fieldset after attempted submissions with missing verdicts linked to the "required" error .

mcking65:
Overall, this is working better. However, there is one change that is problematic.

If you submit with a missing verdict, focus is moving to the fieldset instead of the radio. It should move to the radio.

This also has the quite problematic side effect of putting all the fieldsets in the tab sequence, which unnecessarily doubles the number of tab stops.

This doesn't have to block the August 4 deployment. However, we need to keep this issue open and complete resolution in the next deployment.

ccanash:
Hi @mcking65 , closing this as the the release last week included your latest feedback about the focus

PR #1479, opened 2025-08-04
fix: focus incomplete radios instead of fieldsets

stalgiag:
see title

This addresses the [additional feedback](https://github.com/w3c/aria-at-app/issues/1464#issuecomment-3148850012) in #1464. Originally, the fieldset was chosen as the focus point because in my testing on VO, I felt that the best orientation to the required input needed was from the top level so as not to imply that any one radio response was required, also to give a sense of the query to which the radio button is replying. Feedback disagrees with this design so this PR shifted to focusing the first radio button under the fieldset instead.


howard-e (approved):
@stalgiag changes look good to me. Thanks for handling so quickly!

Issue #1477, opened 2025-08-04
Test status missing from test runner navigation


mcking65:
In the navigation bar of the test runner that lists all the tests, there was previously an icon that let you know if a test was complete or has conflicts. It is no longer present. Or, if it is present, it is not accessible.

PR #1495, opened 2025-08-14
fix: Missing test status in test navigator

howard-e:
Address #1477


stalgiag (approved):
Tested with Voiceover 15 and confirmed that this now announces the status of the test. Thanks Howard!

PR #1498, opened 2025-08-19
Update internal terminology to match UI

jugglinmike:
Rename "unexpected behaviors" and "undesirable behaviors" to "negative side effects".

This patch depends on a corresponding change to the ARIA-AT project. I'll share a link to that pull request once it's available.


howard-e (commented):
Changes look good! Thanks for doing this and cleaning things up further with the migration!

There are a few instances of "undesirable" inferring "undesirable behavior(s)" that could also be cleaned up. Even though that wasn't the target string, I figure it would be easy to just get these last remaining ones. What do you think?
* [L68 in `useTestRendererFocus.js`](https://github.com/w3c/aria-at-app/blob/cb7cc8456b9512d1727296c40454d495fdb6d670/client/hooks/useTestRendererFocus.js#L68)
* [L695 in `TestRun.e2e.test.js`](https://github.com/w3c/aria-at-app/blob/cb7cc8456b9512d1727296c40454d495fdb6d670/client/tests/e2e/TestRun.e2e.test.js#L695)
* [L114 in `NegativeSideEffectsFieldset/index.jsx`](https://github.com/w3c/aria-at-app/blob/cb7cc8456b9512d1727296c40454d495fdb6d670/client/components/TestRenderer/NegativeSideEffectsFieldset/index.jsx#L114)
* [L169 in `NegativeSideEffectsFieldset/index.jsx`](https://github.com/w3c/aria-at-app/blob/cb7cc8456b9512d1727296c40454d495fdb6d670/client/components/TestRenderer/NegativeSideEffectsFieldset/index.jsx#L169)
* _& the follow-up snapshot update_

jugglinmike:
Good catch, @howard-e! That sounds good to me.

I replaced `undesirable-` with `unexpected-behavior-` rather than simply `unexpected-` because the former is more descriptive, easier to `grep`, and less likely to be misinterpreted by human readers when concatenated with other values (e.g. `unexpected-behavior-output-is-excessively-verbose` instead of `unexpected-output-is-excessively-verbose`).

howard-e (approved):
Changes look good! Thanks for handling this extensive tech debt cleanup and for addressing my last bit of feedback. Thanks!

Feel free to merge after reverting the change to the [aria-at import branch](https://github.com/w3c/aria-at-app/pull/1498#discussion_r2286088677).

Issue #1377, opened 2025-04-17
Investigate duplicate test in test Queue


ccanash:
In the [test queue page](https://aria-at.w3.org/test-queue) there were two tests for **Color Viewer Slider** that had been created for the same JAWS version (JAWS 2025.2503.39 or later) for Chrome (any version).

We shouldn't be able to have two tests for the same version.

PR #1499, opened 2025-08-20
Prevent duplicate report entries on the Test Queue

howard-e:
Address #1377


ChrisC (commented):
@howard-e I haven't pulled this down to test yet but looking over the code, it seems like this will just return an existing test plan run if one already exists for the matching AT version. Is that correct? I think it might be a more useful experience for the user if we block the process and instead let the user know that the test plan run for that version already exists, and that they can't add again. 

It would also be great if there were smaller localized tests here, either at the service level on the server side or on the component side. ([Looks like there's already some component tests with mocks here](https://github.com/w3c/aria-at-app/blob/development/client/tests/AddTestToQueueWithConfirmation.test.jsx#L2)...)

ChrisC:
> Good thought! I can add this approach to the PR. Was initially thinking preventing on the gql layer is helpful to prevent any chance of it happening through a query (accidental or otherwise).
> 
I like that. It seems if we could prevent on the gql layer then maybe we could even prevent the AT version even showing up in the dropdown as an option to add to the test queue, if that version is already in the queue?

howard-e:
Agreed internally that the follow up points here can be done in a separate enhancement task. Thanks @ChrisC!

Issue #1485, opened 2025-08-11
Change to normalization in the context of historical comparison


ccanash:
During the [Automation cal](https://www.w3.org/2025/08/11-aria-at-minutes.html)l on August 11th we discussed the following updates to normalization:

1. Preserve all punctuation when comparing bot results to historical results.
2. Lower-case everything when comparing bot results to historical results.

PR #1501, opened 2025-08-20
updates punctuation and caps normalization rules

ChrisC:
Addresses #1485 

(Note that this still preserves the normalization rules for trimming whitespace around hyphens)

### DIFF
--- a/server/util/outputNormalization.test.js
+++ b/server/util/outputNormalization.test.js
@@ -39,18 +39,16 @@ describe('normalizeScreenreaderOutput', () => {
     );
   });
 
-  it('normalizes capitalization for first letters only', () => {
-    expect(normalizeScreenreaderOutput('Hello World')).toBe('hello world');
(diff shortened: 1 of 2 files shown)


server/util/outputNormalization.test.js
```diff
@@ -212,6 +201,6 @@ describe('outputsMatch', () => {
     const output2 =
       'W3C website link\nYou are currently on a link. To click this link, press Control-Option-Space.';
 
-    expect(outputsMatch(output1, output2)).toBe(true);
+    expect(outputsMatch(output1, output2)).toBe(false);
```
ChrisC, line 204: @stalgiag Given the new requirements about keeping punctuation, I think this example should actually fail. I can see how this specific example (`, W3C website link ...`) could be interpreted as possibly truncated output and imo should warrant some additional manual review from a human tester.

PR #1503, opened 2025-08-26
fix: Enforce ownership of deployed project in sandbox environment

howard-e:
This PR enforces deployed project directory (and contents) as being owned by the defined ansible `application_user`

Attempting to solve an issue which caused clashes when deploying to the sandbox when another user did the deploy. Previously, if `userA` did the deploy and then `userB` follows, there would be permission issues with the `Perform source code synchronization` step. The ansible playbook is generally expecting the user to be `{{application_user}}`

Getting through this previously looked like:

> 1. ssh'ing into sandbox
> 2. Running `chown -R myuser:staff /path/to/project` to claim ownership if updated by someone else
> 3. Re-running the ansible-playbook deployment to sandbox

NOTE: This doesn't affect the staging and production environments.

### DIFF
--- a/client/components/CandidateReview/CandidateReview.module.css
+++ b/client/components/CandidateReview/CandidateReview.module.css
@@ -25,6 +25,7 @@ span.cell-row {
   }
 }
 
+a.status-text,
 span.status-text {
   height: var(--24px);
   font-size: var(--14px);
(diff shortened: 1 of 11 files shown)


jugglinmike:
The "synchronize" task is being executed with the `become` directive, so I wouldn't expect the deploying user to have any bearing here. Any thoughts about why it does?

howard-e:
> The "synchronize" task is being executed with the `become` directive, so I wouldn't expect the deploying user to have any bearing here. Any thoughts about why it does?

Good question. And it _should_. Will investigate further why that is

PR #1510, opened 2025-09-04
feat: use pinned aria-at commit with cron to update 

stalgiag:
This PR does the following

1. Adds a pinned aria-at commit SHA to be used when no commit arg is provided to the import-tests script
2. Adds a workflow scheduled to run daily that does the following
  1. Check to see if the pinned commit SHA is the latest, if so exit
  2. Grabs the latest commit SHA, runs import tests using it
  3. Runs all tests *except* the snapshot tests
  4. If above is successful, updates the snapshots
  5. Pushes a commit to `development` with the updated pinned SHA and snapshots
  
This would resolve many issues with snapshots and it would also catch upstream breaking changes.

Leaving in DRAFT until the overall approach is agreed upon. There is likely some additional consideration needed for how this is handled in deployed environments.


docs/local-development.md
```diff
@@ -17,12 +17,26 @@
...
+   `yarn dev`
    Now you can navigate your browser to: [http://localhost:3000/](http://localhost:3000/). You need to use localhost instead of `0.0.0.0` because the cookie needs to be treated as secure.
 
+## Pinned ARIA-AT version
+
+The app imports test plans from the `w3c/aria-at` repository using a pinned commit.
+
+- The pinned SHA lives at `config/aria-at.version`.
+- Local imports without `-c` will read this file (or `ARIA_AT_PINNED_SHA`).
+- CI imports historical commits for coverage and then the pinned commit.
```
ChrisC, line 32: This is an aside, but it's still unclear to me why exactly we need those historical commits (and now in addition to the pinned commit)... Is this some tech debt that we could also clean up before next year? cc @howard-e

docs/local-development.md
```diff
@@ -6,7 +6,7 @@
    - Version 14 or greater
    - It is recommended to install node with [`nvm`](https://github.com/nvm-sh/nvm)
 2. Yarn
-   - Yarn is resposible for installing dependencies, similar to npm. This project is utilizing yarn workspaces to organize the code into a monorepo structure.
+   - Yarn is responsible for installing dependencies, similar to npm. This project is utilizing yarn workspaces to organize the code into a monorepo structure.
```
howard-e, line 9: 👍🏽

server/scripts/import-tests/index.js
howard-e, line 1: Wouldn't this change prevent the deployed environments from getting the latest version changes from aria-at? The `aria-at.version` is effectively static there.

Seems a cron in the deployed environment too or some exclusion of `aria-at.version` in those environments' deploys with the fail condition relaxed is needed here.

howard-e (commented):
Thanks for pushing the work forward on this! Left comments below concerning how this operates in the deployed environments

server/scripts/import-tests/index.js
stalgiag: Curious if this meets your concerns [5c44b6e](https://github.com/w3c/aria-at-app/pull/1510/commits/5c44b6e08625a1ca5a00bfaa6382baccce04a745)? It is difficult to test this but I think this should work

stalgiag: I *think* that this [default path should work for enforcing the pinned commit on initial deploy](https://github.com/w3c/aria-at-app/pull/1510/files#diff-685752680d2f8de53e17916822fc606b89216fc07d726d3d7029ad8cf32f261aR50) but admittedly Ansible is unfamiliar to me

howard-e (changes requested):

server/scripts/import-tests/index.js
howard-e: Yep, these changes look good! After resolving the merge conflicts and reviewing my latest comment, we'd be good to move this forward

PR #1514, opened 2025-09-10
Update automation bot versions (NVDA and JAWS)

gnarf:
I've added both of these versions manually to live and staging servers.


howard-e (approved):
Thanks!

PR #1548, opened 2025-09-30
Fix linters + commit hook

howard-e:
This PR does the following:
* Addresses an issue where after an import, the `commandsV2.json` would show as a change due to formatting discrepancies between the `lint-staged` step and formatters (`prettier` and `eslint`) not accounting for `*.json` files. It was always reverted on the commit step but was a point of annoyance during local development. So this change updates the scripts for the linting in the root `package.json` and updates the `lint-staged` command.
* Centralizes linter dependencies to the root package.json.
* Noticed there is some lag in the project install coming from the storybook dependencies. While it's not the only one, it's been abandoned for some time, so I went ahead and removed all storybook related dependencies and files. For reviewers, if you'd rather this not be included in this change let me know.


stalgiag (approved):
This one has been bugging me for a while. Thanks for centralizes those steps! Also happy to have storybook removed.

PR #1553, opened 2025-10-01
fix: expanded status polling

stalgiag:
The number of responses completed and percentage complete for a report was not updating as collection jobs ran. This was due to limited polling.

This PR adds polling for those values.

As a future enhancement, I would advocate for a subscription with a centralized store to reduce polling.


howard-e (approved):
@stalgiag changes look good, thanks!

> Note: As a future enhancement, I would advocate for a subscription with a centralized store to reduce polling.

Huge +1 on that

PR #1583, opened 2025-10-20
feat: ARIA/HTML support tables

gnarf:
Building out the front end on top of #1581

I was able to get the main page tables implemented, though never got to trying to put them in a "tabbed" interface, and wasn't sure what to put into the description paragraphs.

I started working on the individual report page, but discovered that the data it needs isn't being exported yet on the backend PR either. (see https://github.com/w3c/aria-at-app/pull/1581#issuecomment-3420476738)

Have left an empty "mark" commit to help with rebasing if needed, and you can view only the changes on top here: https://github.com/w3c/aria-at-app/compare/aria-html-features-be...aria-html-features-fe


client/components/Reports/Reports.jsx
```diff
@@ -23,24 +28,36 @@ const Reports = () => {
...
+        <Helmet>
+          <title>Loading - AT Interop Reports | ARIA-AT</title>
+        </Helmet>
+        <KeyMetricsBanner />
+        <section className={clsx(reportsStyles.contentSection)}>
+          <h1>Assistive Technology Interoperability Reports</h1>
+          <p>Loading...</p>
+        </section>
+      </Container>
     );
```
stalgiag: I separated the `KeyMetricsBanner` from report summary to allow the same "slightly wider than the body" look of the home page

client/components/Reports/AriaSupportTable.jsx
```diff
@@ -42,15 +46,25 @@ const AriaSupportTable = ({ ariaHtmlFeaturesMetrics }) => {
...
-              {supportCombos.map(col => (
-                <td key={key + col}>
-                  <ProgressBar
-                    progress={dataByFeature[key][col].passedPercentage}
-                  />
-                </td>
-              ))}
+              {supportCombos.map(col => {
+                const row = dataByFeature[key][col];
+                const ids = atBrowserLinks[key][col];
```
gnarf: Not a huge deal, but I don't think you need `atBrowerLinks` as it just copies values from `row` which you get from `dataByFeature` here

client/components/Reports/AriaSupportTable.jsx
```diff
@@ -6,28 +6,32 @@ import { Table } from 'react-bootstrap';
 
 const columnName = ({ atName, browserName }) => `${atName} and ${browserName}`;
 
-const AriaSupportTable = ({ ariaHtmlFeaturesMetrics }) => {
-  console.log(ariaHtmlFeaturesMetrics);
+const AriaSupportTable = ({ ariaHtmlFeaturesMetrics, headingText }) => {
```
gnarf: Not encouraging it too hard, but if you think it would clean up duplication - now that you added `headingText` I think you could pass only `ariaFeaturesByAtBrowser` or `htmlFeaturesByAtBrowser` to a `dataRows` param here and possibly delete the other `HtmlSupportTable`

ChrisC (approved):
If you have time to address @gnarf's non-blocking suggestions above that would be great, otherwise I think we can merge this to staging!

PR #1597, opened 2025-10-29
fix: browser correction for edge support on automation

gnarf:
It wasn't finding `MicrosoftEdge` browser with the old query.

This resulted in "empty" result values in the report.

This fixes the problem on my local:
image


howard-e (approved):
Thanks @gnarf! This looks good to me. Makes sense considering [this change](https://github.com/w3c/aria-at-app/pull/1587/files#diff-7355b27863ecfde868462935fe55b7369f6386d93327ed0a8f7f29777aec45b4) now

Issue #1519, opened 2025-09-11
Create a report  by ARIA/HTML feature


ccanash:
## Top-Level Report Page Updates
On the [reports page](https://aria-at.w3.org/reports), after the introduction and before the L2 heading with text “Support Levels”, add a tab list with three tabs: “Test Plans”, “ARIA Features”, and "HTML Features."

The L2 heading that currently has content “Support Levels” will serve as a tab title and have content of either: “Test Plan Support Levels,” “ARIA Feature Support Levels,” or "HTML Feature Support Levels".

### Table Summary by AT/browser combination
**ARIA Summary Table example:**
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-6f4dbd62-7fff-8e35-cfa7-9ff896d27951"><div dir="ltr" style="margin-left:0pt;" align="left">
ARIA Feature | JAWS and Chrome | NVDA and Chrome | VoiceOver for macOS and Safari
-- | -- | -- | --
aria-selected | X% of passing | X% of passing | X% of passing
button | X% of passing | X% of passing | X% of passing

</div></b>

**HTML Summary Table example:**
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-6f4dbd62-7fff-8e35-cfa7-9ff896d27951"><div dir="ltr" style="margin-left:0pt;" align="left">
HTML Feature | JAWS and Chrome | NVDA and Chrome | VoiceOver for macOS and Safari
-- | -- | -- | --
Button| X% of passing | X% of passing | X% of passing

</div></b>

#### Percentage and value definitions for AT/Browser Combo Support Level columns
Each percentage is a combination of  must-have and should-have behaviors that should be combined into a single total assertion count (X% of passing  = P/N*100):
1. N: Total number of **“Must-Have” and "Should-Have" assertions** that reference this attribute across all most-recent reports on this AT/browser combination where most-recent is:
  a. Report generated from most recent version of recommended test plan if available.
  b. If a recommended test plan report is not available, report generated from most recent version of candidate test plan if available.
  c. Data from draft test plans should not be included in attribute reports.
2. P: Number of N assertions with a passing verdict.
3. F: Number of N assertions with a failing verdict.
4. U: Number of N assertions that were untestable. 

#### Additional Summary Table Features
1. The attributes link to specifications using [data in support.json](https://github.com/w3c/aria-at/blob/master/tests/support.json#L123-L435).
[...]

PR #1598, opened 2025-10-29
fix: aria/html feature front end fixes

stalgiag:
related to #1519

>  1 id="main" is missing from the main element so the skip to main link does not work.
    2.     The link in the test plan column goes to the report for that test plan rather than the test plan. I think that is OK, possibly better than the specified design. However, I suggest we change the title of the column to "Test Plan Report".
>  3.     I suggest we change the name of the "Test Title" column to "Test"
>  4.     The links in the test title column don't work for me. The right page loads, but the fragment is not scrolling to the specific test. The focus ends up on the first link.

> The test plans tab is out of order. it should be sorted alphabetically, but it appears to be n some random order.

Addresses the above feedback on the aria/html feature reports


howard-e (approved):
Looks good!

> The test plans tab is out of order. it should be sorted alphabetically, but it appears to be n some random order.

#1612 additionally helps with this

Issue #1589, opened 2025-10-26
Make it possible to share a URL to a specific tab on pages with tabs


mcking65:
On the test queue page and the reports page, there are tabs. The browser address bar does not reflect the selected tab, so it is not possible to share a link to a specific tab. Particularly for reports, it would be very helpful to share a URL to the top-level summary for ARIA features or HTML features.

ChrisC:
@mcking65 we're hoping to get this in before TPAC because it's a really low lift, but curious if you have any preferences on the url scheme? 

In our feature branch, we've implemented with query params (`reports?tab=test-plans`, `reports?tab=aria-features`, `reports?tab=html-features`). Perhaps they should just be url params (`reports/aria-features`, etc)?  Any preferences here? 

cc @stalgiag

PR #1606, opened 2025-10-30
feat: tab urls

stalgiag:
Allows links to pages with specific tabs open using fragments

related to #1589

### DIFF
--- a/client/components/Reports/Reports.jsx
+++ b/client/components/Reports/Reports.jsx
@@ -1,6 +1,6 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import { useQuery } from '@apollo/client';
-import { Routes, Route, Navigate } from 'react-router-dom';
+import { Routes, Route } from 'react-router-dom';
 import { Container } from 'react-bootstrap';
 import { Helmet } from 'react-helmet';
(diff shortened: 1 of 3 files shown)


client/components/common/Tabs/Tabs.jsx
```diff
@@ -1,15 +1,57 @@
...
+    if (queryIndex !== selectedTab) {
+      setSelectedTab(queryIndex);
+    }
+  }, [location.search, tabs]);
+
+  const updateSelectedTab = index => {
+    setSelectedTab(index);
+    const tab = tabs[index];
+    const tabKey = tab.tabKey || `tab-${index}`;
+    const searchParams = new URLSearchParams(location.search);
```
ChrisC: It doesn't look like it was specified in the original feature request but I think it's worth confirming with Matt. I would prefer just regular urls like `reports/aria-features` etc... but I can ask on the original issue.

stalgiag: Oh good thought! I like that design better. Should be an easy swap. Can handle it on Monday.

howard-e (commented):
@stalgiag this is such a great quality of life change!

Issue #1521, opened 2025-09-15
Rename test plan directories


ccanash:
The requirements for the renaming are pending. 

The changes would need to be applied to: 

1. Scripts that run in the aria-at repo.
2. JSON embedded in HTML comments in the body of GitHub issues that connect them to test plans in the app.
3. APG content queries that pull in the interop report data.
4. The app creates a TestPlanVersion grouping based on the test plan directory name being imported. Some mechanism would be needed to  confirm the change coming with the newly imported test plan directory name should be propagated to the previous test plan directory names for that "group". Otherwise 'Checkbox' with directory name 'checkbox' would most likely be seen as different in the app from 'Checkbox' with directory name '<prepend->checkbox'

PR #1613, opened 2025-11-03
feat: Support for subfolders in `/aria-at/tests/*`

howard-e:
Addresses #1521 

Related work for this at https://github.com/w3c/aria-at/pull/1334.

So far, all patterns imported into the app have been related to APG, so a migration in this PR renames the directory names to `apg/<directory>`

New examples/patterns will be included at https://github.com/w3c/aria-at/pull/1335 while this work is in development

TODO:
- [ ] Fix the issues tracker finding the current issues with the updated directory names (quick fix applied but a more robust solution would be preferred)
- [x] Fix path for `<TestPlanVersionsPage>` component
- [ ] Fix tests


stalgiag (approved):
Manually tested and read through the core changes here in `scripts/import-tests` and all looks good to me. Thank you, Howard! 🙌

PR #1617, opened 2025-11-05
fix: Support dynamic renames of `aria-at/support.json/references` properties

howard-e:
This PR is to ensure https://github.com/w3c/aria-at/pull/1337 change is reflected as expected.

The support PR [aria-at#1338](https://github.com/w3c/aria-at/pull/1338) should be reviewed and merged first.

* Updates import-tests script to save the latest support.json on the server
* When needed, uses the support.json to mutate references' `linkText` and `value`


gnarf (approved):
nothing looks out of place to me, but I can't really tell what the big migration is actually doing.  I don't think it's that important I understand it however because it didn't seem to break anything on my local :)

PR #1619, opened 2025-11-05
chore(docs): Update documentation on Automation Workflow Configuration

gnarf:
Updating some automation based documentation.  See also https://github.com/bocoup/aria-at-gh-actions-helper/pull/90

Very interested in comments from the crew on what other things might need documentation, happy to add them, just not thinking of much here.


stalgiag (approved):
This reads more clearly. Thanks for the update!

howard-e (approved):

PR #1621, opened 2025-11-06
chore(migrations): avoid using `vendors.txt` to dictate Vendor id in seeder

howard-e:
Previously, the amount of lines in the vendors.txt in combination with postgres' `ON CONFLICT` was unintentionally bumping the `nextval` of new Vendor table entries.

This PR instead uses VENDOR_NAME_TO_AT_MAPPING to dictate currently known vendor information and better support for future entries.

This issue appears to only affect local development environments and shouldn't have any notable impact on deployed environments


gnarf (commented):
Code changes look fine, haven't tested on local

stalgiag (approved):
Tested locally and it works. Thanks!

Issue #1603, opened 2025-10-30
Incorrect value 'None' for 'menu' role in VoiceOver column of ARIA features report


mcking65:
The top-level report of ARIA features support levels shows a value of 'None' in the VoiceOver column for the role 'menu'. It doesn't seem like this should be possible as there are VoiceOver tests with mapping to the menu role; they are the same assertions as JAWS and NVDA.

howard-e:
I checked here and there seems to be 3 patterns for this:
* Action Menu Button Example Using aria-activedescendant - menu-button-actions-active-descendant
* Action Menu Button Example Using element.focus() - menu-button-actions
* Navigation Menu Button - menu-button-navigation

For "Action Menu Button Example Using element.focus()" and "Navigation Menu Button", those assertions with a 'menu' refId are all MAYs and haven't been marked with any exceptions. But the MAYs aren't covered, only MUSTs and SHOULDs.

"Action Menu Button Example Using aria-activedescendant" has 'menu' for the 'interactionModeEnabled' assertion with SHOULD alone. The other is a MAY, 'roleMenu'. 'interactionModeEnabled' is excepted as excluded for the 'navForwardsToMenuButton' and 'navBackToMenuButton' tests then re-included as SHOULDs through the nvda and jaws command.csvs. But the same isn't done for VoiceOver.

mcking65:
@howard-e 

thank you for the excellent analysis. You're right. Since VoiceOver does not have a mode switching behavior, there are no priority 1 or 2 assertions related to the menu role, which is a rather amazing state of affairs. It makes me want to re-examine mappings of assertions to the menu role.

That said, the word "None" is problematic because it implies VoiceOver has no support for the menu role. I really believe we should replace the string "None" with "No Data". The string "None" has appeared in other places in reports. I can't find another example at this moment. If it is simple to globally replace "None" with "No Data" in report rendering, that would improve their clarity and understandability.

PR #1622, opened 2025-11-06
fix: Prefer "No Data" instead of "None" on reports pages

howard-e:
Address [this comment in #1603](https://github.com/w3c/aria-at-app/issues/1603#issuecomment-3493306300)


stalgiag (approved):
Thanks!

Issue #1568, opened 2025-10-02
Infrastructure: Deploy steps should be moved to a more recent node LTS


howard-e:
Currently, the deploy steps install and expect the [project to run on node 18](https://github.com/w3c/aria-at-app/blob/70f7dbdb61d6a53f981567dfceb6e46513e98e66/deploy/roles/nodejs/tasks/main.yml#L21). [Node 18 has entered EoL with Node 22 being the current LTS](https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch).

Due to the age of Node 18, this project is locked into some dependencies which will continue to be aged out. It's also seemingly not the version typically used for local development.

Would recommend that Node 22 becomes the latest node version to support on the deployed environments.

PR #1637, opened 2025-11-10
chore: Support `node 22.x`

howard-e:
Address #1568

Updating support for node 22+ across:
* Packages
* Deployment Environments

Significant callouts:
* [rawgit](https://github.com/rgrove/rawgit) has long been abandoned, for proxying the aria-at tests through the app's test pages. It causes node-gyp dependency issues with node 22+ so moved to handling the *.js, *.css and *.html files manually

TODO:
- [x] Update node 22 to be used instead of node 18 during deployment


howard-e:
Drafting to review dynamic installs of node 22.x through ansible playbook

PR #1640, opened 2025-11-13
fix: Display additional information for reports

howard-e:
Address https://github.com/w3c/aria-at-app/issues/1083#issuecomment-3383473134

This PR does the following on report pages:
* Expands trends table to be displayed for Candidate reports
* Shows exact and minimum AT version information where previously missing

### DIFF
--- a/client/tests/e2e/Reports.e2e.test.js
+++ b/client/tests/e2e/Reports.e2e.test.js
@@ -213,17 +213,17 @@ describe('Report page for candidate report', () => {
       const validJawsHeadingMatchFound = await checkForHeading(
         page,
         'h2 ::-p-text(JAWS)',
-        'JAWS and Chrome'
+        'JAWS 2021.2111.13 or later and Chrome'
       );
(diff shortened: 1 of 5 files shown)


client/tests/e2e/Reports.e2e.test.js
```diff
@@ -242,20 +242,13 @@ describe('Report page for candidate report', () => {
       await page.click(viewResultsButton);
 
       await page.waitForSelector('h1 ::-p-text(Modal Dialog Example)');
-
-      const isVersionsSummaryFound = await page.evaluate(() => {
-        const elementsWithText = Array.from(
-          document.querySelectorAll('h2')
-        ).filter(element => element.textContent.includes('Versions Summary'));
-        return elementsWithText.length > 0;
-      });
```
stalgiag, line 251: This is a weird one. Glad it is much simpler now

stalgiag (approved):
Tested locally and this makes sense. Thanks for addressing this so quickly and precisely!

PR #1643, opened 2025-11-13
fix: populate test plan versions and filter on first disclosure open

stalgiag:
see title

fixes issue where when the "Add Test Plans" disclosure was first opened, the test plan version selected wasn't correctly updated to be filtered by the selected test plan.


howard-e (approved):
Changes look good, thanks for addressing so quickly!

PR #1644, opened 2025-11-13
fix: data management add test plan crash

stalgiag:
quick fix for error seen in the "Add Test Plans" disclosure opening causing crash.

The code now handles two different query structures correctly:
`DATA_MANAGEMENT_PAGE_QUERY`: Returns testPlanVersions at root level (each version already has testPlan.directory)
`TEST_QUEUE_PAGE_QUERY`: Returns testPlans with nested testPlanVersions arrays


howard-e (approved):
Thanks for the quick fix!

PR #1647, opened 2025-11-13
fix: Invalid `querySelector` call when deleting Test Queue reports

howard-e:
Since directory names now include a `/`. eg. 'alert' -> 'apg/alert' and those directory names have gotten directly used in this selector. The `/` isn't valid for `document.querySelector()`. This ensures it is properly escaped.

This issue can be reproduced when deleting a report on the Test Queue.


stalgiag (approved):
Thank you!

PR #1648, opened 2025-11-14
fix: escape directory selector on report delete

stalgiag:
Prior to this PR, deleting a report would throw

```
ERROR
Document.querySelector: 'h2#apg/checkbox' is not a valid selector
./components/TestQueue/Actions.jsx/Actions/deleteReport/_callee4/_callee4$/<@http://localhost:3000/components_TestQueue_index_jsx.bundle.js:2106:69
```

This is due to the directory having a forward slash that needs to be escaped after the nested directory migration


stalgiag:
Just noticed that Howard beat me to it with #1647

PR #1649, opened 2025-11-14
fix: allow add test plans disclosure to add test plan versions without initial reports

stalgiag:
This issue was discovered by @ChrisC while manual testing. The dropdown in `AddTestPlans` is populated with test plans that have at least one version in `DRAFT`, `CANDIDATE`, or `RECOMMENDED`. It also filters for test plans that have a version in one of those phases with an existing report. This secondary filter is incorrect.

This issue is quite old but likely wasn't discovered until now due to another method for adding initial reports being used.

### DIFF
--- a/client/components/ManageTestQueue/useAddTestPlansData.js
+++ b/client/components/ManageTestQueue/useAddTestPlansData.js
@@ -56,24 +56,20 @@ export const useAddTestPlansData = isOpen => {
 
     if (mainQueryData.testPlanVersions) {
       mainQueryData.testPlanVersions.forEach(version => {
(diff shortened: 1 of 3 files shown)


howard-e (approved):
Looks good, thanks!

> This issue is quite old but likely wasn't discovered until now due to another method for adding initial reports being used.

Makes sense, I remember this dropdown being dictated by the test plans. There's defaulting logic to show "Versions in R&D or Deprecated" when a test plan was selected with no viable versions, at least on an older version of Data Management.

This dropdown is now dictated by the test plan versions. Both ways are fine but I could see how the `reports.length` logic got conflated in putting that altogether. It certainly needed a central entry point from earlier. Well done!

PR #1650, opened 2025-11-18
fix: "Mark as Final" unexpectedly disabled

howard-e:
This PR addresses the following:

* The `draftTestPlanRuns.testResultsLength` attribute used to confirm if the "Mark as Final" button can be active was missing from the `TEST_QUEUE_EXPANDED_ROW_QUERY` query
* Various other mocks and silent error handling


stalgiag (approved):
Thanks for getting those other improvements in too!

Issue #1652, opened 2025-11-18
Test plans can't be moved from candidate to recommended on staging


ccanash:
When trying to move a test plan from candidate to recommended, the app is throwing an warning message saying there are conflicts when there aren't.

PR #1653, opened 2025-11-18
fix: Restore phase update to `RECOMMENDED` and restore negative side effects expected mutations

howard-e:
Address #1652 

This PR addresses 2 notable issues which affected older submitted reports but were part of advanced test plan versions:
1. Assertions marked as 'untestable' weren't in sync with another tester's run that marked it as 'failed'. This wasn't a problem before but now that conflicts consider these cases of untestable vs failed since #1599, it has to
2. Now that `AtBugs` and `encodedId` exists on a negative side effect since #1577, those also have to be explicitly ignored or even when the negative side effect matches, it shows up as a conflict

Apart from those, found that a created negative side effect could not be removed or updated for a Test Run page.

For reviewers:
To confirm the changes of the migration, locally I removed the `markedFinalAt` attribute for those TestPlanReport rows so I could easily view them in the Test Queue

### DIFF
--- a/server/migrations/20251118203125-fix-failed-to-untestable.js
+++ b/server/migrations/20251118203125-fix-failed-to-untestable.js
@@ -7,7 +7,9 @@ const { dumpTable } = require('./utils');
 module.exports = {
   async up(queryInterface) {
     // From production_dump_20251118.sql, found these to be the faulty reports which prevented progression to RECOMMENDED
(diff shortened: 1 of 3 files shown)


stalgiag (approved):
Wow! This is an important catch. Thanks for thinking it through and especially for being thorough in capturing the neg side effect copying edge behaviors here

PR #1667, opened 2026-06-08
miscellaneous security fixes

ChrisC:
* disables GraphQL introspection & playground on live servers (prod, staging, sandbox)
* restricts CORS to trusted origins
* Gates user role enumeration, update event data and collection job metadata behind auth

### DIFF
--- a/server/graphql-server.js
+++ b/server/graphql-server.js
@@ -1,17 +1,33 @@
 const { ApolloServer } = require('apollo-server-express');
 const {
-  ApolloServerPluginLandingPageGraphQLPlayground
+  ApolloServerPluginLandingPageGraphQLPlayground,
+  ApolloServerPluginLandingPageDisabled
 } = require('apollo-server-core');
 const graphqlSchema = require('./graphql-schema');
 const getGraphQLContext = require('./graphql-context');
 const resolvers = require('./resolvers');
 
+const isDeployed = ['production', 'staging', 'sandbox'].includes(
+  process.env.ENVIRONMENT
+);
+
 const apolloServer = new ApolloServer({
   typeDefs: graphqlSchema,
   context: getGraphQLContext,
   resolvers,
-  // The newer IDE does not work because of CORS issues
-  plugins: [ApolloServerPluginLandingPageGraphQLPlayground()]
+  introspection: !isDeployed,
+  debug: !isDeployed,
+  formatError: err => {
+    if (isDeployed && err.extensions) {
+      delete err.extensions.exception;
+    }
+    return err;
+  },
+  plugins: [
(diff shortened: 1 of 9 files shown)


daniel-montalvo:
Thanks @ChrisC for putting this together. I've just added my colleague @kfranqueiro  who would want to have a look at this. He works now for WAI and SysTeam, so I thing it'd be nice to have him in the loop.

daniel-montalvo:
Hi @ChrisC

After discussing with @kfranqueiro, he abstains from reviewing this at this point and would prefer for the Bocoup team to review this, as you may be much more aware of how to test it locally.

kfranqueiro:
> Hi @ChrisC
> 
> After discussing with @kfranqueiro, he abstains from reviewing this at this point and would prefer for the Bocoup team to review this, as you may be much more aware of how to test it locally.

Sorry for not chiming in here earlier. I looked over the changes and they appear to address what I expected, but as Daniel explained, I don't think I have any way to verify this prior to it being deployed, so I would defer to any reviewers who are able to test ahead of merging.

daniel-montalvo:
@mcking65 and @ChrisC  Thanks  for your work on this.

while we figure out next steps for the other vulnerability, do you think this one is ready to merge?

daniel-montalvo:
@mcking65 and @ChrisC  Thanks  for your work on this.

while we figure out next steps for the other vulnerability, do you think this one is ready to merge?

server/graphql-server.js
```diff
@@ -1,17 +1,33 @@
...
 const {
-  ApolloServerPluginLandingPageGraphQLPlayground
+  ApolloServerPluginLandingPageGraphQLPlayground,
+  ApolloServerPluginLandingPageDisabled
 } = require('apollo-server-core');
 const graphqlSchema = require('./graphql-schema');
 const getGraphQLContext = require('./graphql-context');
 const resolvers = require('./resolvers');
 
+const isDeployed = ['production', 'staging', 'sandbox'].includes(
```
howard-e, line 10: There's probably a small case for excluding the lower environments here, sandbox and maybe even staging. Just to help devs with some debugging in a deployed environment, if that's ever needed.

howard-e (approved):
Very late to this review but it looks ready to go, to me

cc @daniel-montalvo

server/graphql-server.js
ChrisC: I think since we're not doing heavy dev at the moment, we can err on the side of security. If it becomes an issue in the future, we can bring some of the deployed envs back!

howard-e: Sounds good!

